PR #223, opened 2021-07-31
feature: Add audit to admin client

### LINKED ISSUE #199: Add audit log support to Admin client
- Add support for the audit RPCs to the Admin client
- Replace `cerbosctl` custom client with Admin client

isdevx:
#### Description

Adds audit log methods to admin client. As this is still WIP, I'd like to discuss several things about testing before moving on. For instance; I am not sure how we want to use audit log for testing.

Fixes #199

#### Checklist 


client/admin.go
```diff
@@ -30,7 +35,8 @@ func NewAdminClient(address string, opts ...Opt) (AdminClient, error) {
 }
 
 // NewAdminClientWithCredentials creates a new admin client using credentials explicitly passed as arguments.
-func NewAdminClientWithCredentials(address, username, password string, opts ...Opt) (AdminClient, error) {
+func NewAdminClientWithCredentials(address, username, password string, opts ...Opt) (*GrpcAdminClient, error) {
```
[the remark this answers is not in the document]
charithe: Yeah, I got that was what you were trying to do -- which is the right thing in most cases. It is a suggestion and not a rule though and I am wondering whether it's the right choice here. Maybe we don't call the struct `GrpcClient` and find a slightly more generic name?  My worry here is that once people start using `GrpcClient` in their code, it'll have to be there forever. 

Not sure what's the right balance to strike here. I am open to ideas.

cmd/ctl/audit/audit.go
```diff
@@ -62,16 +62,16 @@ cerbos ctl audit --kind=access --since=3h --raw
 cerbos ctl audit --kind=access --lookup=01F9Y5MFYTX7Y87A30CTJ2FB0S
 `
 
-type clientGenFunc func() (svcv1.CerbosAdminServiceClient, error)
+type withClient func(fn func(c client.AdminClient, cmd *cobra.Command, args []string) error) func(cmd *cobra.Command, args []string) error
```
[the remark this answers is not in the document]
charithe: I am not sure I see the advantage of the more complicated signature but I don't feel too strongly about it to complain 😄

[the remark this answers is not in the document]
isdevx: It seemed more natural to me: (I'm not so persistent btw, just try to explain the motive)
```Go 
func SeriousCmdF(c client.AdminClient, cmd *cobra.Command, args []string) error {
    // requires client
}

cmd := &cobra.Command{
     RunE:    withClient(SeriousCmdF)
 }
```

```Go
func SimpleCmdF(cmd *cobra.Command, args []string) error {
    // no client needed
}

 cmd := &cobra.Command{
     RunE:    SimpleCmdF
 }
```

client/admin.go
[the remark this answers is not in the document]
isdevx: Hmm, if we don't let users to decide which client they should use, I think it's better to return a concrete type of `AdminClient` and remove the interface. Under the hood, it probably up to us to how to implement this.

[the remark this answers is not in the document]
charithe: I think the interface needs to be there. One of the reasons for having an interface from the beginning was to support testing as well. The idea is to provide SDK utilities that let the users easily inject errors etc. to test that their integration is robust. Having a ready-defined interface makes it easier for them to structure their code around it -- which makes stubbing  easy to implement.

client/admin_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/cerbos/cerbos/client"
+	"github.com/cerbos/cerbos/client/testutil"
+	"github.com/cerbos/cerbos/internal/audit"
+	"github.com/cerbos/cerbos/internal/audit/local"
+)
+
+func TestAuditLogs(t *testing.T) {
```
charithe, line 21: So, this test is not really doing what you expect :-). It's still reading from the `NopLog` which never errors. Even if you remove all the data loading calls, the test will still pass. The returned data is always `nil`.

I didn't expect the actual audit log to be ever used in tests to store and retrieve data. So the real audit log tries to read its configuration when you call `NewLog`. In the test environment, there's no configuration file and it falls back to `NopLog` when that's the case.  Therefore, the `WithAudit` option that has been added here does not really do anything. 

If we really want to be able to store and retrieve data from the audit log during tests, then we'll have to refactor the audit package to not rely on the config file being present and introduce other test-friendly options like using the Badger in-mem mode and removing the flush delay etc. 

If it were me, I would just find a way to test the _new_ code added here to make sure it works. The audit log has its own tests so we can be fairly confident that the interface we use to communicate with it is going to work as expected.

isdevx: I see, the best option is to add a audit log mock to the test server 👍

client/admin_test.go
```diff
@@ -0,0 +1,123 @@
...
+	svcv1 "github.com/cerbos/cerbos/api/genpb/cerbos/svc/v1"
+)
+
+func TestCollectLogs(t *testing.T) {
+	type ticker struct {
+		tick <-chan time.Time
+		tock <-chan time.Time
+	}
+
+	newTicker := func() ticker {
```
charithe, line 27: What is the purpose of the tickers here? Can't we achieve the same effect quicker by creating a pre-filled channel (or slice) of `ListAuditLogEntriesResponse` as the source for the receiver functions?

client/admin.go
```diff
@@ -46,23 +54,140 @@ func NewAdminClientWithCredentials(address, username, password string, opts ...O
...
 	}
 
 	return nil
 }
+
+type recvFn func() (*responsev1.ListAuditLogEntriesResponse, error)
+
+// collectLogs collects logs from the receiver function and passes to the channel
+// it will return an error if the channel type is not accepted.
+func collectLogs(receiver recvFn, channel interface{}) error {
```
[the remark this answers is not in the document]
isdevx: I see, this whole magic was derived form having two methods for this :) My ultimate goal was to have most simplistic methods it could have. As you suggested, these are admin API methods so we may overlook the simplicity here. Going to update accordingly.

client/admin_test.go
isdevx: Yeah we could, I was just trying to keep the order but we can simplify it 👍

client/model.go
```diff
@@ -851,20 +851,32 @@ type ServerInfo struct {
...
-	Log *auditv1.AccessLogEntry
-	Err error
+type AuditLogEntry struct {
+	accessLog   *auditv1.AccessLogEntry
+	decisionLog *auditv1.DecisionLogEntry
+	Err         error
+}
+
+func (e *AuditLogEntry) AccessLog() *auditv1.AccessLogEntry {
+	return e.accessLog
```
charithe, line 877: I am not sure why `Err` is a public member but log entries are private and accessible only through methods. Maybe you were trying to do something like the following?

```suggestion
type AuditLogEntry struct {
	accessLog   *auditv1.AccessLogEntry
	decisionLog *auditv1.DecisionLogEntry
	err         error
}

func (e *AuditLogEntry) AccessLog() (*auditv1.AccessLogEntry, error) {
      if e.err != nil {
          return nil, err
      }
	return e.accessLog, nil
```

client/admin_test.go
```diff
@@ -19,102 +18,66 @@ import (
...
-		}
-	}
-
 	t.Run("access logs", func(t *testing.T) {
-		clock := newTicker()
+		i := 0
 		receiver := func() (*responsev1.ListAuditLogEntriesResponse, error) {
-			select {
-			case <-clock.tick:
+			for ; i < 1; i++ {
```
charithe, line 24: Unless I am missing something, this loop doesn't do anything here. Removing it would not affect anything at all.

isdevx: ah yeah, it left there.

client/model.go
isdevx: Hmm, It is actually mean to be used like this:

```Go
if err := entry.Err; err != nil {
	// handle error
}
// do something with entry.AccessLog()
```

Since `error` is not the key value here I decided not to protect it as log entries. We can also protect it by making it private and add a method like we do for log entries.

charithe: Is there a reason to protect log entries? The inconsistent mixing of members and methods  does not feel right to me. Given that users are supposed to check for errors on every log entry anyway, I think I prefer the suggestion I made above where the method returns the log entry or the error.

isdevx: Fair enough 👍

cmd/cerbosctl/audit/audit.go
```diff
@@ -226,6 +226,10 @@ func newRichAuditLogWriter(out io.Writer) *richAuditLogWriter {
 }
 
 func (r *richAuditLogWriter) write(entry proto.Message) error {
+	if entry == nil {
```
charithe, line 229: I got a nil deref panic when running `cerbosctl audit` and I traced it here. It's a common Go pitfall: an interface holding a nil value is not nil. So, even if the underlying log entry is nil, this nil check will just let it through. 😄 

I think either you have to check for nil after the type switch below OR do the nil check in the `streamLogsToWriter` function above. My vote is for the latter because:
a) Not doing the nil check in `streamLogsToWriter` will result in two calls to `writer.write` per loop iteration with very little chance of it getting inlined or eliminated by the compiler -- therefore making it inefficient.
b) If the nil check is added after the type switch, you need to make the same change to `rawLogWriter.write` method as well. It's unnecessary duplication of code and still leaves the trap open for a future implementation of another `auditLogWriter` to accidentally omit the nil check and cause a panic.

isdevx: wow, this is.. 🤯

cmd/cerbosctl/audit/audit.go
```diff
@@ -145,16 +145,20 @@ func streamLogsToWriter(writer auditLogWriter, entries <-chan *client.AuditLogEn
 		if err != nil {
 			return fmt.Errorf("error while receiving access logs: %w", err)
 		}
-		if err := writer.write(aLog); err != nil {
-			return err
+		if aLog != nil {
+			if err := writer.write(aLog); err != nil {
+				return err
+			}
```
charithe, line 151: Maybe add a `continue` here because we know that if access log is not nil, decision log is definitely nil.

charithe (commented):
Small nit, but LGTM otherwise.

charithe (approved):
LGTM 👍🏽 

Please make sure to re-title the commit correctly (should be `feature:` not `(feature)`.

Issue #232, opened 2021-08-06
Compile cerbos ctl into a seperate binary


isdevx:
We can provide a separate ctl tool in our distribution. Most of time it's more convenient to use separate binary and let users explore the tools separately.

PR #237, opened 2021-08-07
chore: Add cerbosctl binary

isdevx:

#### Description

This PR enables `cerbosctl` to be distributed as a separate binary. Also added a server status method to the Admin API to get server version. If we are going to use cerbosctl as a remote tool, we may want to have a status method to check various things beforehand therefore we can provide debug steps. For instance after using a separate ctl tool for remote control, we may want to check server-client version mismatch.

Fixes: #232 

#### Checklist 


cmd/ctl/audit/audit.go
```diff
@@ -100,6 +101,15 @@ func runAuditCmd(clientGen clientGenFunc) func(*cobra.Command, []string) error {
...
 		}
 
+		r, err := client.ServerStatus(context.Background(), &requestv1.ServerStatusRequest{})
+		if err != nil {
+			return err
+		}
+
+		if r.Version != util.Version {
+			return fmt.Errorf("server version and client version does not match: server: %s, client: %s", r.Version, util.Version)
+		}
```
charithe, line 111: Is this necessarily a problem? The API should remain backward and forward compatible at least between minor and patch versions.

isdevx: Yeah actually we should do a semantic version check here. This was for demonstration purposes. BTW, do we provide the forward compatibility? I mean we should be able to add new api methods w/p bumping the major version right? In that case it should be handy to have a version check. At least we can warn the users.

isdevx: I see, what do you think of having a server status method for the API? Should we keep it anyway or go lean? :) BTW, we can use this utility to raise a warning for some possible inconsistent behaviors.

charithe: We can keep the `ServerStatus` RPC (but maybe rename to `ServerInfo` or something similar because "status" has a specific meaning in this context). I would remove the explicit version check for now and maybe change the output of the `version` flag to provide both the server and client API versions (if the server is online). Again, `kubectl` is a good example to follow but we probably don't need to output as much information as they do.

```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.0", GitCommit:"cb303e613a121a29364f75cc67d3d580833a7479", GitTreeState:"clean", BuildDate:"2021-04-08T16:31:21Z", GoVersion:"go1.16.1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.7", GitCommit:"132a687512d7fb058d0f5890f07d4121b3f0a2e2", GitTreeState:"clean", BuildDate:"2021-05-12T12:32:49Z", GoVersion:"go1.15.12", Compiler:"gc", Platform:"linux/amd64"}
```

isdevx: Makes sense 👍 Going to update accordingly.

charithe (commented):
The `cmd` directory is a bit inconsistent now. The subcommands of the `cerbosctl` command are in separate packages under the `ctl` directory whereas the subcommands of the `cerbos` command (`server`, `compile`) are at the same level as the `cerbos` directory. I think we need to rename `ctl` to `cerbosctl` and move `server` and `compile` under `cerbos` to maintain the logical structure.

Some of the targets in `hack/dev/dev.mk` needs to be updated to match the new structure as well. There are targets that expect the entrypoint to be `main.go` at the root -- which has now changed to `cmd/cerbos/main.go`.

api/cerbos/svc/v1/svc.proto
```diff
@@ -63,6 +63,16 @@ service CerbosService {
             description: "Check a principal's permissions to a batch of heterogeneous resources and actions."
         };
     }
+
+    rpc ServerInfo(cerbos.request.v1.ServerInfoRequest) returns (cerbos.response.v1.ServerInfoResponse) {
+        option (google.api.http) = {
+            get: "/api/status"
```
charithe, line 69: This should be `/api/server_info` because `status` means something else in this context.

charithe (commented):
```sh
$ cerbosctl version
ERROR: invalid credentials: username and password must be non-empty strings
```

I think this is because auth validation fires for every command but `version` should be excluded  from that.

api/cerbos/svc/v1/svc.proto
isdevx: Right, I was using the same command as before which included the username and password 🤦

cmd/cerbosctl/main.go
```diff
@@ -63,17 +65,20 @@ highest is: netrc < environment < command line.`
...
-cerbos ctl --server=localhost:3593 --username=user --password=password --plaintext decisions`
+cerbosctl --server=localhost:3593 --username=user --password=password --plaintext decisions`
 
-func NewCommand() *cobra.Command {
+func main() {
 	cmd := &cobra.Command{
-		Use:               "ctl",
-		Short:             "Cerbos control",
+		Use:               "cerbosctl",
+		Short:             "A remmote control tool for Cerbos",
```
charithe, line 76:
```suggestion
		Short:             "A remote control tool for Cerbos",
```

charithe (dismissed):
Spotted a minor typo but LGTM otherwise 👍🏽

PR #241, opened 2021-08-15
feature: Add basic auth (WIP)

### LINKED ISSUE #2: Authentication for API endpoints
Since we are providing a security service, users should be able to configure authentication for our service endpoints.

- Basic authentication (via htpasswd file)
- JWT (users provide public key to validated the claims)
- Client certificates (Premium only?)

isdevx:

#### Description

Adds basic authentication to `CerbosServiceClient`. I added a in memory user store to store credentials. Also, I added a basic authorization to use shared credentials (if necessary). Also, the `cerbosctl` now uses a token to auto login. This is still WIP due to lack of token refreshing and `CerbosAdminServiceClient` does not leverage the `AuthClient` yet.  Before going into deeper, I wanted to get some feedback. Some of my questions are:

- Shall we ask users for signing algorithm for JWT?
- Shall we use separate client for authentication?
- Will there be more than one administrators?

Fixes #2 

#### Checklist 


charithe (commented):
OK, this is an interesting take on the original issue. I do like the idea of `cerbosctl` behaving similar to `kubectl` in terms of logging in once and reusing the tokens. However, the original issue was not about that in particular and even for the `cerbosctl` login bit, I think we need to think about it and plan it a bit more. Let's create an RFC for this and come up with a design before writing any more code.

isdevx:
Closing this PR as it will be revamped when the design gets completed. We can re-open anyway.

PR #265, opened 2021-08-27
chore: Use admin client in decisions subcommand

isdevx:

#### Description

We can use the new admin client here as well. Great opportunity to remove some code :)

#### Checklist 

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -221,11 +221,11 @@ Contents of probable licence file $GOMODCACHE/contrib.go.opencensus.io/exporter/
 
 --------------------------------------------------------------------------------
 Module  : contrib.go.opencensus.io/exporter/prometheus
-Version : v0.3.0
-Time    : 2021-03-09T17:40:04Z
+Version : v0.4.0
+Time    : 2021-06-21T16:58:11Z
 Licence : Apache-2.0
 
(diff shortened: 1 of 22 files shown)


charithe (commented):
Sorry to nitpick, but, I was thinking more along the lines of an `internal` package to hold common code rather than a public one. Also, "commands" in this context is a bit misleading because it gives the impression that it is the package containing all the commands.

charithe:
Hyrum's law kind of applies here. Even though it should be obvious that `cerbosctl` package is not a library, it will still show up in Go docs and nothing prevents someone from building something using it. Given that this is an open source project, we don't have a lot of control over what people do with it so I'd prefer to minimise the exposure area if possible.

Does the additional nesting of packages add much value other than aesthetics? I'd prefer to keep the package structure flatter.

isdevx:
OK, with that point of view -which sounds good to me- we can choose to put every subcommand inside the `internal/commands` package and have a nice flat layout and share the common bits altogether. In fact, I used kind of a similar layout on other projects (w/o the `internal` bit).

charithe:
Just to be clear, all I am suggesting is to rename the `commands` package to `internal`. The current package structure is otherwise fine with me.

charithe (approved):
LGTM

Issue #136, opened 2021-06-08
Flesh out the Admin API


charithe:
- Better audit trail of changes by recording the user and metadata  (requires #2) 
- Allow deleting policies
- List/filter active policies

PR #269, opened 2021-08-30
enhancement: List policies endpoint added to Admin API

isdevx:
#### Description

Adds List Policies endpoint to the admin API.

Part of #136 

#### Checklist 


charithe:
I can see why you went with these choices for simplicity. Unfortunately, I don't think the filters can be expressed so simply in this case. Let's consider a typical user: when they were authoring policies, only derived roles had a `name` field. So, it's reasonable to assume that filtering the list by name would only apply to derived roles. Resource policies are keyed by `resource` and principal policies are keyed by `principal` so when looking for those, `name` is not the first thing that comes to mind.

Off the top of my head, the options that should be available are:

**List options**
- List all 
- List by kind
- List units (a policy and its dependencies)

**Filtering options**
- By name/resource/principal
- By policy version
- By description
- By metadata
- By creation date
- By status

**Sorting options (ascending and descending)**
- By name/resource/principal 
- By creation date
- By filename

Of course, this is a lot of work and all of it doesn't need to be built in one go. However, the scaffolding should be there in the API and the CLI so that we don't break the public contract as we add each piece.

isdevx:
Thanks for the insight @charithe! Much appreciated and I agree on your points. I was not so sure about filtering parameters and wanted to start small and expand from there. Your feedback will give a better perspective for sure. The first question for me is what is the difference between a filter option or list option?

For me `list - all` and `list - kind` sounded like a filter option. Also, should we think the `list -unit` as a second step (or another method) where we query for dependents of a single policy?

charithe:
`kubectl` and `gcloud` are good examples of  CLIs that get this kind of thing right (most of the time). Take a look at the [kubectl filtering commands](https://kubernetes.io/docs/reference/kubectl/cheatsheet/#viewing-finding-resources) for inspiration. 

The `list` command(s) should only provide summaries. So something like `list --all --sort=desc` would print something like the following:

```
ID          KIND         DEPENDENCIES     CREATED
---         ------        -----------------     ----------
leave_request:default  RESOURCE   my_derived_roles  2021-08-01 12:23:43
bugsbunny:default       PRINCIPAL    -                              2021-07-29 09:34:21
```
Then, if the `--format` flag is provided, then the output would be in JSON or YAML.

A good CLI should be human friendly AND machine friendly. For humans, there should be nice, clear and concise views available to quickly glance at the state of the system. For machines, the CLI should be easy to use in scripts so that useful information can be extracted without too much hassle.

internal/storage/disk/index/index.go
```diff
@@ -36,9 +40,11 @@ type Index interface {
 	GetFiles() []string
 	GetAllCompilationUnits(context.Context) <-chan *policy.CompilationUnit
 	Clear() error
+	GetPolicies(context.Context, storage.PolicyFilter) ([]*policy.Wrapper, error)
 }
 
 type index struct {
+	dir          string
```
isdevx, line 47: It seems like this addition is against the abstraction here and I'm not particularly happy breaking it here. But in order to get file creation date, we require the absolute path of the file. As stdlib does not provide a method to access file creation date out of the box.

internal/storage/db/sqlite3/schema.sql
```diff
@@ -22,7 +22,7 @@ CREATE TABLE IF NOT EXISTS policy_dependency (
 
 -- name: policy_revision
 CREATE TABLE IF NOT EXISTS policy_revision (
-    revision_id INTEGER AUTO INCREMENT PRIMARY KEY,
+    revision_id INTEGER PRIMARY KEY AUTOINCREMENT,
```
isdevx, line 25: This may look irrelevant but found that it was not working during testing.

charithe (commented):
I'll schedule some time to talk about this offline because we seem to have different ideas about how this should work.

internal/storage/disk/index/index.go
isdevx: Discussed offline, the creation date will be deferred to another PR.

api/public/cerbos/request/v1/request.proto
```diff
@@ -353,3 +353,35 @@ message ServerInfoRequest {
...
+    string description = 6;
+
+    bool disabled = 7;
+
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
+		json_schema: {
+			description: "List policies request"
+		}
+    };
+}
```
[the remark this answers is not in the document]
isdevx: Updated accordingly.

internal/svc/admin_svc.go
```diff
@@ -121,6 +119,51 @@ func (cas *CerbosAdminService) ListAuditLogEntries(req *requestv1.ListAuditLogEn
...
+	}
+
+	var policyKind string
+	switch req.Kind {
+	case requestv1.ListPoliciesRequest_KIND_RESOURCE:
+		policyKind = policy.ResourceKind.String()
+	case requestv1.ListPoliciesRequest_KIND_PRINCIPAL:
+		policyKind = policy.PrincipalKind.String()
+	case requestv1.ListPoliciesRequest_KIND_DERIVED:
+		policyKind = policy.DerivedRolesKind.String()
```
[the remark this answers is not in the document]
isdevx: Instead of creating the strings we can use the constants in the `internal/policy` package. Updated accordingly.

charithe (commented):
Thanks for reworking this. I appreciate that it was a lot of code to move about.

Unfortunately, I think there are still a few more things to address here. I tried using the list command but found it unintuitive. For example,  if I add `--resource-name=x` as the filter, I would only expect resource policies with the name containing `x` to show up. However it returns all policy kinds -- even those that don't have an `x` anywhere. Even more mind bending is the fact that if `--resource-name=x` doesn't match any resource policy, the output still shows all other policy kinds but NONE of the resource policies.    

The other strange thing is that if I specify the `version` flag more than once, it silently ignores everything except the last flag value. Same applies to `description`. Even though I knew that it was going to happen because I read the code first, it still kind of made me go WTF 😃 

I am also not too sure about pushing the filtering down to each storage driver. Sure, it's the most efficient way to do it but it also duplicates the filtering logic in multiple places. It's very easy to forget to add a new filter to one of the drivers during development or implement it differently from the rest -- thus producing inconsistent results if you switch backends.

I am quite fascinated by the idea that we could allow filtering to be done on any addressable field of a policy (see my comment below about defining filters in the command line) and even supporting JSONPath queries. It's difficult to do that if the filtering logic is in multiple places. So, I think that maybe we should start by implementing filtering _after_ the `Storage` abstraction. We can optimize it later (e.g. predicate push downs) once it starts to become a problem. I think that realistically a server won't have more than a few thousand policies stored in the worst case. Therefore, not doing the filtering within the store is not going to add too much overhead anyway.

cmd/cerbosctl/list/list.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	return nil
+}
+
+func printPolicies(w io.Writer, policies []*policy.Policy, format string) error {
+	switch format {
+	case "json":
+		for _, policy := range policies {
+			b, err := json.MarshalIndent(policy, "", "  ")
```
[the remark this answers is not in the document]
isdevx: Hmm, I am a bit confused here. From the tests I've made I didn't come across any weird output. Isn't `policyv1.Policy` a concrete struct? It has valid JSON tags from what I see. How can it not be used with JSON marshal?

cmd/cerbosctl/list/list.go
```diff
@@ -0,0 +1,103 @@
...
+			if err != nil {
+				return err
+			}
+			fmt.Fprintf(w, "%s\n", b)
+		}
+	case "yaml":
```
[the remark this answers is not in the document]
isdevx: Same as above. It printed me a valid YAML. Could you provide me your data set?

isdevx:
Thanks for the thorough review @charithe much appreciated! If I remember correctly we agreed on the idea that if we don't specify a value for `kind` it would mean all kinds right? And the filters like `resource-name`, `principal-name` etc. are only applicable to specific policy types. It is documented in the flags description. So, the output is expected to be like that as you are not filtering other kind of policies :)

Are you thinking something like a way to determine what kind of policies it should list according to id filters(eg. resource name, principal name and so on)? How about using a single ID filter? It would be applicable for all kinds but to different fields?

For the version flag, we can make it a string array type, currently it expects single value. Therefore it takes the last value you provided. Same for description. 

I agree to make filtering in the application layer, I think it won't cause too much trouble for now. We can do that 👍

cmd/cerbosctl/list/list.go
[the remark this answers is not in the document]
charithe: I am just using `make dev-server` to test. The data set used is the set of policies in `internal/test/testdata/store`.

charithe:
I think the filters should be a bit more intelligent.  I did say that if `kind` is not specified, it can be assumed to mean all. But, that is if you just run `cerbosctl list`. If you run `cerbosctl list --resource-name=x` then the `kind` should be assumed to be `resource` or the output it produces is very confusing. 

Anyway, as I mentioned in the comments, I don't think we should use flags like `--resource-name=x` to define filters. Instead, they should be like `--field-eq=resourcePolicy.resource=x`. As an added bonus, that approach actually enforces strict typing by default without requiring you to write any code because all you have to do is just loop through the list of policies and pluck out anything that contains the path `resourcePolicy.resource`.

isdevx:
@charithe I decided to remove filtering abilities from this PR so that we can have a smaller iterations over filter design. Since not defining any filter would mean "list all policies" current implementation won't break the future compatibility. Let's have a separate discussion on filtering and flags. I am a bit confused with the usage of implicit type filtering as it didn't sounded right way to do this. We can work on a better solution which can make both of us happy.

In fact, we can even expand the `--field-eq` type filtering without specifying types for everything. For example we can check if this field exists (eg. `--field-eq=version=x`) in any type of policy an so on. I am giving this example to emphasize that there are more opportunities on filtering. Don't take it as a proposal. Also thinking with the field paths do we even need `kind` flag anymore? Or, how do we do ands and ors with these? There are some unclear areas that I'd like discuss more.

Yet another reason I want to merge this sooner is we can even create more tasks based on this. We can work on sorting policies, a better human readable formatting and so on. As they are orthogonal with the filtering capabilities.

client/model.go
```diff
@@ -880,3 +880,14 @@ func (e *AuditLogEntry) AccessLog() (*auditv1.AccessLogEntry, error) {
...
+type PolicyKind uint8
+
+const (
+	ResourcePolicyKind PolicyKind = iota + 1
+	PrincipalPolicyKind
+	DerivedRolesPolicyKind
+)
+
+// FilterOpt is used to specify search terms for ListPolicies method.
+type FilterOpt func()
```
charithe, line 893: This is dead code now, isn't it?

cmd/cerbosctl/list/list.go
```diff
@@ -0,0 +1,102 @@
...
+			err := util.WriteYAML(w, policy)
+			if err != nil {
+				return fmt.Errorf("could not write policy: %w", err)
+			}
+			fmt.Fprintln(w, "---")
+		}
+	default:
+		headerFmt := color.New(color.FgGreen, color.Underline).SprintfFunc()
+
+		tbl := table.New("NAME", "KIND", "DEPENDENCIES")
```
charithe, line 76: I think VERSION is required here. This is the output when there are two versions of `leave_request` in the server. 

```
NAME              KIND           DEPENDENCIES      
my_derived_roles  DERIVED_ROLES  -                 
donald_duck       PRINCIPAL      -                 
leave_request     RESOURCE       my_derived_roles  
leave_request     RESOURCE                         
```

Ideally the output should be sorted by name as well.

internal/storage/db/internal/model.go
```diff
@@ -96,3 +97,16 @@ func (gpw *GeneratedPolicyWrapper) Scan(src interface{}) error {
...
+	ID          namer.ModuleID
+	Kind        string
+	Name        string
+	Version     string
+	Description string
+	Action      string
+	Disabled    bool
+	Definition  PolicyDefWrapper
+	Timestamp   time.Time `db:"update_timestamp"`
+}
```
charithe, line 112: Isn't this dead code?

internal/svc/admin_svc.go
```diff
@@ -121,6 +119,30 @@ func (cas *CerbosAdminService) ListAuditLogEntries(req *requestv1.ListAuditLogEn
...
+	if cas.store == nil {
+		return nil, status.Error(codes.NotFound, "store is not configured")
+	}
+
+	units, err := cas.store.GetPolicies(context.Background())
+	if err != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("could not get policies: %s", err.Error()))
+	}
+
+	policies := make([]*policyv1.Policy, 0)
```
charithe, line 136:
```suggestion
	policies := make([]*policyv1.Policy, 0, len(units))
```

charithe (commented):
Good idea to leave out the filters from this PR. It was getting unwieldy. 

LGTM except for a couple of issues highlighted below.

client/model.go
isdevx: Yeah, it is actually. Let's remove the `PolicyKind` and `FilterOpt`.  I left the `FilterOpt` as it may be used in the future but if we are going to use it with `ListPolicies(ctx context.Context, opts ...FilterOpt)` fashion, actually it won't break if we add options in the future.

cmd/cerbosctl/list/list.go
isdevx: Makes sense, going to add it 👍 And yeah this can be sorted alphabetically.

internal/storage/db/internal/model.go
isdevx: I think it isn't needed for now. But it is somehow going to be used when reading the policy revision. Anyway, I can remove it if you want.

charithe (approved):

PR #276, opened 2021-09-04
enhancement: Add version info while using make install

isdevx:

#### Description

We are now adding version and build information when using `make install`. Also, we are now setting a separate field for `cerbosctl`. With this change `cerbosctl` is totally independent from `cerbos/internal`.

#### Checklist 


charithe (commented):
This looks OK to me but I am wondering why (or how) we would release them separately? Why does `cerbosctl` need a different version, build date etc.?

isdevx:
Good point, I didn't thought these for release process though. If someone uses `make install` for development, it can be useful to put the build info inside the binary. I came across to an issue :) Also wanted to enable installing binaries separately so that install process can be faster if you are working on a particular one.

The reason for adding a separate build info to `cerbosctl` is to highlight that the tool is a completely separate client and it showcases how capable our API is. Even though the touch are for `cerbos/internal` was minimal, I think we should be strict about importing something from `cerbos/internal` package.

charithe:
I am not sure I follow your reasoning. The code is in a single repo and released in one go so the build information does not change at all. Why duplicate it? 

Personally I would use `go run` during development rather than `go install` -- which gets rid of the install problem. But even if you prefer the latter, does the build information on it matter for development purposes? Chances are you have dirty changes in the repo anyway so the version or the build time isn't going to give you a lot of useful information.

isdevx:
Yes, from the release perspective nothing has changed, the process is still same. We only store the version information in two different places and this is a cosmetic change.

I am more inclined to use `install` but it's just my workflow as most of the time I have a running server for a quite long time. And yes I'd like to know when did I compiled the last time (if I compiled server and ctl on different times).

charithe:
> I'd like to know when did I compiled the last time

Fair enough. My point is that this is a purely development time problem that can be solved more simply without duplicating things. Simply set the build information in `internal/util/app.go` whenever you install either one or both of the binaries.

isdevx:
Yep, that's correct. Even though the fields are duplicated, they are set automatically by the `make` rules or `goreleaser` commands which makes no hassle IMHO. It's a cosmetic choice to cut connection between `cerbosctl` and `cerbos/internal` packages.

charithe (approved):
LGTM 👍🏽

PR #280, opened 2021-09-07
enhancement: Adds CEL helper list functions 

dbuduev:
Adds helper CEL functions `is_subset`, `intersect`, `has_intersection`.


docs/modules/policies/pages/conditions.adoc
```diff
@@ -214,8 +214,13 @@ NOTE: The IP address functions are Cerbos-specific extensions to CEL.
...
 | filter   | Filter a list using the predicate | size(P.attr.teams.filter(t, t.matches("^comm"))) == 2
 | in       | Check whether the given element is contained in the list or map | ("design" in P.attr.teams) && ("acme" in P.attr.clients)
+| is_subset| Checks whether the list is a subset of another list | ["design", "engineering"].is_subset(P.attr.teams)
+| intersect| Produces the set intersection of two lists | intersect(["design", "engineering"], P.attr.teams)
+| has_intersection| Checks whether the lists have at least one common element | has_intersection(["design", "engineering"], P.attr.teams)
+| except | Produces the set difference of two lists | P.attr.teams.except(["design", "engineering"])
+| + | Concatenates lists | P.attr.teams + ["design", "engineering"]
 | map      | Transform each element in a list | "DESIGN" in P.attr.teams.map(t, t.upperAscii())
-| size     | Number of elements in a list or map | size(P.attr.teams) == 4 && size(P.attr.clients) == 2 
+| size     | Number of elements in a list or map | size(P.attr.teams) == 4 && size(P.attr.clients) == 2
```
charithe, line 223: Let's keep the table entries sorted alphabetically by function name. 

It would be nice to show  the expected output of functions so that users can better understand what they do. For example, `intersect(["design", "engineering"], P.attr.teams) == ["design", "engineering"]`. I tried to stick to this convention by writing test cases for all the examples in the documentation. Because the condition evaluator always expects an expression to return a boolean, it automatically forces you to write the example in that way. The test cases are in https://github.com/cerbos/cerbos/tree/main/internal/test/testdata/cel_eval and executed by https://github.com/cerbos/cerbos/blob/092fb0454672aead84ee0cd5cbe232abf5717b77/internal/compile/evaluator_test.go#L224

charithe (commented):
Nice work! 👍🏽 
Apart from the nit about the documentation sort order, it's good to go in.

internal/codegen/cel_lib.go
```diff
@@ -67,10 +106,138 @@ func (clib cerbosLib) ProgramOptions() []cel.ProgramOption {
...
+		for bi := b.Iterator(); !found && bi.HasNext() == types.True; {
+			vb := bi.Next()
+			if va.Equal(vb) == types.True {
+				found = true
+			}
+		}
+		if !found {
+			items = append(items, va)
+		}
+	}
```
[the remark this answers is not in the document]
dbuduev: I think for the majority of list item types we can use `ref.Val`, which is an interface, as a map key, but I presumed the lists are short, so nested loop implementation is good enough. I will add a benchmark test though.

docs/modules/policies/pages/conditions.adoc
```diff
@@ -194,7 +194,7 @@ NOTE: The IP address functions are Cerbos-specific extensions to CEL.
   "id": "elmer_fudd",
   "attr": {
     "id": "125",
-    "teams": ["design", "product_marketing", "communications", "commercial"],
```
dbuduev, line 197: I've updated the documentation here to align it with the tests.

internal/codegen/cel_lib.go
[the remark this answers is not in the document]
dbuduev: Benchmark before optimisation:
```
goos: darwin
goarch: amd64
pkg: github.com/cerbos/cerbos/internal/codegen
cpu: Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
BenchmarkIntersect50
BenchmarkIntersect50-8   	   23661	     49981 ns/op
BenchmarkIntersect25
BenchmarkIntersect25-8   	   74607	     16267 ns/op
BenchmarkIntersect15
BenchmarkIntersect15-8   	  153432	      7958 ns/op
BenchmarkIntersect5
BenchmarkIntersect5-8    	  372901	      3228 ns/op
PASS
ok  	github.com/cerbos/cerbos/internal/codegen	5.928s
```
Benchmark after:
```
goos: darwin
goarch: amd64
pkg: github.com/cerbos/cerbos/internal/codegen
cpu: Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
BenchmarkIntersect50
BenchmarkIntersect50-8   	   91986	     12666 ns/op
BenchmarkIntersect25
BenchmarkIntersect25-8   	  160879	      7312 ns/op
BenchmarkIntersect15
BenchmarkIntersect15-8   	  230719	      5165 ns/op
BenchmarkIntersect5
BenchmarkIntersect5-8    	  418629	      2850 ns/op
PASS
ok  	github.com/cerbos/cerbos/internal/codegen	5.338s
```
The optimisation uses a map if an underlying item type is hashable (see `hashable` function).
I ended up applying this technique to all functions.

docs/modules/policies/pages/conditions.adoc
dbuduev: Done

charithe (approved):
Great job!

PR #281, opened 2021-09-10
enhancement: Adds support for test fixtures

### LINKED ISSUE #14: Support for test fixtures
Right now every test case requires defining the whole request in full. Instead, we should allow parts (or the whole) of the request to be defined only once. For example `resource` or `principal` fragments that can be reused multiple times in different test scenarios.

dbuduev:

#### Description

Fixes #14 

#### Checklist 


internal/test/testdata/verify/suite.yaml
```diff
@@ -1,26 +1,25 @@
...
+    id: john
+    policyVersion: '20210210'
+    roles:
+      - employee
+    attr:
+      department: marketing
+      geography: GB
+      team: design
 tests:
   - name: Alicia tries to view her own private album
```
dbuduev, line 15: The test suite seems to follow the "HR application" policy example, but the test is named "Alicia tries to view her own private album". If I am not mistaken, can we align the test and the policies in `internal/test/testdata/store` with the playground "HR application"?

charithe: Agree on changing the inconsistent test names. That is an omission on my part.

I think that the HR application in the playground has deviated quite a bit from the test policies in `internal/test/testdata/store`.  If we try to sync them, a lot of tests will fail because those "store" policies are used in many places. Since there's no real link between the playground and these tests, let's not bother with that.

dbuduev: I see.

What do you think of having root-level principals and resources and then referring?
```yaml
principals:
  - &john
    id: john
    policyVersion: '20210210'
    roles:
      - employee
    attr:
      department: marketing
      geography: GB
      team: design
```
and then a policy can refer to the principal with YAML anchor
```yaml
principal: *john
```

dbuduev: I remember we discussed this. I submitted a draft PR for transparency and to collect comments if there are any. I think an RFC would be overkill here.
I will be following this approach to add root level resources, then I will add more tests and will update documentation showcasing this.

charithe (commented):
YAML anchors are a perfectly sensible solution for this but I am biased because I have spent an inordinate amount of time wrangling YAML. This simple case looks reasonable enough but let's try to model a sufficiently realistic test suite a user might write (just a YAML prototype, no need to have proper schemas yet) to see what the UX looks like.

internal/test/testdata/verify/suite.yaml
charithe: I commented below about prototyping a realistic test suite. Let's see what's the most user-friendly way to model a test matrix is. Also worth remembering is that there will be more than one policy to test and the user might want to have a common set of principals, resources etc. to be shared between all test suites.

dbuduev:
I've created an [RFC](https://github.com/cerbos/cerbos/discussions/285). Please have a look.

PR #293, opened 2021-09-14
enhancement: Strict configuration

charithe:
Fail if configuration contains unknown or invalid config values. The
only exception is for the `audit` section which needs to support plugins
that are not previously known.


internal/audit/conf.go
```diff
@@ -3,17 +3,58 @@
...
+	//   backend: local
+	//   local:
+	//     storageDirectory: /var/cerbos
+	//
+	// Because the config for the "local" backend is nested under the "audit" key, we will
+	// need to have "local" defined as a field in the Conf struct when strict parsing is on.
+	// However, backends are self-contained plugins and it is not practical to add a new field
+	// definition to the audit config struct for each plugin we introduce.
+	// This hack is slightly inefficient because it marshals and unmarshals YAML twice. It is an
+	// acceptable sacrifice to make because config is only read once on startup.
```
dbuduev, line 42: Does it mean schema (of the audit section) should be enforced except for the `local` field?

charithe: The reverse actually. If there's an unknown field in the `audit` section, it will be ignored by the parser. The `local` section is strictly typed.

internal/audit/conf.go
```diff
@@ -3,17 +3,58 @@
 
 package audit
 
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v2"
```
dbuduev, line 9: The project already uses `gopkg.in/yaml.v3` indirectly. What are your thoughts about v3 vs v2?

charithe: The config library (https://github.com/uber-go/config) uses `yaml.v2` , hence why I stuck to using `yaml.v2` here as well.

dbuduev:
LGTM

dbuduev (approved):

PR #294, opened 2021-09-15
enhancement: Add support for test fixtures

### LINKED ISSUE #14: Support for test fixtures

dbuduev:
#### Description

Adds support for test fixtures.
Discussion: https://github.com/cerbos/cerbos/discussions/285

Fixes #14 

#### Checklist 


charithe (commented):
Looking good 👍🏽  I didn't dive too deep because it's still a draft but I pointed out some of the things that stood out to me. 

One thing I am wondering is whether we should support inline principal and resource definitions as well. Otherwise, in order to get a test suite up, the user needs to create three files and a directory -- which is a lot of ceremony. I think we should make it very easy to get started writing tests and then refactor them as they get complicated. WDYT?

Finally, once the implementation is done, let's remember to update the testing documentation and the photo-share tutorial to reflect the changes made here.

dbuduev:
I have implemented the changes you have proposed including the embedded principals and resources option.

TODO:
- [ ] update the documentation 
- [ ] check the UX of misconfigured tests is good

internal/verify/test_fixture.go
```diff
@@ -0,0 +1,203 @@
...
+	return sr, failed
+}
+
+var (
+	ErrPrincipalNotFound = errors.New("principal not found")
+	ErrResourceNotFound  = errors.New("resource not found")
+)
+
+func formatTestName(tableTestName, principal string) string {
+	return fmt.Sprintf("%q by principal %q", tableTestName, principal)
```
charithe, line 149: Just wondering aloud here: should we make this structured (machine and human readable)? Is there a benefit to that?

The normal output looks OK

```
Test results
= TestSuite (suite_test.yaml) 
== "John and his leave request" by principal "john" [OK]
```

But, the JSON output not so much:

```
go run ./cmd/cerbos compile --tests=internal/test/testdata/verify internal/test/testdata/store -f json
{
  "results": [
    {
      "file": "suite_test.yaml",
      "suite": "TestSuite",
      "tests": [
        {
          "name": "\"John and his leave request\" by principal \"john\""
        }
      ]
    }
  ]
}
```

dbuduev: What about wrapping these string values in single quotes?

charithe: Also realised that the JSON output doesn't show whether the test passed or failed -- which is a bug that has existed even before this refactoring. Mea culpa.

charithe: > What about wrapping these string values in single quotes?

How about converting `Test.name` field from a `string` to a `message` that can hold both the test name and the principal name?. Then the output function can determine how to format it based on context.

dbuduev: Fixed both above issues.

dbuduev: Having field `name` of type struct is a bit unintuitive. What do you think about adding principal to TestResult type instead and make TestResult implement the `Stringer` interface to produce the value we have now.

dbuduev: >Also realised that the JSON output doesn't show whether the test passed or failed -- which is a bug that has existed even before this refactoring. Mea culpa.

Fixed.

charithe: > Having field `name` of type struct sounds a bit unintuitive to me in this context. What do you think about adding principal to TestResult type instead and make TestResult implement the `Stringer` interface to produce the value we have now.

Sounds good to me.

dbuduev: >How about converting Test.name field from a string to a message that can hold both the test name and the principal name? Then the output function can determine how to format it based on context.

I ended up going this way. Please have a look.
```
go run ./cmd/cerbos compile --tests=internal/test/testdata/verify internal/test/testdata/store -f json
{
  "results": [
    {
      "file": "suite_test.yaml",
      "suite": "TestSuite",
      "tests": [
        {
          "name": {
            "name": "John and his leave request",
            "principal": "john"
          },
          "failed": false
        }
      ]
    }
  ]
}
```
`name.name` doesn't look good. Probably need to rename the inner `name` to something else.

charithe: How about calling the outer field `case`? Then it would look like the following;

```json
{
  "results": [
    {
      "file": "suite_test.yaml",
      "suite": "TestSuite",
      "tests": [
        {
          "case": {
            "name": "John and his leave request",
            "principal": "john"
          },
          "failed": false
        }
      ]
    }
  ]
}
```

dbuduev: Done

dbuduev:
I deliberately misspelt a field name in the test suite and I think the error message ("failed to unmarshal JSON:  proto" part) might look a bit scary for a user:
```
ERROR: failed to run tests: failed to unmarshal JSON: proto: (line 1:240): unknown field "resources"
```

docs/modules/policies/pages/compile.adoc
```diff
@@ -14,45 +14,73 @@ docker run -i -t -v /path/to/policy/dir:/policies {app-docker-img} compile /poli
...
-<3> List of tests in this suite
-<4> Name of the test
-<5> Input to the policy engine
-<6> Expected outcome for each of the actions defined in the input
+<3> Map of resources. A key is a string that can be used to refer to the associated resource
+<4> Map of principals. A key is a string that can be used to refer to the associated principal
+<5> List of tests in this suite
+<6> Name of the test
+<7> Input to the policy engine
+<8> List of outcome expected for a specified principal and a given action.
```
charithe, line 66:
```suggestion
<8> List of outcomes expected for a specified principal and a given action.
```

docs/modules/policies/pages/compile.adoc
```diff
@@ -14,45 +14,73 @@ docker run -i -t -v /path/to/policy/dir:/policies {app-docker-img} compile /poli
...
 
+It is possible to share principals and resources between
+test suites stored in the same directory. Create a `testdata` directory in the directory with your test suite files, then put shared resources and principals to `testdata\resources.yml` and `testdata\principals.yml` respectively (`yaml` and `json` extensions are also supported).
```
charithe, line 69:
```suggestion
test suites stored in the same directory. Create a `testdata` directory in the directory containing your test suite files, then define shared resources and principals in `testdata\resources.yml` and `testdata\principals.yml` respectively (`yaml` and `json` extensions are also supported).
```

docs/modules/policies/pages/compile.adoc
```diff
@@ -14,45 +14,73 @@ docker run -i -t -v /path/to/policy/dir:/policies {app-docker-img} compile /poli
...
+---
+principals:
+  john:
+    id: johnID
+    roles:
+      - user
+      - moderator
+----
+
+To run the tests, provide the path to the tests' directory using the `--tests` flag.
```
charithe, line 83:
```suggestion
To run the tests, provide the path to the tests directory using the `--tests` flag.
```

docs/modules/policies/pages/compile.adoc
```diff
@@ -14,45 +14,73 @@ docker run -i -t -v /path/to/policy/dir:/policies {app-docker-img} compile /poli
...
+<5> List of tests in this suite
+<6> Name of the test
+<7> Input to the policy engine
+<8> List of outcome expected for a specified principal and a given action.
 
+It is possible to share principals and resources between
+test suites stored in the same directory. Create a `testdata` directory in the directory with your test suite files, then put shared resources and principals to `testdata\resources.yml` and `testdata\principals.yml` respectively (`yaml` and `json` extensions are also supported).
 
-To run the tests, provide the path to the tests directory using the `--tests` flag.
+.An example of `testdata\principals.yml`
```
charithe, line 71: Let's add an example for `resources.yaml` as well. We can also illustrate how to use YAML anchors and link to an external site that has good documentation on the subject.

internal/verify/verify.go
```diff
@@ -36,13 +34,24 @@ type SuiteResult struct {
...
+func (r TestName) String() string {
+	return fmt.Sprintf("'%s' by principal '%s'", r.TableTestName, r.PrincipalKey)
+}
+
 type TestResult struct {
-	Name    string `json:"name"`
-	Skipped bool   `json:"skipped,omitempty"`
-	Failed  bool   `json:"failed,omitempty"`
-	Error   string `json:"error,omitempty"`
+	Name    *TestName `json:"case"`
```
charithe, line 47: nit: Any reason why this has to be a pointer?

charithe (commented):
Almost there. Just a few small things to address.

internal/verify/verify.go
dbuduev: To exclude TestName struct from JSON output if the field is empty. But `Name` is always set, so there's no such problem. I can remove the pointer.

docs/modules/policies/pages/compile.adoc
dbuduev: Done

charithe (approved):
LGTM. Nice job 👍🏽

charithe:
I'll do a sudo merge to override the protobuf check.

PR #309, opened 2021-09-20
enhancement: Ignore invalid test suites

### LINKED ISSUE #307: Ignore invalid test suites and produce better error messages

### Current Behavior

If any of the `*_test.yml` files in the tests directory is invalid (e.g. empty), the whole test run fails and even the valid tests are not executed. 

```
ERROR: failed to run tests: failed to unmarshal JSON: proto: syntax error (line 1:1): unexpected token null
```

### Expected Behavior

- Valid tests should be run.
- Error message should output which test file contains the error

### Steps To Reproduce

```sh
touch tests/x_test.yaml
./cerbos compile --tests=tests policies
```

### Environment

```markdown
- OS:
- Cerbos version:
- Docker version:
- Go version:
```

### Anything else?

_No response_

charithe:
This change prevents the test runner from exiting as soon as it encounters an invalid test fixture or a suite. More context is added to the error messages to help locate the source of the problem as well.

Additionally, the test filtering logic has been slightly modified to be more user friendly. The new change makes it easier to run a specific test from a test suite without having to write a complex regex. 

Fixes #307


internal/verify/test_fixture.go
```diff
@@ -96,7 +96,7 @@ func (tf *testFixture) runTestSuite(ctx context.Context, eng *engine.Engine, sho
 		}
 
 		testResult := TestResult{Name: TestName{TableTestName: test.Name.TestTableName, PrincipalKey: test.Name.PrincipalKey}}
-		if test.Skip || !shouldRun(test.Name.TestTableName) {
+		if test.Skip || !shouldRun(fmt.Sprintf("%s/%s", ts.Name, test.Name.String())) {
```
dbuduev, line 99: Maybe we can run `shouldRun` check for both the test suite and each test in it.

dbuduev (approved):
LGTM. 👍🏻

internal/verify/test_fixture.go
charithe: The reason I removed the check from the suite is because it makes it tricky to write a regex for an individual test. For example, if you write `AlbumTestSuite.*delete.*`, then the match would fail for the suite name and we would skip the whole suite -- which prevents that specific test we wanted from being executed as well.

PR #318, opened 2021-09-23
enhancement: Allow policy rules to be named

### LINKED ISSUE #235: Policies should have unique names for manipulation with admin API
I checked the admin API: https://docs.cerbos.dev/cerbos/api/admin_api.html. 

I wanted to see how a policy is updated.

Not sure how to update a single policy. I don't see any user defined unique fields on the policies.

Maybe I am missing a point. ... I have spent limited time on the project overall.

charithe:
This change introduces the ability to name individual rules in a policy
definition. This helps produce better error messages and, in the future,
ability to update individual rules in a policy via the API.

To avoid breaking existing policies, names are optional for the time being.

Fixes #235


internal/codegen/codegen.go
```diff
@@ -96,7 +100,7 @@ func generatePrincipalPolicy(parent *policyv1.Policy, p *policyv1.PrincipalPolic
 
 	for i, rule := range p.Rules {
 		if err := rg.AddPrincipalRule(rule); err != nil {
-			return nil, newRuleGenErr(parent, i+1, err)
+			return nil, newRuleGenErr(parent, i+1, rule.Resource, err)
```
dbuduev, line 103:
```suggestion
			return nil, newRuleGenErr(parent, i+1, rule.Name, err)
```

dbuduev, line 103: nit:
A principal rule resource is used as a rule name, so an error message might be:
`Failed to generate code for rule 'leave_request' (#3)`. 

What do you think about having a separate `newPrincipalRuleGenErr` function producing the following?
`Failed to generate code for rule (#3) associated with resource 'leave_request'`.

internal/test/testdata/index/no_roles_in_rule.yaml
```diff
@@ -0,0 +1,47 @@
+---
+wantErrJson: |-
+  {
+    "disabled": null,
+    "duplicateDefs": null,
+    "loadFailures": null,
+    "codegenFailures": [
```
dbuduev, line 7:
```suggestion
    "codegenFailures": [
      {
        "error": "principal.yaml: Failed to generate code for rule 'leave_request' (#2) [Invalid match expression: Syntax error: token recognition error at: '= ']",
        "file": "principal.yaml"
      },
```

internal/test/testdata/index/no_roles_in_rule.yaml
```diff
@@ -0,0 +1,47 @@
...
+      version: "20210210"
+      rules:
+        - resource: leave_request
+          actions:
+            - action: "*"
+              condition:
+                match:
+                  expr: "request.resource.attr.dev_record == true"
+              effect: EFFECT_ALLOW
+
```
dbuduev, line 43:
```suggestion
        - resource: leave_request
          actions:
            - action: "*"
              condition:
                match:
                  expr: "request.resource.attr.dev_record === true"
              effect: EFFECT_ALLOW
              name: bad_cel_expr_leave_request

```

dbuduev (approved):
LGTM. I put a minor comment and suggest a test case for the new principal policy error message.

internal/codegen/codegen.go
charithe: Good idea.

dbuduev (approved):
LGTM. 👍🏻

PR #320, opened 2021-09-25
enhancement: Add sorting option to the list policies method options

### LINKED ISSUE #298: Adding sorting capabilities to the cerbosctl list command

### Feature description

It will be a useful feature if we can sort policies by their field values. We can use a similar logic with filtering and enable users to sort according to any field they desire. We can have two types of sort flags:
- `--sort-asc` for ascending
- `--sort-desc` for descending
This two flags can take a field selector as a parameter. For instance if someone wants to sort resource policies by their resource, they can simply use `--sort-asc resourcePolicy.resource`. The sorting logic should validate the path against the policy list if the fields are correct (we can either throw an error here, or do a filtering and return only matched ones, need some feedback).

We can have a special keyword (e.g. ID for resource/principal/name) to be able to sort different fields. Alternatively we if the parameter is not provided for the flag, we can sort by id by default. 

```
$ cerbosctl list --sort-asc=id
NAME              KIND           DEPENDENCIES      VERSION   
donald_duck       PRINCIPAL      -                 20210210  
leave_request     RESOURCE       my_derived_roles  20210210  
leave_request     RESOURCE                         staging   
my_derived_roles  DERIVED_ROLES  -                 -     
```

### Anything else?

This feature should be considered together with the #297. Probably needs to be implemented afterwards as both of the features should use same field selector logic.

isdevx:
#### Description

Added sorting option the the ListPolicies method. I had to rename the `FilterOpt` to `ListOpt` in the Go client so that we can use sort option together with the filter options. And we don't need to break the method signature published with `v.0.6`. Also did some minor file organization (ie. added `list.go` in `svc` package for `ListPolicies` related code)

As per discussions done offline, we decided to define an Enum for sortable fields instead of allowing users to use JSONPath expressions.

Two flags added to the `cerbosctl` and `--sort name` is set by default:
```
Flags:
      ...
      --sort string  Sort policies by ascending order (available sorting targets: name, version) (default "name")
      --sort-desc bool Sort policies by descending order
```
The Go client can accept two new options as sorting option:
```Go
func Sort(field ListPoliciesSortingType, descending ...bool) ListOpt {}
```

The sorting is done in the server so that it can work better with pagination and also we can proxy it to the DB.

Fixes #298 

#### Checklist 


api/public/cerbos/request/v1/request.proto
```diff
@@ -381,7 +381,25 @@ message ListPoliciesRequest {
...
+
+    message SortOptions {
+        bool descending = 1;
+        SortTarget target = 2 [
+            (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            description: "The field going to be sorted against"
+                enum: ["SORT_TARGET_NAME", "SORT_TARGET_VERSION"]
+            }
+        ];
+    }
```
charithe, line 398:
```suggestion
    message SortOptions {
        enum Column {
            COLUMN_UNSPECIFIED = 0;
            COLUMN_NAME = 1;
            COLUMN_VERSION = 2;
        }
        
        enum Order {
            ORDER_UNSPECIFIED = 0;
            ORDER_ASCENDING = 1;
            ORDER_DESCENDING = 2;
        }
        
        Column column = 1 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = { description: "Sort column"}];
        Order order = 2 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = { description: "Sort order"}];
    }
```

cmd/cerbosctl/internal/flags.go
```diff
@@ -180,6 +182,8 @@ func (lpfd *ListPoliciesFilterDef) FlagSet() *pflag.FlagSet {
 	fs := pflag.NewFlagSet("filters", pflag.ExitOnError)
 	fs.StringArrayVar(&lpfd.fieldEq, "field-eq", []string{}, "Filter a field with an exact match")
 	fs.StringArrayVar(&lpfd.fieldMatch, "field-match", []string{}, "Filter a field with a regex match")
+	fs.StringVar(&lpfd.sortAsc, "sort-asc", "", "Sort the output by ascending order (available sorting targets: name, version)")
+	fs.StringVar(&lpfd.sortDesc, "sort-desc", "", "Sort the output by descending order (available sorting targets: name, version)")
```
charithe, line 186: Because there's no support for multi-level sorting (e.g. by name then by version), I think it would be clearer to use `--sort` to get the column to sort by and an optional boolean flag `--sort-desc` to reverse the order.

client/model.go
```diff
@@ -923,20 +923,16 @@ func FieldMatchesFilter(path, value string) ListOpt {
...
 }
 
-// Sort enables sorting the policies by ascending order with given field.
-func Sort(field ListPoliciesSortingType) ListOpt {
-	return func(pf *policyListOptions) {
-		pf.sortingOptions = &sortingOptions{
-			field: field,
-		}
+// Sort enables sorting the policies by with given field.
+func Sort(field ListPoliciesSortingType, descending ...bool) ListOpt {
```
isdevx, line 927: This might be a little bit controversial but fits nicely as an optional field. Let me know if we should make this a mandatory argument.

charithe: I prefer to avoid boolean arguments in public methods. I think there should be two `ListOpt`s here called `SortAscending` and `SortDescending`.

internal/svc/list.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+type nameOrder []*policyv1.Policy
+
+func (s nameOrder) Len() int { return len(s) }
+
+func (s nameOrder) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
+
+func (s nameOrder) Less(i, j int) bool {
+	return getPolicyName(s[i]) < getPolicyName(s[j])
```
charithe, line 121: If the policy names are the same, then compare version to determine which comes first.

internal/svc/list.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+type versionOrder []*policyv1.Policy
+
+func (s versionOrder) Len() int { return len(s) }
+
+func (s versionOrder) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
+
+func (s versionOrder) Less(i, j int) bool {
+	return getPolicyVersion(s[i]) < getPolicyVersion(s[j])
```
charithe, line 131: Same as above. When versions are the same, compare names.

charithe (approved):
LGTM

PR #325, opened 2021-09-28
enhancement: User defined vars

dbuduev:
#### Description

Fixes #

#### Checklist 


dbuduev:
TODO:
[ ] s/aliases/globals/g
[ ] condition should refer to globals map, not to an individual variable.
[ ] if possible to calculate globals only once inject a call to an appropriate function in Rego code

dbuduev:
This PR introduces global vars for resource policies only. 
I haven't updated the docs yet. The current resource policy example is simple and will not benefit from globals.
I suggest adding the second more complex example.

internal/compile/conditions.go
```diff
@@ -124,6 +143,38 @@ func (ce *CELConditionEvaluator) Eval(input interface{}) (bool, error) {
...
+		ast, issues := stdenv.Compile(def)
+		if issues != nil && issues.Err() != nil {
+			celLog.Warn("Global variable compilation failed", zap.Error(err))
+			return nil, issues.Err()
+		}
+		prg, err := stdenv.Program(ast)
+		if err != nil {
+			celLog.Warn("Global variable AST generation failed", zap.String(codegen.CELGlobalsIdent, name), zap.Error(err))
+			return nil, err
+		}
```
charithe, line 165: What if we generate the program for globals at code generation time and store it in the `CELConditionEvaluator`? Then we don't have to do this compilation process each time the condition is evaluated.

charithe (commented):
I wonder if we can use [activations](https://pkg.go.dev/github.com/google/cel-go@v0.8.0/interpreter#NewActivation) to lazily evaluate globals when they are referenced in a condition.

charithe:
Let's not worry about the once-per-policy evaluation in this PR. Let's get the `globals` working (even if they are a bit inefficient) and document how to use them. We can come back to the once-per-policy optimization later.

internal/compile/conditions.go
dbuduev: Yes, I think we can extend structures `codegen.Result` and `GeneratedPolicy` to include `globals map[string]*CELCondition` and `globals CelConditions map[string]*v1alpha1.CheckedExpr` respectively.

dbuduev:
@charithe I've updated the documentation. Please have a look.

docs/modules/policies/pages/resource_policies.adoc
```diff
@@ -39,3 +39,34 @@ resourcePolicy:
...
+  - actions: ["delete"]
+    condition:
+      match:
+        expr: request.resource.attr.geography == globals.principal_location
+    derivedRoles:
+    - direct_manager
+    effect: EFFECT_ALLOW
+----
+<1> This resource policy includes global variables section. Each variable is evaluated before evaluating any rule condition.
+A variable expression can contain anything that condition expression can have.
```
charithe, line 72: I would just add a `globals` section to the policy above and annotate it. The idea is to show the anatomy of a policy with all the possible fields you can have on one. Having two examples adds noise and breaks that flow.

Then I would add a new section to the conditions documentation and explain how to use `globals` when writing conditions. This example can be used there to make it clearer. 

It just occurred to me that `globals` should be supported in both derived roles and principal policies as well. Let's not forget those in this PR.

dbuduev: Done

internal/codegen/cel.go
```diff
@@ -79,26 +81,35 @@ type CELCondition struct {
...
-	return cc.env.Program(cc.ast)
+func (cc *CELCondition) Program(vars ...*exprpb.Decl) (cel.Program, error) {
+	if len(vars) == 0 {
+		return cc.env.Program(cc.ast)
+	}
+	opts := append([]cel.EnvOption{cel.Declarations(vars...)}, NewCELEnvOptions()...)
+	env, err := cel.NewEnv(opts...)
+	if err != nil {
+		return nil, err
+	}
```
charithe, line 92: Is it possible to use [`env.Extend`](https://pkg.go.dev/github.com/google/cel-go@v0.8.0/cel#Env.Extend) here?

docs/modules/policies/pages/conditions.adoc
```diff
@@ -99,7 +99,9 @@ condition:
         !(R.attr.qa == true || R.attr.canary == true))
 ----
 
-
+== Globals
+If your policy repeats the same expression in multiple conditions, you can declare it in the globals section of a policy and refer in the conditions to avoid a repeat.
+Please refer to xref:principal_policies.adoc[Principal policies] for a complete example.
```
charithe, line 104: I liked the example you added to the resource policy doc in a previous commit. Can we add it here? I know it feels weird to repeat things in multiple places, but, clicking a link to look at an example tends to interrupt the mental flow of the reader and it's frustrating to switch back and forth between tabs when trying to understand something.

internal/codegen/cel.go
dbuduev: Yes, thanks for pointing it out.

dbuduev:
@charithe I've resolved outstanding comments (except activations). Please have a look.

charithe (approved):
LGTM 👍🏽 

I think we can potentially make `evaluateGlobals` a bit more efficient by storing the compiled program for globals in the evaluator and skipping the compile step on each evaluation. But, if you want to address that in a different PR, I am OK with that.

internal/codegen/rego.go
```diff
@@ -344,3 +345,7 @@ func (rg *RegoGen) addEffectStringFunc(defaultEffect string) {
 	rg.line(`} else = `, defaultEffect)
 	rg.line()
 }
+
+func (rg *RegoGen) AddGlobals(globals map[string]string) {
```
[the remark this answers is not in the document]
dbuduev: It is being used, but it does very little now indeed. I'll address this later.

PR #327, opened 2021-09-29
enhancement: Functions for working with hierarchies

charithe:
Introduces CEL functions to work with hierarchies represented as dot-separated strings.


internal/test/testdata/cel_eval/hierarchy_funcs.yaml
```diff
@@ -0,0 +1,39 @@
+---
```
dbuduev, line 1: Do these tests cover something that is not already covered by tests in `cel_lib_test.go`?

internal/codegen/cel_types.go
```diff
@@ -0,0 +1,252 @@
+// Copyright 2021 Zenauth Ltd.
```
dbuduev, line 1: What do you think about putting this in a separate package `internal/codegen/types`? E.g. `internal/codegen/types/hierarchy.go`

internal/test/testdata/cel_eval/hierarchy_funcs.yaml
charithe: They are executed through the same code path used by the engine with access to the request object etc. I also use these tests to make sure that the examples work before I add them to the documentation.

internal/codegen/cel_types.go
charithe: Good idea.

charithe: I got carried away and moved almost all of CEL code (those not directly related to code generation) out of the `codegen` package to a `conditions` package. Then I realised that I would break your pending PR with that 🤦🏽 

https://github.com/cerbos/cerbos/compare/main...charithe:enhancement/cel_hierarchy_funcs_refactor

I think we should probably defer that refactoring until both PRs are merged. WDYT?

dbuduev: Yes, probably there will be fewer merge conflicts if we merge my PR before that refactoring.

dbuduev (approved):
LGTM

PR #330, opened 2021-09-30
bug: Preserve all imported derived roles during execution

### LINKED ISSUE #329: Only the last derived role import is effective

### Current Behavior

When a resource policy imports more than one derived role sets, only the last imported set becomes "active" during execution. 

### Expected Behavior

All imported derived role sets should be active during execution.

### Steps To Reproduce

See https://play.cerbos.dev/p/ftUWhd6XpyPh6155d1b9uQCeemHtkGHc

### Environment

```markdown
- OS:
- Cerbos version:
- Docker version:
- Go version:
```

### Anything else?

_No response_

charithe:
When multiple derived roles are imported, only the last set is "active"
within the policy execution environment. This is because the Rego import
statements override each other.

This PR fixes the above problem by merging all imports together
explicitly.

Also fixes an issue with the DB driver where multiple imports were not
getting stored correctly in the table.

Fixes #329

### DIFF
--- a/internal/codegen/rego.go
+++ b/internal/codegen/rego.go
@@ -69,11 +69,19 @@ func NewRegoGen(packageName string, imports ...string) *RegoGen {
 	}
 
 	rg.line("package ", packageName)
-	for _, imp := range imports {
-		rg.line("import ", imp)
+	if len(imports) > 0 {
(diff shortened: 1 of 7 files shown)


dbuduev (approved):
LGTM

PR #336, opened 2021-10-03
enhancement: Re-use compiled global expressions

dbuduev:
Evaluates global variables more efficiently by re-using compiled expressions.


charithe (approved):
LGTM

PR #344, opened 2021-10-05
enhancement: Compound expressions without parentheses

charithe:
[CEL does not short-circuit when evaluating logical
expressions](https://github.com/google/cel-spec/blob/master/doc/langdef.md#logical-operators).
Because of this, if a condition contains a compound logical expression,
it should be parenthesised to obtain the desired result.

For example:

```
match:
  all:
    of:
      - expr: a
      - expr: b || c
```

should be written in one of the following ways to be correct:

```
match:
  all:
    of:
      - expr: a
      - expr: (b || c)
```

```
match:
  all:
    of:
      - expr: a
      - any:
         of:
           - expr: b
           - expr: c
```

This PR automatically parenthesises expressions so that users do not
have to remember to do it themselves.

### DIFF
--- a/internal/codegen/cel.go
+++ b/internal/codegen/cel.go
@@ -79,7 +79,9 @@ type celGen struct {
 func (cg *celGen) addMatch(m *policyv1.Match) error {
 	switch t := m.Op.(type) {
 	case *policyv1.Match_Expr:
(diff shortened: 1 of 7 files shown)


charithe:
Superseded by #354

PR #358, opened 2021-10-14
docs: Offer minimal configuration and full configuration in documentation

### LINKED ISSUE #357: Offer two configuration types in the documentation

### Feature description

The configuration file shown in [the configuration page](https://docs.cerbos.dev/cerbos/latest/configuration/index.html) is too long and may seem complex.

We may split this section into two; 
1) Basic Configuration: Write a bare minimum configuration to run cerbos in local or docker.
2) Advanced Configuration: Use the currently displayed configuration from the page referred.

### Anything else?

[the configuration page in question](https://docs.cerbos.dev/cerbos/latest/configuration/index.html)

oguzhand95:
Signed-off-by: Oğuzhan D <oguzhandurgun95@gmail.com>

#### Description

Split configuration section to two;

Basic Configuration: Minimum configuration to run Cerbos in local. (For experimenting)
Advanced Configuration: All configuration parameters with example.

Fixes #357

#### Checklist 

### DIFF
--- a/docs/modules/configuration/pages/index.adoc
+++ b/docs/modules/configuration/pages/index.adoc
@@ -12,30 +12,16 @@ The Cerbos server is configured with a YAML file. Start the server by passing th
 
 NOTE: Config values can reference environment variables by enclosing them between `${}`. E.g. `$$${HOME}$$`.
 
-Following are the examples of how to provide the configuration file to run Cerbos server in container and as binary;
-[source,sh,subs="attributes"]
-----
-docker run --rm --name cerbos -d -v path/to/policies:/policies -p 3592:3592 {app-docker-img} server --config=/path/to/conf.yaml
-----
 
 [source,sh,subs="attributes"]
 ----
 ./{app-name} server --config=/path/to/config.yaml --set=server.httpListenAddr=:3592 --set=engine.defaultPolicyVersion=staging
 ----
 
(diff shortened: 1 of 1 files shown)


charithe:
Thanks for this. It's a very good effort. I have a few suggestions on how to structure it.

I think the way we should document this is by showing the _minimal_ configuration required to start Cerbos and then the _full_ configuration that shows all the possible configuration keys. Let's not call them "basic" or "advanced".

When it comes to the minimal configuration, there are two more things to note:
- The container does not require any configuration to start. It is configured by default to look for policy files mounted at `/policies` (See https://github.com/cerbos/cerbos/blob/main/conf.default.yaml)
- The binary requires at least one storage driver to be configured. So, an example of the minimal configuration would be the `storage` stanza in https://github.com/cerbos/cerbos/blob/main/conf.default.yaml.

(I am going by memory here, so please test to make sure I haven't gotten anything wrong)

So, the structure of the document should be:

- Configuration: description of how configuration works in general (e.g. using `${VAR}` to access environment variables and how configuration is passed to Cerbos via the `--conf` flag)
- Minimal configuration
    - Minimal configuration for the container
    - Minimal configuration for the binary 
-  Full configuration: annotated example showing all possible configuration values

Additionally, it might be worth double checking the existing configuration (grep for `conf.go` files) to make sure that we haven't missed anything from the full configuration example.

PR #418, opened 2021-11-02
docs: Generate full configuration overview from code

### LINKED ISSUE #374: Generate full configuration overview from code
The full configuration overview at https://docs.cerbos.dev/cerbos/prerelease/configuration/index.html#_full_configuration can be generated from the code automatically. It would help to keep the section of the documentation up-to-date without manual intervention.

The configuration keys in `conf.go` files are usually well documented, so that's a good starting point. We might need to use our own struct tags to make this more configurable (indicate optional fields, note mutual exclusions, show acceptable values etc.).

oguzhand95:
#### Description

Code generation draft for docs

Fixes #374

#### Subtasks

- [X] Index all structs given a list of packages.
- [X] Find structs implementing a certain type of interface.
- [x] Get tags.
- [x] Get comments.
- [x] Parse tags using `fatih/structtag`.
  - [x] Use `yaml` tag as field names instead of using struct fields
  - [x] Use `conf` tag to add extra information to documentation files (possible options: `required`, `optional`, `defaultValue`).
  - [ ] Add `mutuallyExclusive` to `conf` options.
- [x] Add required `conf` struct tags to all `conf.go` files in the project.
- [x] Generate docs from indexed configuration structs and write them to `docs/modules/configuration/partials/*`.
- [x] Combine all partial configuration docs into one.

#### Checklist 


hack/tools/confdocs.go
```diff
@@ -0,0 +1,67 @@
...
+
+	dir = filepath.Join(dir, "internal")
+
+	absPath, err := filepath.Abs(dir)
+	if err != nil {
+		return "", err
+	}
+
+	return absPath, nil
+}
```
charithe, line 67: I wonder whether there's a simpler way to do this. One option might be to call [`debug.ReadBuildInfo()`](https://pkg.go.dev/runtime/debug#ReadBuildInfo) and read `Main.Path` to find the module path. (I haven't tried it myself so I could be wrong about it.)

The other way is to find the path to this file (see https://github.com/cerbos/cerbos/blob/aa252b517c94a435a880771f173c762a198b01e5/internal/test/test.go#L59) and  then construct the path from that. We know that this file is in `hack/tools` so the root of the repo is `<this_path>/../..`.

oguzhand95: I tried using `_, currFile, _, ok := runtime.Caller(0) ` as we used it in tests, and it worked like a charm and it is much shorter now.

Thanks for the suggestion!

hack/tools/confdocs/confdocs.go
```diff
@@ -0,0 +1,324 @@
...
+
+			if !foundInterfaceMethod {
+				structImplementsInterface = false
+				break
+			}
+		}
+		if structImplementsInterface {
+			filteredStructs[structKey] = indexedStruct
+		}
+	}
```
charithe, line 89: I think you can use [`types.Implements`](https://pkg.go.dev/go/types#Implements) for this

hack/tools/confdocs/confdocs.go
```diff
@@ -0,0 +1,324 @@
...
+}
+
+func (cd *ConfDoc) addMethodsToIndexedStructs(file *ast.File, indexedStructs map[string]*Struct) {
```
charithe, line 92: Probably not necessary to do this because you can use `types.Implements`

PR #420, opened 2021-11-03
enhancement: Better errors when disabled services are accessed

charithe:
Produces helpful error messages if users try to access the Admin API
or the Playground API when they are disabled by configuration.


oguzhand95 (commented):
I think semantic PRs doesn't support `enhancement`,  maybe we can use `feat`?
Otherwise LGTM.

oguzhand95 (approved):

PR #435, opened 2021-11-11
docs: Generate docs from branches

charithe:
Generating from branches allows us to update documentation for previous releases if we find any bugs in them


oguzhand95 (approved):
LGTM

PR #448, opened 2021-11-18
feat: Attribute schemas with JSONSchema

oguzhand95:
#### Description

Addition of attribute schemas for requests made to Cerbos check endpoint.

##### Tasks
- [x] Basic support for JSONSchema draft-04 
- [x] Read schema from stores
  - [x] Disk Store
  - [x] Git Store
- [x] Add configuration parameters for customizing schema validation behaviour
  - [x] Add `ignoreExtraFields` and `enforcement` to schema configuration
  - [x] Depending on the `enforcement` level; reject, warn or do nothing
  - [x] Ignore extra fields in attributes depending whether `ignoreExtraFields` configuration is set or not.
- [x] Return validation errors in http response
- [x] Write docs
- [x] Add tracing
- [x] Write test cases 
 
#### Configuration section for schemas
```
schema:
 ignoreExtraFields: true # Ignore extra fields provided in request. If it is false, cerbos will reject it.
 enforcement: reject # possible values: none, warn, reject
```

#### Example schema file
```
---
apiVersion: "api.cerbos.dev/v1"
description: |-
 Schema definition file

principalSchema: # define schema for request.principal.attr (single)
  type: object
  properties:
    permissions:
      type: array
      description: "permissions defined for this user"
    editor_for_categories:
      type: array

resourceSchema: # define schemas for request.resource.attr (multiple)
  resource_blog:
    type: object
    properties:
      categories:
        type: string
  
  resource_something:
     ...
     ...
```

#### Checklist 


internal/schema/schema.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	var validation []jsonschema.KeyError
+	for _, keyError := range principalValidation {
+		validation = append(validation, keyError)
+	}
+
+	for _, keyError := range resourceValidation {
+		validation = append(validation, keyError)
+	}
```
charithe, line 77:
```suggestion
        numErrors := len(principalValidation) + len(resourceValidation)
        validation :=  make([]jsonschema.KeyError, 0, numErrors)
        messages := make([]string, 0, numErrors)
	
	for _, keyError := range principalValidation {
		validation = append(validation, keyError)
		messages = append(messages, fmt.Sprintf("%s: %s", keyError.PropertyPath, keyError.Message))
	}

	for _, keyError := range resourceValidation {
		validation = append(validation, keyError)
		messages = append(messages, fmt.Sprintf("%s: %s", keyError.PropertyPath, keyError.Message))
	}
	
	printFn := logging.FromContext(ctx).Error
	if m.conf.Enforcement == EnforcementWarn {
	    printFn = logging.FromContext(ctx).Warn
	}
	
	printfFn("Validation failure", zap.Strings("errors", messages"))
	
	
```

internal/schema/schema.go
```diff
@@ -0,0 +1,192 @@
...
+
+	for _, validationData := range validation {
+		msg := fmt.Sprintf("Schema validation failed: Path=%s - Message=%s", validationData.PropertyPath,
+			validationData.Message)
+		if m.conf.Enforcement == EnforcementWarn {
+			logging.FromContext(ctx).Warn(msg)
+		} else {
+			logging.FromContext(ctx).Error(msg)
+		}
+	}
```
charithe, line 87:
```suggestion

```

internal/schema/schema.go
```diff
@@ -0,0 +1,192 @@
...
+
+	if m.conf.Enforcement == EnforcementNone {
+		m.log.Warn("Ignoring schema validation due to enforcement being set to 'none'")
+		return nil, nil
+	}
+
+	principalValidation, err := m.validatePrincipal(ctx, input.Principal.Attr)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
oguzhand95: We only stop when there is something wrong internally (ex: marshalling failed etc.). We continue with the validation if there are validation errors.

Do you suggest that we continue even if there is an internal failure. (marshalling failed etc.)?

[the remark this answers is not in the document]
charithe: Ah, no. I misread the code, ignore me.

internal/storage/git/store.go
```diff
@@ -140,8 +141,18 @@ func (s *Store) GetPolicies(ctx context.Context) ([]*policy.Wrapper, error) {
 }
 
 func (s *Store) GetSchema(ctx context.Context) (*schemav1.Schema, error) {
-	// TODO(oguzhan): Implement reading schema for this Store type
-	return nil, fmt.Errorf("not implemented")
+	schemaFileAbsPath, err := filepath.Abs(filepath.Join(s.conf.CheckoutDir, s.conf.SubDir, schema.RelativePathToSchema))
+	if err != nil {
+		return nil, fmt.Errorf("failed to determine absolute path to the schema file [%s - %s]: %w",
```
charithe, line 146: This error message excludes `s.conf.SubDir` which we use above to construct the path. It could cause confusion when trying to debug this error message.

api/public/cerbos/schema/v1/schema.proto
```diff
@@ -0,0 +1,53 @@
...
+  ERROR_TYPE_UNSPECIFIED = 0;
+  ERROR_TYPE_PRINCIPAL = 1;
+  ERROR_TYPE_RESOURCE = 2;
+}
+
+message Error {
+    string path = 1;
+    string message = 2;
+    ErrorType type = 3;
+}
```
charithe, line 53:
```suggestion


message ValidationError {
enum Source {
  SOURCE_UNSPECIFIED = 0;
  SOURCE_PRINCIPAL = 1;
  SOURCE_RESOURCE = 2;
}
    string path = 1;
    string message = 2;
    Source source = 3;
}
```

internal/engine/engine.go
```diff
@@ -274,6 +277,31 @@ func (engine *Engine) evaluate(ctx context.Context, input *enginev1.CheckInput,
...
+	}
+
+	// If there are errors present in schema validation, return EFFECT_DENY for all actions
+	if validationErrorList != nil && len(validationErrorList.Errors) > 0 {
+		output.ValidationErrors = validationErrorList.SchemaErrors()
+
+		for _, action := range input.Actions {
+			output.Actions[action] = &enginev1.CheckOutput_ActionEffect{
+				Effect: effectv1.Effect_EFFECT_DENY,
+				Policy: noPolicyMatch,
```
charithe, line 298: Let's change this to a value like "skipped due to invalid input" or something like that.

internal/schema/error.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+}
+
+func MergeValidationErrorLists(validationErrors ...*ValidationErrorList) *ValidationErrorList {
+	noOfErrors := 0
+	for _, validationErrorList := range validationErrors {
+		noOfErrors += len(validationErrorList.Errors)
+	}
+
+	var errors = make([]ValidationError, 0, noOfErrors)
```
charithe, line 37: I don't think this is necessary here. It's good practice to pre-allocate slices if we know the size in advance. But, this is not a piece of code in the critical path and the readability hit from the optimisation is probably not worth it.

internal/engine/engine.go
```diff
@@ -274,6 +277,31 @@ func (engine *Engine) evaluate(ctx context.Context, input *enginev1.CheckInput,
...
+	err = engine.schemaMgr.Validate(ctx, input)
+	if err != nil && !schema.IsValidationErrorList(err) {
+		return nil, err
+	}
+
+	var validationErrorList *schema.ValidationErrorList
+	ok := errors.As(err, &validationErrorList)
+	if !ok {
+		validationErrorList = nil
+	}
```
charithe, line 289:
```suggestion
	var validationErrorList *schema.ValidationErrorList
	if err != nil  {
	         if ok := errors.As(err, &validationErrorList); !ok {
		    return nil, err
		 }
	}
```

internal/schema/error.go
```diff
@@ -0,0 +1,99 @@
...
+	Errors []ValidationError
+}
+
+func (e *ValidationErrorList) Error() string {
+	var errorString = ""
+	for _, value := range e.Errors {
+		errorString += fmt.Sprintf("%s\n", value.Error())
+	}
+
+	return errorString
```
charithe, line 80:
```suggestion
	var errorString strings.Builder
	for _, value := range e.Errors {
	        errorString.WriteString(value.Error())
	        errorString.WriteString("\n")
	}

	return errorString.String()
```

internal/schema/schema.go
```diff
@@ -0,0 +1,332 @@
...
+type Manager struct {
+	conf                *Conf
+	log                 *zap.SugaredLogger
+	store               storage.Store
+	principalProperties map[string]string
+	resourceProperties  map[string]map[string]string
+	schema              *schemav1.Schema
+	principalSchema     *jsonschema.Schema
+	resourceSchemas     map[string]*jsonschema.Schema
+	mu                  sync.RWMutex // mutex for schema, principalSchema and resourceSchemas fields
```
charithe, line 31: Convention is to declare the mutex just above the fields it protects

internal/schema/schema.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+	}
+
+	store.Subscribe(mgr)
+
+	return mgr, nil
+}
+
+func (m *Manager) Validate(ctx context.Context, input *enginev1.CheckInput) error {
+	if m.schema == nil {
```
charithe, line 64: Whenever you access `schema`, `principalSchema` or `resourceSchema` it has to be protected by a read lock ( `mu.RLock`) otherwise the mutex is pointless. I am only commenting here but the issue persists across all of the methods below.

charithe (commented):
It's getting there! I only did a quick read through. Will do a thorough one later. I commented on a few issues that need to be addressed. Also, I don't see any tests that exercise any of this new code in a substantial way. I think it's pretty important to make sure we have good test coverage because it's in the critical path of the application.

internal/schema/error.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	var schemaErrors = make([]*schemav1.ValidationError, 0, noOfErrors)
+	for _, validationError := range e.Errors {
+		schemaErrors = append(schemaErrors, &schemav1.ValidationError{
+			Path:    validationError.Path,
+			Message: validationError.Message,
+			Source:  validationError.Source,
+		})
+	}
```
isdevx, line 91:
```suggestion
	for i, validationError := range e.Errors {
		schemaErrors[i] = &schemav1.ValidationError{
			Path:    validationError.Path,
			Message: validationError.Message,
			Source:  validationError.Source,
		}
	}
```

internal/schema/schema.go
```diff
@@ -0,0 +1,340 @@
...
+	schema              *schemav1.Schema
+	principalSchema     *jsonschema.Schema
+	resourceSchemas     map[string]*jsonschema.Schema
+}
+
+func New(store storage.Store) (*Manager, error) {
+	conf := &Conf{}
+	if err := config.GetSection(conf); err != nil {
+		return nil, fmt.Errorf("failed to get config section %q: %w", confKey, err)
+	}
```
isdevx, line 38: nit: Usually it's easier for reader to use address operator when passing pointers to functions such as `config.GetSection(&conf)`.

internal/schema/schema.go
```diff
@@ -0,0 +1,340 @@
...
+	mgr := &Manager{
+		conf:                conf,
+		log:                 zap.S().Named("schema"),
+		store:               store,
+		principalProperties: make(map[string]string),
+		resourceProperties:  make(map[string]map[string]string),
+	}
+
+	err := mgr.ReadOrUpdateSchemaFromStore()
+	if err != nil {
```
isdevx, line 49: nit:
```suggestion
	if err := mgr.ReadOrUpdateSchemaFromStore(); err !=nil {
```

internal/schema/schema.go
```diff
@@ -0,0 +1,340 @@
...
+		var msg = fmt.Sprintf("schema file not found: %s", err)
+		if conf.Enforcement == EnforcementReject {
+			return nil, fmt.Errorf(msg)
+		} else {
+			mgr.log.Warn(msg)
+		}
```
isdevx, line 55:
```suggestion
		if conf.Enforcement == EnforcementReject {
			return nil, fmt.Errorf("schema file not found: %w", err)
		} else {
			mgr.log.Warnf("schema file not found: %s", err)
		}
```

isdevx (commented):
Did a quick pass, great work @oguzhand95 🚀

oguzhand95:
Thanks for the reviews! I have handled some of them and currently working on some more tests. After the tests are done, I think the PR is ready for the final review.

docs/modules/configuration/pages/index.adoc
```diff
@@ -161,6 +161,10 @@ audit: # Optional
       maxBatchSize: 32
       flushInterval: 1s
       gcInterval: 60s
+
+schema: # Optional
+  enforcement: reject # level of the validations (none, warn, reject)
```
charithe, line 166:
```suggestion
  enforcement: reject # How to enforce schema validation (none: no enforcement, warn: log warnings, reject: reject invalid requests) 
```

docs/modules/configuration/pages/index.adoc
```diff
@@ -161,6 +161,10 @@ audit: # Optional
...
+  ignoreExtraFields: false # allows extra fields to be accepted
```
charithe, line 167: Let's rename this to `ignoreUnknownFields`
```suggestion
  ignoreUnknownFields: false # Ignore unknown fields not defined in the schema
```

docs/modules/configuration/nav.adoc
```diff
@@ -5,3 +5,4 @@
 * xref:storage.adoc[Storage]
 * xref:audit.adoc[Audit]
 * xref:tracing.adoc[Tracing]
+* xref:schema.adoc[Schema]
```
charithe, line 8: Let's only link this up to the documentation once the feature is complete. We have a couple of more PRs to do to implement the other stores before the feature is ready for use.

internal/schema/conf.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2021 Zenauth Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+package schema
+
+import "path/filepath"
+
+const (
+	confKey                  = "schema"
+	defaultIgnoreExtraFields = true
```
charithe, line 10:
```suggestion
	defaultIgnoreUnknownFields = true
```

internal/schema/conf.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+var (
+	RelativePathToSchema = filepath.Join("_schemas", "schema.yaml")
+)
+
+// Conf holds configuration related to schema validation
+type Conf struct {
+	// IgnoreExtraFields allows extra fields to be accepted.
+	IgnoreExtraFields bool `yaml:"ignoreExtraFields"`
```
charithe, line 21:
```suggestion
	// IgnoreUnknownFields Ignores fields not defined in the schema 
	IgnoreUnknownFields bool `yaml:"ignoreUnknownFields"`
```

internal/schema/conf.go
```diff
@@ -0,0 +1,42 @@
...
+	// Enforcement level of the validations. (none, warn, reject)
+	Enforcement Enforcement `yaml:"enforcement"`
+}
+
+func (c *Conf) Key() string {
+	return confKey
+}
+
+func (c *Conf) SetDefaults() {
+	c.IgnoreExtraFields = defaultIgnoreExtraFields
```
charithe, line 31:
```suggestion
	c.IgnoreUnknownFields = defaultIgnoreUnknownFields
```

internal/schema/error.go
```diff
@@ -0,0 +1,94 @@
...
+	"fmt"
+	schemav1 "github.com/cerbos/cerbos/api/genpb/cerbos/schema/v1"
+	"github.com/qri-io/jsonschema"
+	"strings"
+)
+
+type ValidationErrorType string
+
+func NewValidationError(keyError jsonschema.KeyError, source schemav1.ValidationError_Source) *ValidationError {
+	return &ValidationError{
```
charithe, line 13: I don't think this needs to be a pointer because all fields are primitive types

internal/schema/error.go
```diff
@@ -0,0 +1,94 @@
...
+	return ok
+}
+
+type ValidationError struct {
+	Path    string
+	Message string
+	Source  schemav1.ValidationError_Source
+}
+
+func (e *ValidationError) Error() string {
```
charithe, line 61: Might be better to use non-pointer receiver here

internal/schema/error.go
```diff
@@ -0,0 +1,94 @@
...
+	return fmt.Sprintf("%s: %s", e.Path, e.Message)
+}
+
+type ValidationErrorList struct {
+	Errors []ValidationError
+}
```
charithe, line 67: I don't think we need the struct here

```suggestion
type ValidationErrorList []ValidationError
```

internal/schema/schema.go
```diff
@@ -0,0 +1,343 @@
...
+		principalProperties: make(map[string]string),
+		resourceProperties:  make(map[string]map[string]string),
+	}
+
+	if err := mgr.ReadOrUpdateSchemaFromStore(ctx); err != nil {
+		if conf.Enforcement == EnforcementReject {
+			return nil, fmt.Errorf("schema file not found: %w", err)
+		} else {
+			mgr.log.Warnf("schema file not found: %s", err)
+		}
```
charithe, line 54:
```suggestion
		} 
		
		mgr.log.Warnw("schema file not found", "error", err)
```

internal/schema/schema.go
```diff
@@ -0,0 +1,343 @@
...
+	store.Subscribe(mgr)
+
+	return mgr, nil
+}
+
+func (m *Manager) Validate(ctx context.Context, input *enginev1.CheckInput) error {
+	ctx, span := tracing.StartSpan(ctx, "schema.Validate")
+	defer span.End()
+	m.mu.RLock()
+	defer m.mu.RUnlock()
```
charithe, line 66: There are a couple of problems here:
- The lock is held for longer than necessary (we only need the lock to check whether the schema is nil or not)
- There's nested locking because `validatePrincipal` and other methods called from this function obtain the lock again

I would just use the lock once to grab copies of the protected fields and then pass those copies around to the other methods. It's better to avoid locking as much as possible.

```suggestion
	m.mu.RLock()
	schema := m.schema
	principalSchema := m.principalSchema
	// whatever other protected fields you need in the body below
	m.mu.RUnlock()
```

internal/schema/schema.go
```diff
@@ -0,0 +1,343 @@
...
+			validationError.Message))
+	}
+
+	printFn := logging.FromContext(ctx).Error
+	if m.conf.Enforcement == EnforcementWarn {
+		printFn = logging.FromContext(ctx).Warn
+	}
+
+	if len(validation) != 0 {
+		printFn("Schema validation failure", zap.Strings("Errors", messages))
```
charithe, line 120:
```suggestion
		printFn("Schema validation failure", zap.Strings("errors", messages))
```

internal/schema/schema.go
```diff
@@ -0,0 +1,343 @@
...
+		} else {
+			logging.FromContext(ctx).Error(msg)
+		}
+	}
+
+	if m.conf.Enforcement == EnforcementReject {
+		return MergeValidationErrorLists(principalValidationErrorList, resourceValidationErrorList)
+	} else {
+		return nil
+	}
```
charithe, line 138:
```suggestion
	} 
	
	return nil
```

internal/schema/schema.go
```diff
@@ -0,0 +1,343 @@
...
+		}
+		if value.Properties != nil {
+			m.walkSchemaProperties(fmt.Sprintf("%s/%s", path, name), value.Properties, writeTo)
+		}
+	}
+}
+
+func (m *Manager) ReadOrUpdateSchemaFromStore(ctx context.Context) error {
+	ctx, span := tracing.StartSpan(ctx, "schema.ReadOrUpdateSchemaFromStore")
+	defer span.End()
```
charithe, line 268: This method is not called from any of the API methods, is it? These trace spans won't be very useful because of that.

internal/schema/schema_test.go
```diff
@@ -0,0 +1,370 @@
...
+
+func TestValidate(t *testing.T) {
+	t.Run("schema.Manager.Validate", func(t *testing.T) {
+		var testCases = []struct {
+			name    string
+			input   enginev1.CheckInput
+			wantErr bool
+		}{
+			{
+				name: "Test 1",
```
charithe, line 23: Not a very good name 😄

internal/schema/schema_test.go
```diff
@@ -0,0 +1,370 @@
...
+				input: enginev1.CheckInput{
```
charithe, line 24: You could have used similar logic to [engine tests](https://github.com/cerbos/cerbos/blob/2a660b51e837a22f7c9f31a36e1b9cb0f16ebe53/internal/engine/engine_test.go#L57) (and most other tests in this repo) to read these from YAML files -- which are much easier to construct than writing protobufs by hand.

internal/schema/schema_test.go
```diff
@@ -0,0 +1,370 @@
...
+		var testCases = []struct {
+			name          string
+			inputSchema   string
+			wantPrincipal map[string]string
+			wantResource  map[string]string
+			wantErr       bool
+		}{
+			{
+				name:        "Test 1",
+				inputSchema: "LS0tCmFwaVZlcnNpb246ICJhcGkuY2VyYm9zLmRldi92MSIKZGVzY3JpcHRpb246IHwtCiBTY2hlbWEgZGVmaW5pdGlvbiBmaWxlCgpwcmluY2lwYWxTY2hlbWE6ICMgZGVmaW5lIHNjaGVtYSBmb3IgcmVxdWVzdC5wcmluY2lwYWwuYXR0cgogIHR5cGU6IG9iamVjdAogIHByb3BlcnRpZXM6CiAgICBwZXJtaXNzaW9uczoKICAgICAgdHlwZTogYXJyYXkKICAgICAgZGVzY3JpcHRpb246ICJwZXJtaXNzaW9ucyBkZWZpbmVkIGZvciB0aGlzIHVzZXIiCiAgICBlZGl0b3JfZm9yX2NhdGVnb3JpZXM6CiAgICAgIHR5cGU6IGFycmF5CiAgICBwZXJzb25hbF9kZXRhaWxzOgogICAgICB0eXBlOiBvYmplY3QKICAgICAgcHJvcGVydGllczoKICAgICAgICBuYW1lOgogICAgICAgICAgdHlwZTogc3RyaW5nCiAgICAgICAgbWFpbDoKICAgICAgICAgIHR5cGU6IHN0cmluZwoKcmVzb3VyY2VTY2hlbWFzOiAjIGRlZmluZSBzY2hlbWEgZm9yIHJlcXVlc3QucmVzb3VyY2UuYXR0cgogIHJlc291cmNlX2Jsb2c6CiAgICB0eXBlOiBvYmplY3QKICAgIHByb3BlcnRpZXM6CiAgICAgIGNhdGVnb3JpZXM6CiAgICAgICAgdHlwZTogc3RyaW5nCg==",
```
charithe, line 197: This test is not very maintainable because whoever is looking at this will have to decode this and reencode it. Didn't you introduce templating to the test framework to handle this kind of thing already?

internal/schema/schema.go
```diff
@@ -0,0 +1,343 @@
...
+	principalAttributes map[string]*structpb.Value) error {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	inputValidationErr := m.validateInput(principalAttributes, m.principalProperties, schemav1.ValidationError_SOURCE_PRINCIPAL)
+	if inputValidationErr != nil && !IsValidationErrorList(inputValidationErr) {
+		return fmt.Errorf("failed to validate the input: %w", inputValidationErr)
+	}
+
+	principalAttrBytes, err := json.Marshal(principalAttributes)
```
charithe, line 177: Would this correctly produce the JSON for `structpb` values? Protobufs have a special JSON encoding that the standard library `json` package can't handle (you need to use `protojson`).

charithe (approved):
Good job 👏🏽 

This is a large piece of work so we are going to collaborate and keep developing it in a feature branch until the feature is ready for release.

PR #457, opened 2021-11-25
chore: Refactor schema validation and add engine tests for schemas

oguzhand95:
#### Description

Fixes some errors on validation logic and adds schemas to some of the engine tests.

- Fixes the problem where if a resource type doesn't have the corresponding schema definition, schema manager ignores and continues.
- Fixes the problem where the wrong part of the input is passed down to schema manager parameters.
- Added some methods to schema manager to ease the process of testing it.
- Engine tests can declare which schema file to use now.
- Engine tests cover cases for schema validation now.

Fixes

#### Checklist 


internal/engine/engine_test.go
```diff
@@ -107,7 +114,7 @@ func runBenchmarks(b *testing.B, eng *Engine, testCases []test.Case) {
 	}
 }
 
-func mkEngine(tb testing.TB, enableAuditLog bool) (*Engine, context.CancelFunc) {
+func mkEngine(tb testing.TB, enableAuditLog, enableSchemaValidation bool) (*Engine, context.CancelFunc) {
```
charithe, line 117: Let's declare a struct to hold these boolean parameters. Understanding what `mkEngine(b, true, false)` means is harder compared to something like `mkEngine(b, param{enableAuditLog: true, enableSchemaValidation: false})`

charithe (approved):
Since there's only a nit about readability, I'll merge this because I am also working on this branch and don't want to have too many merge conflicts later.

PR #461, opened 2021-11-27
enhancement: Add store and Admin API services for schema feature

oguzhand95:
#### Description

This PR adds support for databases and blob stores.
Also, in order to be used with mutable stores, Admin API endpoints were created.
- `GET` /admin/schema
- `DELETE` /admin/schema
- `POST` /admin/schema

#### Checklist 


internal/storage/blob/store.go
```diff
@@ -228,6 +231,11 @@ func (s *Store) updateIndex(ctx context.Context) error {
 	var p *policyv1.Policy
 	var event storage.Event
 	for _, f := range changes.updateOrAdd {
+		if strings.HasSuffix(f, filepath.Join(schema.Directory, schema.File)) {
```
charithe, line 234: Is there a way we can ensure that this matches only  `<ROOT>/_schemas/schema.yaml` and nothing else like  `<ROOT>/some/path/_schemas/schema.yaml` etc?

internal/storage/db/internal/model.go
```diff
@@ -21,8 +22,49 @@ const (
...
+		return fmt.Errorf("unexpected type for schema definition: %T", src)
+	}
+
+	sdw.Schema = &schemav1.Schema{}
+	if err := sdw.Schema.UnmarshalVT(source); err != nil {
+		return fmt.Errorf("failed to unmarshal schema definition: %w", err)
+	}
+
+	return nil
+}
```
charithe, line 66: I think we should store the schema as JSON using the database's native JSON type. We have a request to do the same for policies as well but that requires a migration step that we can't do at the moment. However, because this is a new table, we can store the data as JSON from the start.

internal/storage/db/internal/model.go
```diff
@@ -21,8 +22,49 @@ const (
 	PolicyDepTbl            = "policy_dependency"
 	PolicyDepTblPolicyIDCol = "policy_id"
 	PolicyDepTblDepIDCol    = "dependency_id"
+
+	SchemaTbl              = "schema"
+	SchemaTblIDCol         = "id"
+	SchemaTblDefinitionCol = "definition"
+	SchemaDefaultID        = 1
```
charithe, line 29: Is this guaranteed to be always 1?

internal/storage/db/internal/db.go
```diff
@@ -48,9 +50,63 @@ type dbStorage struct {
...
+
 func (s *dbStorage) GetSchema(ctx context.Context) (*schemav1.Schema, error) {
-	// TODO(oguzhan): Implement reading schema for this Store type
-	return nil, fmt.Errorf("not implemented")
+	var sch Schema
+
+	_, err := s.db.
+		Select(goqu.C(SchemaTblIDCol), goqu.C(SchemaTblDefinitionCol)).
+		From(SchemaTbl).
+		Where(goqu.Ex{SchemaTblIDCol: SchemaDefaultID}).
```
charithe, line 103: I don't think we can guarantee that the ID will be 1 given that the table definitions use auto incrementing keys. If you delete the schema and reinsert it, the ID would change.

charithe (commented):
Something that I have been thinking is whether it makes sense to have a single schema file any more. I know we did it to help with avoiding repetition but the schema definitions are quite verbose and when you have more than one or two resources, the file becomes quite large and hard to read. Now, looking at the database implementation, if we go with the single schema approach, it is possible that it could become too large to fit into a single column in a table. 

I think we might have to rethink whether It might be better to have a schema file per resource (or a database record per resource) rather than a single file (or record).

oguzhand95:
Yes, I think keeping the schema definitions under a single record might harm at the end.

The two solutions you've suggested;
1) Have multiple schema files.
2) Have a single schema file, but use multiple rows in the table to keep them.

For 1,
Having separation of context at the file system level might be nice since we are doing the same for policies. On the other hand, this might be tiresome for users since potentially they would have to create a schema file per resource.

For 2,
If there are too many resource policies, naturally there will be also schemas covering them and this would mean the schema file might get really long and this could harm the readability. We could solve the storage problem by having rows per resource kind, but this wouldn't affect the user.

internal/storage/db/internal/db.go
oguzhand95: Currently, as we only have one schema file, we only use the row with ID=1. (we always delete and insert with ID=1)
But, this will need to change, of course, if we introduce multiple schema files or multiple DB records per resource kind.

charithe (commented):
Let's discuss the issue of single vs. multiple schemas offline and update the issue with our conclusion later.

In the mean time I think this PR looks OK to merge after adding some tests to the DB test suite to make sure it all works.

internal/storage/db/internal/db.go
```diff
@@ -48,9 +50,71 @@ type dbStorage struct {
...
+	if err != nil {
+		return fmt.Errorf("failed to delete the schema before inserting: %w", err)
+	}
+
+	if _, err := s.db.Insert(SchemaTbl).
+		Rows(schemaRecord).
+		Executor().
+		ExecContext(ctx); err != nil {
+		return fmt.Errorf("failed to insert the schema: %w", err)
+	}
```
charithe, line 79: Since there's only one record, can't we simply do an `UPSERT` instead? See `AddOrUpdate` method.

internal/storage/db/internal/tests.go
```diff
@@ -140,5 +156,14 @@ func TestSuite(store DBStorage) func(*testing.T) {
 
 			checkEvents(t, storage.Event{Kind: storage.EventDeletePolicy, PolicyID: rpx.ID})
 		})
+
+		t.Run("delete_schema", func(t *testing.T) {
+			checkEvents := storage.TestSubscription(store)
+
+			err := store.DeleteSchema(ctx)
+			require.NoError(t, err)
+
```
charithe, line 165: Check that calling `GetSchema` returns nothing now

internal/test/schema.go
```diff
@@ -0,0 +1,93 @@
...
+package test
+
+import (
+	"google.golang.org/protobuf/types/known/structpb"
+
+	schemav1 "github.com/cerbos/cerbos/api/genpb/cerbos/schema/v1"
+)
+
+// GenSchema generates a sample schema.
+func GenSchema() *schemav1.Schema {
```
charithe, line 16: I see you're trying to follow the precedent of `GenPolicy` functions -- which is good. But, if you take a closer look, they are there to generate dynamic policies with variable number of policy rules whereas this `GenSchema` function simply produces a static schema. You could have easily just read it from a file instead and not spend so much effort on hand-crafting the protobufs 😄

charithe (commented):
Just a couple of nits but looks good to me otherwise

charithe (approved):
👍🏽

PR #470, opened 2021-12-06
enhancement: schema support for db stores

oguzhand95:
#### Description

Adds database support (Postgres, MySQL, SQLite) for schema storage. This PR also re-enables Admin API service endpoints for schemas.

Reason(s) to be draft;
- List schemas endpoint is not done yet

#### Checklist 


internal/storage/db/postgres/schema.sql
```diff
@@ -33,9 +33,7 @@ CREATE TABLE IF NOT EXISTS policy_revision (
 );
 
 CREATE TABLE IF NOT EXISTS attr_schema_defs (
-    id BIGINT PRIMARY KEY,
-    description TEXT,
-    disabled BOOLEAN default false,
+    id VARCHAR PRIMARY KEY,
```
charithe, line 36:
```suggestion
    id VARCHAR(255) PRIMARY KEY,
```

internal/storage/db/sqlite3/schema.sql
```diff
@@ -35,9 +35,7 @@ CREATE TABLE IF NOT EXISTS policy_revision (
 
 -- name: attr_schema_defs
 CREATE TABLE IF NOT EXISTS attr_schema_defs (
-    id BIG INT PRIMARY KEY,
-    description TEXT,
-    disabled BOOLEAN default false,
+    id VARCHAR PRIMARY KEY,
```
charithe, line 38:
```suggestion
    id VARCHAR(255) PRIMARY KEY,
```

internal/svc/admin_svc.go
```diff
@@ -85,71 +85,64 @@ func (cas *CerbosAdminService) AddOrUpdatePolicy(ctx context.Context, req *reque
...
+	}
 
-		sch, err := cas.store.GetSchema(context.Background(), req.Id)
-		if err != nil {
-			return nil, status.Error(codes.Internal, fmt.Sprintf("could not get schema: %s", err.Error()))
-		}
+	ms, ok := cas.store.(storage.MutableStore)
+	if !ok {
+		return nil, status.Error(codes.Unimplemented, "Configured store is not mutable")
+	}
```
charithe, line 117: For `GetSchema` it doesn't matter whether the store is mutable or not.

internal/storage/db/internal/db.go
```diff
@@ -42,82 +49,136 @@ func NewDBStorage(ctx context.Context, db *goqu.Database) (DBStorage, error) {
...
+		return fmt.Errorf("schema definition cannot be nil")
+	}
 
-		schemaRecord := Schema{
-			ID:          SchemaDefaultID,
-			Description: sch.Description,
-			Disabled:    sch.Disabled,
-			Definition:  SchemaDefWrapper{sch},
-		}
+	defJSON := pgtype.JSON{}
```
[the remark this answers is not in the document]
charithe: It appears to be so. 👍🏽

internal/storage/store.go
```diff
@@ -92,6 +92,7 @@ type MutableStore interface {
 	Store
 	AddOrUpdate(context.Context, ...policy.Wrapper) error
 	AddOrUpdateSchema(context.Context, string, []byte) error
+	GetSchema(context.Context, string) ([]byte, error)
```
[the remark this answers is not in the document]
oguzhand95: LoadSchema returns the json schema object with kind `*jsonschema.Schema`. We cannot send it as a response for `GET /admin/schema/{id}` as users probably will expect to see the original schema definition file. WDYT?

internal/storage/db/internal/tests.go
```diff
@@ -28,12 +29,14 @@ func TestSuite(store DBStorage) func(*testing.T) {
 	return func(t *testing.T) {
 		ctx, cancelFunc := context.WithCancel(context.Background())
 		defer cancelFunc()
-
+		ja := jsonassert.New(t)
```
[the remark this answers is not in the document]
oguzhand95: Yes, that does the job. Thanks!

internal/storage/store.go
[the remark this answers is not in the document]
charithe: That's a good point. Shall we change the signature of the loaders to just return an `io.ReadCloser` instead then?   

I think that having a `GetSchema` and `LoadSchema` is probably just confusing and needless. We can change `LoadSchema` to return `io.ReadCloser, error` instead (so, essentially be the `LoadURL` function that `jsonschema.Compiler` expects).

internal/schema/db_loader.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2021 Zenauth Ltd.
```
charithe, line 1: I don't think we need this anymore after the refactoring from #473

internal/schema/db_loader_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/cerbos/cerbos/internal/schema"
+	"github.com/cerbos/cerbos/internal/storage"
+	"github.com/cerbos/cerbos/internal/storage/db/sqlite3"
+	"github.com/cerbos/cerbos/internal/test"
+	"github.com/cerbos/cerbos/internal/util"
+)
+
+func TestDBLoader(t *testing.T) {
```
charithe, line 28: We can roll these tests into https://github.com/cerbos/cerbos/blob/4859fc2bad31b64862caa16f450fb89047e94833/internal/schema/schema_test.go#L30 by adding an extra loop to iterate through the different store implementations

api/public/cerbos/response/v1/response.proto
```diff
@@ -270,6 +270,16 @@ message AddOrUpdateSchemaResponse {
...
 }
 
+message ListSchemasResponse {
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
+        json_schema: {
+            description: "List of schemas stored in the Cerbos server"
+        }
+    };
+
+    repeated .cerbos.schema.v1.Schema schemas = 1;
```
charithe, line 280: I don't think we need the dot in front of the type name

```suggestion
    repeated cerbos.schema.v1.Schema schemas = 1;
```

api/public/cerbos/svc/v1/svc.proto
```diff
@@ -154,9 +154,25 @@ service CerbosAdminService {
...
+                    value: {};
+                }
+            }
+        };
+    }
+
     rpc GetSchema(cerbos.request.v1.GetSchemaRequest) returns (cerbos.response.v1.GetSchemaResponse) {
         option (google.api.http) = {
-            get: "/admin/schema"
+            get: "/admin/schema/{id}",
```
charithe, line 175: Shall we make `id` a query parameter instead of a path variable? I am a bit concerned about the asymmetry of the API. Currently it looks like:

- `POST /admin/schema`: Add/update schema
- `POST /admin/schema/list`: List schemas 
- `GET /admin/schema/list`: Get the schema with ID "list". 
- `DELETE /admin/schema/list`: Delete schema with ID "list" 

I think it would be better as follows: 
- `POST /admin/schema?id=list`: Add/update schema with ID "list"
- `POST /admin/schema/list`: List schemas 
- `GET /admin/schema?id=list`: Get the schema with ID "list". 
- `DELETE /admin/schema?id=list`: Delete schema with ID "list" 

WDYT?

oguzhand95: I think this looks much better, thanks for the suggestion!

internal/schema/db_loader_test.go
oguzhand95: I have come up with something, let me know if it looks good to you.

internal/schema/schema_test.go
```diff
@@ -68,26 +70,19 @@ func TestLoad(t *testing.T) {
...
+			testName := fmt.Sprintf("%s-%s", tc.name, storeName)
+			t.Run(testName, func(t *testing.T) {
+				err := mgr.CheckSchema(context.Background(), tc.url)
+				if tc.wantErr {
+					require.Error(t, err)
+					return
+				}
+				require.NoError(t, err)
+			})
+		}
```
charithe, line 85: We could nest them so that it's easier to run the tests for a single manager if we need to (e.g. `go test -tags=test -run=TestLoad/disk` to run just the disk store tests.

```suggestion
	for storeName, mgr := range mgrs {
	        t.Run(storeName, func(t *testing.T) {
		    for _, tc := range testCases {
			  tc := tc
			  t.Run(tc.name, func(t *testing.T) {
				err := mgr.CheckSchema(context.Background(), tc.url)
				if tc.wantErr {
					require.Error(t, err)
					return
				}
				require.NoError(t, err)
			  })
			})
		}
```

charithe (approved):

PR #507, opened 2021-12-16
docs: Generate full configuration overview from code

### LINKED ISSUE #374: Generate full configuration overview from code

oguzhand95:
#### Description

Code generation draft for docs

Fixes #374

#### Usage
Run `make docs` in the project root.

#### Subtasks

- [X] Index all structs given a list of packages.
- [X] Find structs implementing a certain type of interface.
- [X] Get tags.
- [X] Get comments.
- [X] Parse tags using `fatih/structtag`.
  - [X] Use `yaml` tag as field names instead of using struct fields
  - [X] Use `conf` tag to add extra information to documentation files (possible options: `required`, `optional`, `defaultValue`).
- [X] Add required `conf` struct tags to all `conf.go` files in the project.
- [X] Generate docs from indexed configuration structs and write them to `docs/modules/configuration/partials/*`.
- [X] Combine all partial configuration docs into one.
- [X] Change comment lines of `Conf` structs to make them more suitable for documentation.

#### Checklist 

- [X] All commits are signed-off (`git commit -s ...`) to provide the [DCO](https://developercertificate.org/)


charithe (commented):
Great job! It's really cool! 👍🏽 

Couple of nits:

- Let's try to make `hack/tools/confdocs` a Go module of its own so that its dependencies don't become dependencies of the main Cerbos project. This might make it tricky to use as a `go:generate` command but I'll come back to that in a bit.
- Let's make logging optional and only let it be enabled if an environment variable is set. Otherwise `make docs` looks very noisy
- I was a bit surprised that it was quite slow to run. Then I realised that the program gets invoked multiple times by `go generate` and it scans the entire project each time! I think we can optimize it by running the generator just once as a `make command` (unless I am missing something very obvious; let's discuss that further).

oguzhand95:
Thanks for the review!

- I will take a look at separation later because I couldn't figure out how to load packages from another go module yet.
- Checking the env parameter `CONFDOCS_LOG_LEVEL` now, and if it is not supplied the default is to only log errors. I re-used the `doInitLogging` function from the project.
- I have fixed the comments so that they conform to the Go convention.
- Yes, it takes a bit of time to generate documentation. Currently we have two possible ways to run; 
  - If we directly run the `confdocs.go`, it loads all the packages from `internal` directory and generates the docs from there. The problem with this approach is, a struct called `CompilationUnit` is also implements the `Section` interface hence we see the `.adoc` file for it too. 
  - If we run `go generate ./...` in the project root, each execution of the confdocs only load packages from the related sub-folder. For example, for `audit` the confdocs only load the packages under `internal/audit`. That's why we do not see the side effect I mentioned above.
 
After discussing the performance of the tool offline with you, by your advise I changed what to load and cut down unnecessary ones. The required `Mode`s are;

```
packages.NeedTypes | packages.NeedTypesInfo | packages.NeedFiles | packages.NeedSyntax | packages.NeedName,
```

It made the tool work much faster than before. But, if we are going to use the second method (`go generate ./...`) I think it still lacks some performance.

I had to get creative (:D) in some parts of the code while trying to generate the documentation from the code without modifying the project structure or code. If you feel there are better ways to do some parts, I would gladly try to adapt the into it.

charithe:
If there are things that we can do (without breaking external compatibility) to make this easier, I am totally open to doing them. 😃 
For example, we could rename `Key()` to `ConfKey()` in the `Section` interface to avoid false positives.

docs/build.sh
```diff
@@ -9,6 +9,15 @@ SOURCE_DIR=$(cd "${SCRIPT_DIR}/.." && pwd)
 WORKSPACE="/github/workspace"
 ANTORA_VERSION=${ANTORA_VERSION:-"3.0.0-alpha.10"}
 
+PARTIALS_DIR="${SOURCE_DIR}/docs/modules/configuration/partials"
+FULL_CONFIGURATION_DOC="${PARTIALS_DIR}"/fullconfiguration.adoc
+rm -f "${FULL_CONFIGURATION_DOC}"
+for path in "${PARTIALS_DIR}"/conf*.adoc; do
+    file=$(basename "${path}")
+    content=$(printf 'include::partial$%s[]' "$file")
```
dbuduev, line 17: Is it possible to direct `printf` output to "${FULL_CONFIGURATION_DOC}"?

internal/observability/tracing/conf.go
```diff
@@ -20,25 +20,25 @@ var (
 	errJaegerEndpointUndefined = errors.New("jaeger endpoint undefined")
 )
 
-// Conf holds the tracing configuration.
+// Conf is optional configuration for tracing.
 type Conf struct {
 	// SampleProbability is the probability of sampling expressed as a number between 0 and 1.
-	SampleProbability float64 `yaml:"sampleProbability"`
+	SampleProbability float64 `yaml:"sampleProbability" conf:",defaultValue=0.1"`
```
charithe, line 26: I believe the default is 0

docs/build.sh
```diff
@@ -9,6 +9,13 @@ SOURCE_DIR=$(cd "${SCRIPT_DIR}/.." && pwd)
 WORKSPACE="/github/workspace"
 ANTORA_VERSION=${ANTORA_VERSION:-"3.0.0-alpha.10"}
 
+PARTIALS_DIR="${SOURCE_DIR}/docs/modules/configuration/partials"
+FULL_CONFIGURATION_DOC="${PARTIALS_DIR}"/fullconfiguration.adoc
+rm -f "${FULL_CONFIGURATION_DOC}"
+for path in "${PARTIALS_DIR}"/conf*.adoc; do
+    printf 'include::partial$%s[]' "$(basename "${path}")" >> "${FULL_CONFIGURATION_DOC}"
+done
+
```
charithe, line 18: This not required any more, right?

charithe (commented):
One final thing: let's rename the `defaultValue` struct tag to `example` because that's what it's being used for. I was confused while reading the code because the default values didn't seem correct to me. Then I realised they were infact examples.

docs/build.sh
oguzhand95: Yeap, got rid of it.

charithe (approved):
LGTM. Nice work 👍🏽

PR #516, opened 2021-12-21
enhancement(refactor)!: Refactor policy Admin API endpoints

### LINKED ISSUE #487: Refactor Admin API
`ListPolicies` returns all the policies and their contents in full -- which is not efficient and can cause resource usage peaks that are hard to predict in advance. We should instead follow the recent pattern we introduced in `ListSchemas` where the response returns only the IDs of the resources. Fetching the full definition should require a second API call to `GetPolicy`.

oguzhand95:
#### Description

Fixes #487 

#### Checklist 


api/public/cerbos/request/v1/request.proto
```diff
@@ -476,6 +476,22 @@ message ListPoliciesRequest {
     SortOptions sort_options = 2;
 }
 
+message GetPolicyRequest {
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
+        json_schema: {
+            description: "Get policy(s) request"
```
charithe, line 482:
```suggestion
            description: "Get policy request"
```

api/public/cerbos/request/v1/request.proto
```diff
@@ -476,6 +476,22 @@ message ListPoliciesRequest {
...
+        }
+    };
+    repeated string id = 1 [
+        (google.api.field_behavior) = REQUIRED,
+        (validate.rules).repeated = { unique: true, min_items: 1, max_items: 25, items { string { min_len: 1, max_len: 255} } },
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            description: "id (fully qualified name) for the policy"
+            example: "\"\""
```
charithe, line 490: Better include an example here

api/public/cerbos/response/v1/response.proto
```diff
@@ -260,6 +260,16 @@ message ListPoliciesResponse {
 		}
     };
 
+    repeated string policy_ids = 1;
+}
+
+message GetPolicyResponse {
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
+        json_schema: {
+            description: "Get policy(s) response"
```
charithe, line 269:
```suggestion
            description: "Get policy response"
```

client/admin.go
```diff
@@ -181,10 +181,13 @@ func (c *GrpcAdminClient) ListPolicies(ctx context.Context, opts ...ListOpt) ([]
...
+	/*
+		TODO(oguzhan): Take a look at this
+		pc, err := c.client.ListPolicies(ctx, req, grpc.PerRPCCredentials(c.creds))
+		if err != nil {
+			return nil, fmt.Errorf("could not list policies: %w", err)
+		}
+	*/
 
-	return pc.Policies, nil
+	return nil, nil
```
charithe, line 192: ?

client/admin_test.go
```diff
@@ -116,9 +116,12 @@ func TestListPolicies(t *testing.T) {
...
-		require.NotEmpty(t, policies)
-	})
+	// TODO(oguzhan): Take a look at this
+	/*
+		t.Run("should get the list of policies", func(t *testing.T) {
+			policies, err := ac.ListPolicies(context.Background())
+			require.NoError(t, err)
+			require.NotEmpty(t, policies)
+		})
+	*/
```
charithe, line 126: ?

internal/storage/db/internal/db.go
```diff
@@ -349,25 +364,24 @@ func (s *dbStorage) Delete(ctx context.Context, ids ...namer.ModuleID) error {
 	return nil
 }
 
-func (s *dbStorage) GetPolicies(ctx context.Context) ([]*policy.Wrapper, error) {
-	res, err := s.db.From(PolicyTbl).Executor().ScannerContext(ctx)
+func (s *dbStorage) ListPolicyIDs(ctx context.Context) ([]string, error) {
+	res, err := s.db.Select(goqu.C(PolicyTblIDCol)).From(PolicyTbl).Executor().ScannerContext(ctx)
```
charithe, line 368: Shouldn't this be FQN or `name` + `version`?

internal/storage/db/mysql/schema.sql
```diff
@@ -7,6 +7,7 @@ CREATE TABLE IF NOT EXISTS policy (
     kind VARCHAR(128) NOT NULL,
     name VARCHAR(1024) NOT NULL,
     version VARCHAR(128) NOT NULL,
+    fqn VARCHAR(255) NOT NULL,
```
charithe, line 10: FQN is the combination of `cerbos` + `kind` + `name` + `version`.  So, given that the name column is 1024 chars, FQN should be more than that. 

I think that instead of the FQN, we should let the human readable identifier be the tuple (`kind`, `name`, `version`). That saves the users from being exposed to internal implementation details and work with the concrete data they deal with in the policy files.

internal/storage/index/builder.go
```diff
@@ -136,6 +136,8 @@ func Build(ctx context.Context, fsys fs.FS, opts ...BuildOpt) (Index, error) {
 type indexBuilder struct {
 	executables  map[namer.ModuleID]struct{}
 	modIDToFile  map[namer.ModuleID]string
+	modIDTOFQN   map[namer.ModuleID]string
+	FQNToModID   map[string]namer.ModuleID
```
charithe, line 140: This is unnecessary because it's easy to calculate the module ID given the FQN. Let's not use up memory to store something that can be easily calculated.

charithe (commented):
I think that instead of the FQN, we should let the human readable identifier be the tuple `(kind, name, version)` which can be represented as a dot separated string (`resource.leave_request.default`). That shields the users from being exposed to internal implementation details and allow them to work with the concrete data they deal with in the policy files. 

Also, I believe the `cerbosctl` commands need to be updated to handle this change as well. I think we are now losing the ability to filter and sort policies by fields.

client/admin.go
```diff
@@ -155,36 +155,42 @@ func (c *GrpcAdminClient) auditLogs(ctx context.Context, opts AuditLogOptions) (
...
+	if err := getReq.Validate(); err != nil {
+		return nil, fmt.Errorf("could not validate get policy request: %w", err)
+	}
+
+	getRes, err := c.client.GetPolicy(ctx, getReq, grpc.PerRPCCredentials(c.creds))
+	if err != nil {
+		return nil, fmt.Errorf("could not get policy: %w", err)
+	}
+
+	return getRes.Policies, nil
```
charithe, line 195: Is this guaranteed to work? If `ListPolicies` returns 100 policies, then `GetPolicy`  would fail because the limit is 25. 
Furthermore, the client should correspond to the actual API methods exposed by the server. If the `ListPolicies` API call only returns the list of policy IDs, then the client's `ListPolicies` method should return a list of policy IDs as well.

internal/storage/blob/store.go
```diff
@@ -322,3 +322,7 @@ func (s *Store) ListSchemaIDs(ctx context.Context) ([]string, error) {
 func (s *Store) LoadSchema(ctx context.Context, url string) (io.ReadCloser, error) {
 	return s.idx.LoadSchema(ctx, url)
 }
+
+func (s *Store) LoadPolicy(ctx context.Context, file string) (*policy.Wrapper, error) {
```
charithe, line 326: We can pass a list of files here instead of just one. For disk stores it might not make much difference whether we retrieve policies one by one or as a list. But, for network stores like blob or database, it's almost always much more efficient to retrieve a batch at once.

internal/storage/db/internal/db.go
```diff
@@ -24,17 +24,23 @@ import (
...
 	GetCompilationUnits(ctx context.Context, ids ...namer.ModuleID) (map[namer.ModuleID]*policy.CompilationUnit, error)
 	GetDependents(ctx context.Context, ids ...namer.ModuleID) (map[namer.ModuleID][]namer.ModuleID, error)
 	Delete(ctx context.Context, ids ...namer.ModuleID) error
-	GetPolicies(ctx context.Context) ([]*policy.Wrapper, error)
+	ListPolicyIDs(ctx context.Context) ([]string, error)
 	ListSchemaIDs(ctx context.Context) ([]string, error)
 	AddOrUpdateSchema(ctx context.Context, schemas ...*schemav1.Schema) error
 	DeleteSchema(ctx context.Context, ids ...string) error
 	LoadSchema(ctx context.Context, url string) (io.ReadCloser, error)
+	LoadPolicy(ctx context.Context, policyKey string) (*policy.Wrapper, error)
```
charithe, line 43: See my comment above about batch retrieval

internal/storage/db/internal/db.go
```diff
@@ -122,6 +128,33 @@ func (s *dbStorage) DeleteSchema(ctx context.Context, ids ...string) error {
...
+func (s *dbStorage) LoadPolicy(ctx context.Context, policyKey string) (*policy.Wrapper, error) {
+	knv := strings.Split(policyKey, ".")
+	if len(knv) < policyKeyMinSubItem || len(knv) > policyKeyMaxSubItem {
+		return nil, fmt.Errorf("invalid format for policyKey")
+	}
+
+	version := ""
+	if len(knv) == policyKeyMaxSubItem {
+		version = knv[2]
+	}
```
charithe, line 140: Why do we need to handle the case where `version` is not part of the policy key? Also, if the defaultversion is empty string, would it match anything?

internal/storage/db/internal/db.go
```diff
@@ -349,25 +382,29 @@ func (s *dbStorage) Delete(ctx context.Context, ids ...namer.ModuleID) error {
 	return nil
 }
 
-func (s *dbStorage) GetPolicies(ctx context.Context) ([]*policy.Wrapper, error) {
-	res, err := s.db.From(PolicyTbl).Executor().ScannerContext(ctx)
+func (s *dbStorage) ListPolicyIDs(ctx context.Context) ([]string, error) {
+	res, err := s.db.Select(goqu.C(PolicyTblKindCol), goqu.C(PolicyTblNameCol), goqu.C(PolicyTblVerCol)).
```
charithe, line 386: Might be worth creating a composite index or a computed column for this in the database. That would help with efficient batch retrieval from `GetPolicy` calls as well.

internal/storage/db/internal/db.go
```diff
@@ -349,25 +382,29 @@ func (s *dbStorage) Delete(ctx context.Context, ids ...namer.ModuleID) error {
...
 			return nil, fmt.Errorf("could not scan row: %w", err)
 		}
 
-		p := policy.Wrap(rec.Definition.Policy)
-		policies = append(policies, &p)
+		policyKey := fmt.Sprintf("%s.%s", strings.ToLower(rec.Kind), rec.Name)
+		if rec.Version != "" {
+			policyKey += fmt.Sprintf(".%s", rec.Version)
+		}
+		policyKeys = append(policyKeys, policyKey)
```
charithe, line 404: I think we should be able to do the concatenation in SQL itself and remove the extra work done here to scan the records into a policy struct and then concatenate the fields. It'd be much more efficient to scan the result set into a string array directly and it won't generate as much garbage for the GC to deal with either.

internal/svc/admin_svc.go
```diff
@@ -104,6 +105,52 @@ func (cas *CerbosAdminService) AddOrUpdateSchema(ctx context.Context, req *reque
...
+	log := ctxzap.Extract(ctx)
+	policies := make([]*policyv1.Policy, 0, len(req.Id))
+	for _, id := range req.Id {
+		p, err := cas.store.LoadPolicy(ctx, id)
+		if err != nil {
+			log.Error(fmt.Sprintf("Could not get the policy with id %s", id), zap.Error(err))
+			return nil, status.Errorf(codes.Internal, "could not get the policy with id %s", id)
+		}
+		policies = append(policies, p.Policy)
+	}
```
charithe, line 147: When you know the size of the slice in advance, it's slightly faster to use indexing rather than `append`. I would rewrite it as follows:

```suggestion
	policies := make([]*policyv1.Policy, len(req.Id))
	for i, id := range req.Id {
		p, err := cas.store.LoadPolicy(ctx, id)
		if err != nil {
			log.Error(fmt.Sprintf("Could not get the policy with id %s", id), zap.Error(err))
			return nil, status.Errorf(codes.Internal, "could not get the policy with id %s", id)
		}
		policies[i] =  p.Policy
	}
```

There are some other places above where this can be done as well.

```
BenchmarkArrayAppend
BenchmarkArrayAppend/append
BenchmarkArrayAppend/append-8         	23759680	       43.08 ns/op	     23 B/op	      0 allocs/op
BenchmarkArrayAppend/index
BenchmarkArrayAppend/index-8          	32172147	       41.67 ns/op	     23 B/op	      0 allocs/op
```

internal/storage/index/index.go
```diff
@@ -329,22 +331,17 @@ func (idx *index) Inspect() map[string]meta {
...
-		if err != nil {
-			return nil, err
-		}
-
-		wp := policy.Wrap(pol)
-		entries = append(entries, &wp)
+	files := make([]string, 0, len(idx.modIDToFile))
+	for _, f := range idx.modIDToFile {
+		files = append(files, f)
 	}
```
charithe, line 341: Better to use indexing rather than append here. See my comment below.

internal/storage/db/internal/db.go
charithe: Annoyingly, computed column would require a database migration step. So let's not do that. Creating a composite index would be fine though. If it doesn't exist, the query would still work, albeit a little slower.

internal/storage/db/internal/db.go
charithe: I forgot about derived roles. My bad.

oguzhand95: Yes, unfortunately derived roles doesn't have version field :( They caused some headaches with this implementation and also with the new implementation.

oguzhand95:
The table of support for concatenation using ||  and CONCAT()  look like this;
```
SQLite MySQL PostgresSQL
|   -   |   +   |   +   | CONCAT()
|   +   |   -   |   +   | ||
```

MySQL documentation suggests executing the following makes MySQL recognize ||  as concatenation operator. (By default, MySQL treats ||  as OR.)
`SET sql_mode = 'PIPES_AS_CONCAT';`

Let's discuss whether this could be an option or not.
One other option may be using `goqu` to handle the case.

client/admin.go
```diff
@@ -180,21 +168,25 @@ func (c *GrpcAdminClient) ListPolicies(ctx context.Context, opts ...ListOpt) ([]
...
+		Id: ids,
 	}
-	if err := getReq.Validate(); err != nil {
-		return nil, fmt.Errorf("could not validate get policy request: %w", err)
+	if err := req.Validate(); err != nil {
+		valErr := new(requestv1.GetPolicyRequestValidationError)
+		ok := errors.As(err, &valErr)
+		if !ok || valErr.Reason() != errMsgItemsLen {
+			return nil, fmt.Errorf("could not validate get policy request: %w", err)
+		}
```
charithe, line 183: Why check for this specific error? If `Validate` returns an error, that means the request is invalid and we just need to return that error back to the user.

cmd/cerbosctl/list/list.go
```diff
@@ -35,15 +37,21 @@ func NewListCmd(fn internal.WithClient) *cobra.Command {
...
+	policies := make([]*policy.Policy, 0, len(policyIds))
+	for _, chunk := range chunks(policyIds, maxPolicyPerReq) {
+		var p []*policy.Policy
+		p, err = c.GetPolicy(context.Background(), chunk...)
+		if err != nil {
+			return fmt.Errorf("error while requesting policy: %w", err)
+		}
+		policies = append(policies, p...)
+	}
+
```
charithe, line 54: How about just iterating through the ID list and whenever you hit an index divisible by 25, pass the slice range `policyIDs[idx-25:idx]` to  `GetPolicyList` and then immediately printing the batch you  get from the server? That way you don't need to store all policies in memory (which could be quite a lot of them).

internal/storage/db/internal/db.go
```diff
@@ -383,25 +376,19 @@ func (s *dbStorage) Delete(ctx context.Context, ids ...namer.ModuleID) error {
...
 		return nil, fmt.Errorf("could not execute %q query: %w", "ListPolicyIDs", err)
 	}
 	defer res.Close()
 
 	var policyKeys []string
 	for res.Next() {
-		var rec Policy
-		if err := res.ScanStruct(&rec); err != nil {
+		var rec string
+		if err = res.ScanVal(&rec); err != nil {
```
charithe, line 388: I think you can use `ScanVals` to directly scan the whole resultset into the slice without iterating through the result set one by one.

charithe:
It's a shame that MySQL doesn't conform to ANSI SQL. We'll have to figure out a way to generate the correct `CONCAT` SQL based on the database dialect. Goqu should have a way of doing that but it's not very obvious. I'll do a bit more digging and discuss the options with you.

cmd/cerbosctl/list/list.go
oguzhand95: I think this wouldn't cover the scenario where there are less than 25 policy ids in the slice if I am not misunderstanding.
I came up with the following instead;
```
for idx := range policyIds {
    if idx%maxPolicyPerReq == 0 {
        var p []*policy.Policy
	p, err = c.GetPolicy(context.Background(), policyIds[idx:minInt(idx+maxPolicyPerReq, len(policyIds)-idx)]...)
	if err != nil {
	    return fmt.Errorf("error while requesting policy: %w", err)
	}
	if err = printPolicies(cmd.OutOrStdout(), p, listPoliciesFlags.OutputFormat()); err != nil {
	    return fmt.Errorf("could not print policies: %w", err)
        }
    }
}
```

charithe: Good point.

internal/storage/db/postgres/schema.sql
```diff
@@ -12,6 +12,8 @@ CREATE TABLE IF NOT EXISTS policy (
     definition BYTEA
 );
 
+CREATE INDEX policy_knv_idx ON policy ((lower(kind) || '.' || name || '.' || version));
```
charithe, line 15:
```suggestion
CREATE INDEX IF NOT EXISTS policy_knv_idx ON policy ((lower(kind) || '.' || name || '.' || version));
```

internal/svc/list.go
```diff
@@ -4,167 +4,13 @@
...
-	}
-}
-
 func sortSchemas(schemasIds []string) {
 	sort.Strings(schemasIds)
 }
+
+func sortPolicies(ids []string) {
+	sort.Strings(ids)
+}
```
charithe, line 16: I think we can simply inline these two functions and get rid of `list.go` completely.

charithe (commented):
LGTM 👍🏽  Just a couple of small nits to address. 
Will do a bit of testing tomorrow morning and we can merge this afterwards.

internal/storage/index/index.go
charithe: Ah, it's a map. We could have a manually incremented index variable but I am not too bothered about it.

charithe (approved):
LGTM 👍🏽

PR #536, opened 2022-01-04
feat!: Replace `cerbosctl list` with `cerbosctl get` 

### LINKED ISSUE #532: Policy listing and viewing with cerbosctl
Follow up from #516 

- Allow `cerbosctl list` to be filtered (by kind, version or name) 
- Add `cerbosctl get` to retrieve a policy by ID

oguzhand95:
This PR adds the following command to `cerbosctl`;
```
# List derived_roles
cerbosctl get derived_roles
cerbosctl get derived_role
cerbosctl get dr

# List and filter derived_roles
cerbosctl get derived_roles --name my_derived_roles

# List principal policies
cerbosctl get principal_policy
cerbosctl get principal_policies
cerbosctl get pp

# List and filter principal policies
cerbosctl get principal_policies --name donald_duck --version default

# List resource policies
cerbosctl get resource_policy
cerbosctl get resource_policies
cerbosctl get rp

# List and filter resource policies
cerbosctl get resource_policies --name leave_request --version default

# List all schemas
cerbosctl get schema
cerbosctl get schemas
cerbosctl get s

# Get policy
cerbosctl get policy derived_roles.my_derived_roles -oyaml

# Get policy (disk, git, blob stores)
cerbosctl get policy blog_derived_roles.yaml principal.yaml

# Get policy (mutable stores)
cerbosctl get policy derived_roles.my_derived_roles
```

Fixes #532


charithe (commented):
Unfortunately, I think the CLI is a bit inconsistent now. I should have paid more attention earlier. 

- `list` and `get` both retrieve the full policy definitions from the server which kind of defeats the purpose. The reason we introduced the List API call is to avoid fetching all policies in the first place. 
- `list` prints a table of formatted policy names that the user  can't directly use in the `get` command. They have to know how to construct the policy ID by hand. It gets even more complicated when the backing store is disk-based because the ID is the file name. 

So, I think that we need to refactor the commands to do the following:
- `list` returns the result of the `ListPolicy` API. 
   - Obviously we can't filter this list by kind,name or version. So the only filter should be a regex   filter
   - There's not much point in supporting `json` or `yaml` formats for list output
- `get` returns the output of the `GetPolicy` API
   - In addition to `json` and `yaml` we should support a `pretty-json` format to output the policies as nicely formatted JSON

cmd/cerbosctl/internal/print.go
```diff
@@ -0,0 +1,125 @@
...
+		var prettyJSON bytes.Buffer
+		err = json.Indent(&prettyJSON, b, "", prettyPrintIndent)
+		if err != nil {
+			return fmt.Errorf("could not indent policy: %w", err)
+		}
+
+		_, err = fmt.Fprintf(w, "%s\n", prettyJSON.String())
+		if err != nil {
+			return fmt.Errorf("failed to print policy: %w", err)
+		}
```
charithe, line 107: I think you can do `protojson.Format` to get formatted JSON

docs/modules/cli/pages/ctl.adoc
```diff
@@ -129,8 +129,9 @@ cerbosctl decisions --tail=20
 
 This command lists the policies available in the configured policy repository.
 
-NOTE: Possible values for `kind` are `derived_roles`, `resource` and `principal`
-NOTE: Multiple values seperated by comma may be stated for `kind`, `name` and `version`.
+You can filter the output using the `kind`, `name` and `version` flags. Each flag accepts multiple comma-separated values which are OR'ed together. For example, `--kind=resource,principal` matches policies that are either resource policies or principal policies.
```
charithe, line 132: There's no `--kind` flag anymore, is there?

cmd/cerbosctl/get/get.go
```diff
@@ -4,50 +4,150 @@
...
-		RunE:  fn(runListCmdF),
+		Use:     "get",
+		Short:   "Get",
+		Example: getExample,
+		RunE:    fn(runGetCmd),
 	}
 
-	cmd.Flags().AddFlagSet(getPolicyFlags.FlagSet())
+	cmd.Flags().BoolVar(&getArgs.NoHeaders, "no-headers", false, "Print no headers")
+	cmd.Flags().StringVarP(&getArgs.Output, "output", "o", "json", "Output format for the policies; json, yaml, prettyjson formats are supported")
```
charithe, line 73: Let's make the default YAML because that's what users usually deal with
```suggestion
	cmd.Flags().StringVarP(&getArgs.Output, "output", "o", "yaml", "Output format for the policies; json, yaml, prettyjson formats are supported")
```

cmd/cerbosctl/get/policy.go
```diff
@@ -0,0 +1,119 @@
...
+		if idx%internal.MaxIDPerReq == 0 {
+			idxEnd := internal.MinInt(idx+internal.MaxIDPerReq, len(policyIds))
+			policies, err := c.GetPolicy(context.Background(), policyIds[idx:idxEnd]...)
+			if err != nil {
+				return fmt.Errorf("error while requesting policy: %w", err)
+			}
+
+			filtered := filterPolicies(policies, policyIds[idx:idxEnd], getArgs.Name, getArgs.Version, resType)
+
+			err = internal.PrintIds(cmd.OutOrStdout(), filtered...)
```
charithe, line 43: since we have already retrieved the policies, why not also print out the name and version in addition to the ID

charithe (commented):
Nice work! We are very close.
I think that we should make `resource_policies`, `principal_policies`, `derived_roles` and `schemas` sub commands of the `get` command. That way Cobra can validate the commands properly without us having to write code for that,  we can add the aliases (e.g. `rp`, `resource_policies`) using the built-in features of Cobra, and also have tighter control over the flags each sub command accepts. (For example, right now `get dr derived_roles/derived_roles_01.yaml --name=beta,alpha` doesn't complain about the `--name` flag. If it was a subcommand, I think it would work as expected.)

charithe (commented):
Very clean implementation 👏🏽 
Just a couple of small nits and cosmetic changes to address before we can merge this.

cmd/cerbosctl/get/schema/schema.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/cerbos/cerbos/cmd/cerbosctl/internal"
+)
+
+const example = `# List schemas
+cerbosctl get schemas
+cerbosctl get schema
+cerbosctl get s
+
+# Get schema definition
+cerbosctl get schemas principal.yaml`
```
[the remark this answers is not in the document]
oguzhand95: oops :)

charithe (approved):
LGTM. A really clean implementation 🎖️

Issue #147, opened 2021-06-13
E2E Tests


charithe:
Design an E2E framework to run Cerbos through production-like scenarios in every architecture, OS, storage engines and other possible permutations of configurations we support. Initially it could be a manual step in the release checklist but eventually it should be a nightly or weekly job.

charithe:
Current tests exercise the following:
- x86 arch (the test suite has been run on ARM64 as a one off but it is not regularly run as part of CI)
- Kubernetes service deployment using Helm
- All currently supported storage drivers (blob, git, MySQL, Postgres, SQLite). Blob tests only run on Minio and not S3 and GCS. 
- Audit logs on
- Schemas on
- TLS on
- HTTP and gRPC

charithe:
There's no good solution for tests that require secrets (e.g. credentials to access a GitHub policy repo, credentials for actual GCS, S3 etc.). Although we can use CI secrets, that makes it painful for developers to locally run the tests.

Currently the tests that require secrets are the git and blob stores. The blob store should be reasonably covered by the Minio S3-compatibility guarantee. With Git, we use passwordless checkout from GitHub at the moment. For testing the https and ssh auth methods, one possible solution might be to spin-up [Gogs](https://gogs.io) or [Gitea](https://gitea.io/en-us/) to simulate the cloud providers. The Helm charts for those products don't seem to provide an easy way to automatically provision the repositories though. They seem to require manual configuration using the UI.

PR #544, opened 2022-01-10
chore(test): Add end-to-end test suite

charithe:
This PR adds a set of E2E tests that verify the Cerbos deployment process and exercises the different storage backends. I discovered two bugs whilst working on it so the work has already paid off in some way.

This is purely foundational work with lots of room for improvement in the future. However, it should be good enough for our immediate requirements of increasing the confidence on new releases.

CI workflow will be added in a separate PR.

Part of #147


charithe:
Proto check failure is due to `buf.lock` in `main` having references to versions that seem to have been purged from BSR. It should  be fixed after the updated `buf.lock` file in this PR is merged.

e2e/blob/helmfile.yaml
```diff
@@ -0,0 +1,121 @@
...
+      - auth:
+          rootUser: admin
+          rootPassword: passw0rd
+      - defaultBuckets: "cerbos"
+      - persistence:
+          enabled": false
+
+  - name: cerbos
+    namespace: '{{ requiredEnv "E2E_NS" }}'
+    needs: ["postgres"]
```
dbuduev, line 46:
```suggestion
    needs: ["minio"]
```

dbuduev (approved):
LGTM 👍🏻 

> telepresence: error: connector.Connect: failed to start traffic manager: the helm operation timed out.

I ran into this problem though. It might be due to Apple silicon.

charithe:
I ran into that problem too. Had to increase the timeout:

```sh
cat ~/.config/telepresence/config.yml 
timeouts:
  helm: 60s
```

dbuduev:
Thank you! All good now.

I had tried to increase the timeout but misplaced the config.

PR #555, opened 2022-01-12
enhancement: Query planner to return a detailed error

dbuduev:
#### Description

if principal.attr in the query plan request is missing an attribute that is referenced by the policy, the call fails with {"code":13, "message":"Resources query plan request failed"}.
Reason is "error evaluating condition \"request.principal.attr.department == \\\"Sales\\\"\": no such key: department".
This PR detecs this particular case and produces a detailed error message.

#### Checklist 


internal/svc/cerbos_svc.go
```diff
@@ -61,6 +61,9 @@ func (cs *CerbosService) ResourcesQueryPlan(ctx context.Context, request *reques
 	response, err := cs.eng.ResourcesQueryPlan(logging.ToContext(ctx, log), input)
 	if err != nil {
 		log.Error("Resources query plan request failed", zap.Error(err))
+		if errors.Is(err, engine.ErrNoSuchKey) {
+			return nil, status.Errorf(codes.Internal, err.Error())
```
charithe, line 65: In this case the request is bad, so we should return a 4xx status

```suggestion
			return nil, status.Errorf(codes.InvalidArgument, "Bad request: %v", err)
```

internal/engine/planner.go
```diff
@@ -419,6 +423,9 @@ func evaluateCELExprPartially(expr *exprpb.CheckedExpr, input *enginev1.Resource
 
 	val, details, err := prg.Eval(vars)
 	if err != nil {
+		if strings.HasPrefix(err.Error(), "no such key:") {
+			return nil, nil, fmt.Errorf("%w%s", ErrNoSuchKey, err)
+		}
```
charithe, line 428:
```suggestion
		if errStr := err.Error(); strings.HasPrefix(errStr, "no such key:") {
			return nil, nil, fmt.Errorf("missing required attribute %q%w", strings.TrimPrefix(errStr, "no such key:"), ErrNoSuchKey)
		}
```

charithe (commented):
LGTM. Just added some suggestions to make the error message a bit more friendlier.

charithe:
Hmm.. the error message actually includes the condition that failed to evaluate. I don't think we should output that in the response because it could contain sensitive information.

```
{"code":13, "message":"error evaluating condition \"request.principal.attr.department == \\\"Sales\\\"\": missing required attribute: \" department\""}
```

internal/engine/planner.go
```diff
@@ -424,7 +428,7 @@ func evaluateCELExprPartially(expr *exprpb.CheckedExpr, input *enginev1.Resource
 	val, details, err := prg.Eval(vars)
 	if err != nil {
 		if strings.HasPrefix(err.Error(), "no such key:") {
-			return nil, nil, fmt.Errorf("%w%s", ErrNoSuchKey, err)
+			return nil, nil, &NoSuchKeyError{msg: err.Error()}
```
charithe, line 431: The error message is `{"code":3,"message":"Bad request: no such key: department"}` which is not very helpful. So let's make it a bit more friendlier.

```suggestion
			return nil, nil, &NoSuchKeyError{msg: fmt.Sprintf("missing principal attribute "%q", strings.TrimPrefix( err.Error(), "no such key: "))}
```

charithe:
Let's also sneak in a fix to this PR to produce an error message when the requested policy does not exist in the store. Currently the output does not provide any indication that the policy is missing and produces something like `{"requestId":"c2db17b8-4f9f-4fb1-acfd-9162a02be42b","action":"read","resourceKind":"contact","policyVersion":"default"}`

PR #557, opened 2022-01-12
chore(ci): Enable server tests in CI

charithe:
Server tests didn't work inside GH Actions before. Checking to see whether that's still the case.


charithe:
Apparently they work now! 🎉

PR #562, opened 2022-01-14
feat: Enable switching off schema validation for certain actions

charithe:
Allows exclusions to be defined so that schema validation can be
switched off for actions that are in the exclusion list. This is useful
for special cases like `CREATE` where the resource might not contain all
the required fields.

```yaml
schemaRefs:
  principalSchema:
    ref: cerbos:///complex_object.json
    ignoreWhen:
      actions: ["view:*"]
  resourceSchema:
    ref: cerbos:///complex_object.json
    ignoreWhen:
      actions: ["view:public"]
```


dbuduev (approved):
LGTM 👍🏻

PR #574, opened 2022-01-17
chore(test): Wait for server to be ready before tests

charithe:
GitHub Actions runners don't seem to be fast enough to bring up the test
servers quickly enough so this change adds explicit checks that wait for
the servers to become ready before running any tests that rely on them.


oguzhand95 (approved):
LGTM

PR #575, opened 2022-01-18
feat: Add SQL Server support

dbuduev:

#### Description

Adds SQL Server as a cerbos policies storage.

#### TODO

- [ ] Manual testing
- [ ] E2E test
- [x] Documentation
 
#### Checklist 

### DIFF
--- a/internal/storage/db/internal/db.go
+++ b/internal/storage/db/internal/db.go
@@ -28,6 +28,7 @@ const policyKeySep = "."
 
 type DBStorage interface {
 	storage.Subscribable
+	NotifySubscribers(events ...storage.Event)
 	AddOrUpdate(ctx context.Context, policies ...policy.Wrapper) error
 	GetCompilationUnits(ctx context.Context, ids ...namer.ModuleID) (map[namer.ModuleID]*policy.CompilationUnit, error)
 	GetDependents(ctx context.Context, ids ...namer.ModuleID) (map[namer.ModuleID][]namer.ModuleID, error)
@@ -47,6 +48,7 @@ func NewDBStorage(ctx context.Context, db *goqu.Database) (DBStorage, error) {
 			return nil, err
 		}
 
+		log.Println("YAHOO!")
 		db.Logger(log)
 	}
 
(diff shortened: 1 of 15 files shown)


internal/storage/db/internal/db.go
```diff
@@ -263,7 +264,7 @@ func (s *dbStorage) GetCompilationUnits(ctx context.Context, ids ...namer.Module
 		Where(
 			goqu.And(
 				goqu.C(PolicyDepTblPolicyIDCol).Table("pd").In(ids),
-				goqu.C(PolicyTblDisabledCol).Table("p").Eq(false),
+				goqu.C(PolicyTblDisabledCol).Table("p").Eq(DialectFalseValue(s.db.Dialect())),
```
charithe, line 267: Any idea why the dialect doesn't handle this conversion automatically? It seems to be configured correctly: https://github.com/doug-martin/goqu/blob/31d438d38230118d377f11cfaad9627aa3a23864/dialect/sqlserver/sqlserver.go#L33

charithe (commented):
Looks OK to me although I am a bit concerned about why the dialect doesn't handle the boolean conversion correctly. 

The driver needs to be imported in the `server` package to register it (https://github.com/cerbos/cerbos/blob/9411f6289e0f6a7d3846e1033627fab6fc876e1f/internal/server/server.go#L63-L79)

internal/storage/db/internal/db.go
dbuduev: Not sure why it wasn't being picked up. It works if we replace `false` with `goqu.V(false)`.

dbuduev:
I've refactored the implementations, updated docs and manually tested, but probably I haven't done enough of end-2-end testing.

charithe (approved):
LGTM. I like the refactoring. It's really clean 👍🏽 

Not sure whether https://github.com/microsoft/mssql-docker/tree/master/linux/sample-helm-chart works. If it does, hopefully the E2E tests won't be too hard. That could be a separate PR or part of this as well.

PR #594, opened 2022-01-28
fix: cerbosctl get subcommands retrieve unmatching policy types

### LINKED ISSUE #593: `cerbosctl get` subcommands retrieve unmatching policy types

### Current Behavior

Given the policies are in the store, the following commands print the policies;
```
cerbosctl get derived_roles principal.donald_duck.default
cerbosctl get principal_policies derived_roles.my_derived_roles
cerbosctl get resource_policies resource_policies.leave_request.default
```
This shouldn't be the case. Users are supposed to use relevant policy types in the get command to retrieve the policies.

### Expected Behavior

Given the policies are in the store, the following commands should return nothing. Because this is invalid.
```
cerbosctl get derived_roles principal.donald_duck.default
cerbosctl get principal_policies derived_roles.my_derived_roles
cerbosctl get resource_policies resource_policies.leave_request.default
```

### Steps To Reproduce

1. Add a principal policy to the store.
2. Try to use `cerbosctl get derived_roles` to retrieve the principal policy, even though the policy type is not derived_roles.
3. Observe the principal policy is returned, while it shouldn't.

### Environment

```markdown
- OS: Mac OS X Big Sur 11.6
- Cerbos version: 0.12.0
- Docker version: 20.10.11
- Go version: 1.17.5 (darwin amd64)
```

### Anything else?

_No response_

oguzhand95:
#### Description

Fixes #593 and adds tests

### DIFF
--- a/cmd/cerbosctl/cerbosctl_test.go
+++ b/cmd/cerbosctl/cerbosctl_test.go
@@ -188,6 +188,28 @@ func testGetCmd(fn internal.WithClient) func(*testing.T) {
 					require.JSONEq(t, string(expected), out.String())
 				}
 			})
+
+			t.Run("invalid policy key type for commands", func(t *testing.T) {
+				testCases := []struct {
+					args    []string
+				}{
+					{strings.Split("derived_roles principal.donald_duck_1.default", " ")},
+					{strings.Split("derived_roles resource.leave_request_1.default", " ")},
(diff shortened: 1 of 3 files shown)


charithe (dismissed):
LGTM. 
I think there's a small optimization that we can make. It could be part of this PR or a separate one : whichever is your preference.

charithe (approved):
👍🏽

PR #597, opened 2022-01-31
chore(util): Update to xxhash v2 and add policy hash functions

charithe:
- Update to the v2 release of xxhash for hashing
- Integrate [protoc-gen-go-hashpb](https://github.com/cerbos/protoc-gen-go-hashpb) and add utility functions to hash policies and other protobufs.


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/aws/aws-sdk-go v1.42.23
 	github.com/bluele/gcache v0.0.2
 	github.com/cespare/xxhash v1.1.0
```
oguzhand95, line 11: Do we still supposed to see `xxhash` module while we are using `xxhash/2`?

oguzhand95: Never mind, I missed we are using the v1 in the tests.

oguzhand95 (approved):

PR #606, opened 2022-01-31
feat: Add --sort-by flag to cerbosctl get subcommands

oguzhand95:
#### Description

Adds `--sort-by` flag to `cerbosctl get derived_roles`, `cerbosctl get resource_policies` and `cerbosctl get principal_policies`. Also refactors the code related to `cerbosctl get`.


cmd/cerbosctl/get/internal/policy/policy.go
```diff
@@ -57,24 +53,22 @@ func List(c client.AdminClient, cmd *cobra.Command, filters *flagset.Filters, fo
 				return fmt.Errorf("error while requesting policy: %w", err)
 			}
 
-			wp := make([]policy.Wrapper, len(policies))
+			keyPolicyPairs := make([]KeyPolicyPair, len(policies))
```
charithe, line 56: It looks like we have to do some extra work on the client side to match-up the "store identifier" to the policy. That feels like a problem with the API to me. When possible, we should strive to make the API pleasant to use. So here's my suggestion:

- Let's add a new field named `store_identifer` to [`Policy.Metadata`](https://github.com/cerbos/cerbos/blob/ff338f84dad041d0676ed9abaf2b39e45eb54719/api/public/cerbos/policy/v1/policy.proto#L30-L33) to hold the store identifier.  
- In the `LoadPolicy` methods of the storage implementations, make sure to populate the `store_identifier` metadata field of the  policy. (See the `policy.WithMetadata` function we use in the [index builder](https://github.com/cerbos/cerbos/blob/ff338f84dad041d0676ed9abaf2b39e45eb54719/internal/storage/index/builder.go#L125) to add the source file information that we later use during compilation)
- After doing the above, the policies returned by the `GetPolicy` Admin API is guaranteed to have the store identifier in the metadata (add tests to make sure of that)
- Now the client side implementation is much simpler because all the required information is contained within the policy object and you don't have to create new structs and copies of the returned list in order to sort/filter by fields.

charithe (commented):
LGTM. Nice work.

I added a comment below about a potential improvement to the Admin API. I think it's worth doing to make it pleasant to use.

oguzhand95:
I have updated the PR with respect to reviews and added a couple of tests to increase coverage. I think adding `store_identifier` made the code much cleaner. Thanks for the detailed steps. Hopefully, the implementation is what you had in mind, otherwise please let me know.

The only question in my mind is where to validate the `--sort-by version` input to the `derived_roles` command. (we should return an error in this case due to derived_roles not having a version field.) I would appreciate your input on this subject @charithe.

cmd/cerbosctl/get/internal/flagset/sort.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (kf SortByFlagType) String() string {
+	return SortByPolicyID.String()
+}
+
+func (kf *SortByFlagType) Set(v string) error {
+	switch v {
+	case SortByPolicyID.String(), SortByName.String(), SortByVersion.String():
+		*kf = SortByFlagType(v)
+		return nil
```
charithe, line 50: Because `SortByValue` is a type alias of `string`, it would be more efficient to simply cast it than call a method on it.
```suggestion
	switch SortByValue(v) {
	case SortByPolicyID, SortByName, SortByVersion:
		*kf = SortByFlagType(v)
		return nil
```

cmd/cerbosctl/get/internal/flagset/sort.go
```diff
@@ -0,0 +1,58 @@
...
+
+const (
+	SortByPolicyID SortByValue = "policyId"
+	SortByName     SortByValue = "name"
+	SortByVersion  SortByValue = "version"
+)
+
+func (sbv SortByValue) String() string {
+	return string(sbv)
+}
```
charithe, line 34: Probably don't need this because `SortByValue` is an alias for `string`

charithe (commented):
👍🏽 

Because `cerbosctl` tests can't exercise all of the storage engines, let's also add checks to the test suites of the storage engines to verify that the `store_identifier` gets populated by all of them.

I can't think of a good solution for `version` when handling derived roles. I think we might have to check for that special case in the `PreRun` function.

charithe (approved):
LGTM 👍🏽

charithe (approved):

PR #623, opened 2022-02-08
enhancement: Allow aux_data to be accessed as auxData

### LINKED ISSUE #504: Inconsistent field naming for AuxData
In the request, users can use `auxData` to send the JWT but in the policies, they have to access it as `request.aux_data`. This is due to a quirk in how Protobufs handle JSON vs how CEL handles protobuf messages. We should try to make it consistent across the board so that users don't get confused.

charithe:
This is an inconsistency stemming from the way the CEL library handles
protobuf field names. It's easy for users to be confused because in the
request the field is named `auxData` but in the CEL expression it can
only be accessed as `aux_data`. This patch fixes that inconsistency.

Fixes #504


dbuduev (approved):
LGTM. 👍🏻

PR #639, opened 2022-02-15
chore: Make log level flag case insensitive

charithe:
With the switch to Kong in #625 the log-level flag became case
sensitive. This was causing the recent E2E failures because the Helm
chart defaults are in upper case.

This patch makes the log-level flag case insensitive again to avoid a
breaking change for existing users.


oguzhand95 (approved):
LGTM

PR #640, opened 2022-02-15
test: Load testing script

oguzhand95:
#### Description

- Added load testing scripts under `hack/loadtest`.
- `hack/loadtest/genres.go` tool is able to generate policy sets according to the given argument.
- Running `./loadtest.sh` in `hack/loadtest` runs the load test with disk (5, 50, 500, 5000 policy sets) and postgres (5, 50, 500, 5000 policy sets) stores and puts the resulting logs in `hack/loadtest/results/<store>_<noOfPolicySets>.log`


charithe (approved):
Nice one 👍🏽

PR #641, opened 2022-02-16
feat: Scoped policies

charithe:
This is the first cut of the scoped policies implementation. We introduce a new field named `scope` to principal and resource policies that can hold a dot-separated "scope" that describes the set of ancestors of the policy.

- If a scope exists  it becomes part of the policy FQN (hence each scoped policy will have a unique module ID). Given a resource policy for `leave_request` with version `staging`, the FQNs look like the following:
    - No scope: `cerbos.resource.leave_request.vstaging`. This preserves existing behaviour. 
    - With scope `acme.corp.cloud`: `cerbos.resource.leave_request.vstaging/acme.corp.cloud`
-  In the check request, users can define the optional scope for the policy. We combine this with the resource kind and the policy version to compute the unique module ID.
- When retrieving the compilation unit using the module ID, the storage engine detects that the policy has a scope attached and automatically fetches the ancestors of the policy as well. For the policy example described above, the compilation unit will fetch the following policies and their dependencies:
    -  `cerbos.resource.leave_request.vstaging/acme.corp.cloud`
    - `cerbos.resource.leave_request.vstaging/acme.corp`
    - `cerbos.resource.leave_request.vstaging/acme`
- The compiler then creates a runnable policy set from all of these policies. Then the engine evaluates each policy in order from most specific (`cerbos.resource.leave_request.vstaging/acme.corp.cloud`) to the least specific (`cerbos.resource.leave_request.vstaging/acme`) until all actions have a decision.
- The first policy that matches wins. That is to say that if both `cerbos.resource.leave_request.vstaging/acme.corp.cloud` and `cerbos.resource.leave_request.vstaging/acme.corp` have a rule that matches the action, the decision from `cerbos.resource.leave_request.vstaging/acme.corp.cloud` is considered to be the final decision.

To do:

- [ ] Update query planner to handle scopes
- [ ] Add more checks to the compiler to detect issues like inconsistent schemas in the tree and unused variables
- [ ] Write documentation and migration guide
- [ ] Add scope to `cerbosctl get` output
- [ ] Add more tests


dbuduev:
`cerbos.resource.leave_request.vstaging` (no scope) could be an ancestor of `cerbos.resource.leave_request.vstaging/acme`. What do you think?

internal/namer/namer.go
```diff
@@ -17,12 +17,17 @@ import (
...
 var invalidIdentifierChars = regexp.MustCompile(`[^\w.]+`)
 
+// validKinds holds valid policy kinds and the minimum number of components that must be contained in the policy key.
+//nolint:gomnd
+var validKinds = map[string]int{"derived_roles": 2, "principal": 3, "resource": 3}
+
 const (
 	DerivedRolesPrefix      = "cerbos.derived_roles"
 	PrincipalPoliciesPrefix = "cerbos.principal"
 	ResourcePoliciesPrefix  = "cerbos.resource"
```
dbuduev, line 27:
```suggestion
	DerivedRolesPrefix      = fqnPrefix + "derived_roles"
	PrincipalPoliciesPrefix = fqnPrefix + "principal"
	ResourcePoliciesPrefix  = fqnPrefix + "resource"
```

dbuduev (approved):
LGTM. 👍🏻 
I get `forcetypeassert` linter errors, but I guess we will resolve that later.
As far as I understand there are not many changes from the evaluator point of view, so I don't anticipate a lot of changes for the query planner either.

charithe:
> `cerbos.resource.leave_request.vstaging` (no scope) can potentially be viewed as a direct ancestor of `cerbos.resource.leave_request.vstaging/acme`. What do you think?

Yes. I struggled a bit with that. The reason I decided not to make the no-scope policy the root of the tree is because I think that users should opt-in to using scopes. That way it's obvious to them that they are using the feature and they can easily reason about the tree.
I am happy to reconsider that, of course. Do you think it would be confusing if no-scope is not the root?

charithe:
I am not getting the lint warnings through `make lint` 🤔 The version of golangci-lint seems to be the latest one too: https://github.com/cerbos/cerbos/blob/915eac33caef3868b035d91877020385862bc662/tools/go.mod#L11

charithe:
Since we are working on a feature branch, I will merge this now to avoid making this PR even bigger. Let's address the remaining tasks and the decision about making no-scope policies the root in subsequent PRs

PR #644, opened 2022-02-17
test: Add missing ancestor tests for scoped policies

charithe:
Add tests for the case where one of the ancestors is missing or disabled.

Also moves the MySQL and Postgres migration scripts to the docs directory to be used in the release documentation.


dbuduev (approved):
LGTM

PR #645, opened 2022-02-18
feat: Update query planner to handle scopes

dbuduev:
Introduce scopes support in the query planner.


dbuduev:
> Just one question: I don't see the short-circuit (if a previous rule had matched, ignore the rest) logic in the principal policy plan. 
If I understand the question it is
https://github.com/dbuduev/cerbos/blob/feature/scoped-policies-query-planner/internal/engine/planner.go#L87

charithe (approved):
Ah OK. Thanks!

PR #647, opened 2022-02-18
docs: Add scoped policies documentation

charithe:
Document how scoped policies work.

### DIFF
--- a/docs/modules/policies/pages/scoped_policies.adoc
+++ b/docs/modules/policies/pages/scoped_policies.adoc
@@ -1,0 +1,80 @@
+include::ROOT:partial$attributes.adoc[]
+
+= Scoped Policies
+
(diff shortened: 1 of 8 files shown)


docs/modules/policies/pages/scoped_policies.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+When processing the above request, the decision flow chart for the Cerbos engine would look like the following:
+
+image::decision_flow.png[]
+
+
+== Working with scoped policies
+
+* The policy without any scope defined is always the base policy. It is the policy used by default if a request does not specify any scope.
+* There must be no gaps in the policy chain. For example, if you define a policy with scope `a.b.c`, then policies with scopes `a.b`, `a`, and no-scope should also exist in the policy repository.
```
dbuduev, line 76: What if the policy `a.b` is disabled (effectively removed) after policy with scope `a.b.c` is created?

charithe: It's an error. Both `a.b` and `a.b.c` must be disabled or the compiler will complain about it.

dbuduev (approved):
LGTM 👍🏻

PR #658, opened 2022-02-21
chore: Run linters

oguzhand95:
#### Description
Fixes linting issues

### DIFF
--- a/internal/storage/db/sqlserver/sqlserver.go
+++ b/internal/storage/db/sqlserver/sqlserver.go
@@ -4,9 +4,16 @@
 package sqlserver
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
+	"io"
+	"path/filepath"
+	"runtime"
+	"strings"
 
 	// Import the mssql driver.
 	_ "github.com/denisenkom/go-mssqldb"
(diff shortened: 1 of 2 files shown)


oguzhand95:
With golangci-lint v1.42.2, the following errors are present;
```
Error: Expected '\t', Found '\n' at internal/storage/db/postgres/postgres.go[line 11,col 1] (gci)
Error: Expected '\t', Found '\n' at internal/storage/db/sqlite3/sqlite3.go[line 12,col 1] (gci)
Error: Expected '\t', Found '\n' at schema/schema.go[line 7,col 1] (gci)
Error: Expected '\t', Found '\n' at internal/server/codec.go[line 11,col 1] (gci)
Error: Expected '\t', Found '\n' at internal/server/server.go[line 39,col 1] (gci)
Error: Expected '\t', Found '\n' at internal/storage/blob/store.go[line 22,col 1] (gci)
Error: Expected '\t', Found '\n' at internal/storage/db/internal/funcs_test.go[line 11,col 1] (gci)
Error: Expected '\t', Found '\n' at internal/storage/db/mysql/mysql.go[line 17,col 1] (gci)
Error: Expected '\t', Found '\n' at internal/storage/db/sqlserver/sqlserver.go[line 21,col 1] (gci)
``
I think these are caused by comments made in those files, I couldn't find a way to make it work.

**Comment by charithe** [comment]
I think that the version of `gci` included in 1.44.2 is a bit buggy. I can make it happy by adding the comment immediately after the import statement.

```go
_ "modernc.org/sqlite" // import sqlite3 driver.
```
But, it seems like `gci` doesn't work well with `gofmt` because when  there are multiple comments, `gofmt` indents the comments so that they all start at the same column. That means adding more than one space between the import statement and the comment and `gci` doesn't like that :(

I guess we'll have to disable `gci` until these issues are fixed.

PR #670, opened 2022-02-23
feat: Add `now` function that returns the current timestamp

### LINKED ISSUE #667: Add function to get current timestamp
Add a CEL function to get the current server timestamp. What would be a good name: `now`, `currentTimestamp` or both?

haines:
#### Description

Fixes #667 

This PR adds a `now` function that returns the current timestamp. `now` returns a consistent timestamp (the time that the expression evaluation started) when invoked repeatedly, similarly to how in PostgreSQL the `now` function always returns the time that the transaction started. This guarantees that `now() == now()`.

I've also refactored `timeSince` so that it computes durations from the same timestamp as `now`, to guarantee that durations are computed consistently within an expression, and that `now().timeSince() == duration("0")`.

#### Caveats 

The drawback to the approach I've taken is that we can't reuse instances of `cel.Program`. To avoid creating a footgun, I've wrapped the typical call sequence

```go
prg, err := env.Program(ast, opts...)
result, details, err := prg.Eval(vars)
```

in a single function `conditions.Eval(env, ast, vars, opts...)`, so that the program instance can't be re-used. However, we do have to remember to use `conditions.Eval` rather than calling `env.Program` directly. I considered encapsulating our usage of `*cel.Env` structs behind an interface to prevent this, but it looked like quite a significant change so I opted not to.


internal/conditions/cerbos_lib.go
```diff
@@ -104,16 +113,6 @@ func (clib cerbosLib) ProgramOptions() []cel.ProgramOption {
 				Binary:   callInStringStringOutBool(clib.inIPAddrRangeFunc),
 			},
 
-			&functions.Overload{
-				Operator: timeSinceFn,
```
haines, line 108: Previously there were two overloads defined here, `timeSince` and `timeSince_timestamp` - I don't think this was necessary because only the latter one was registered in the `decls.NewFunction` call above. I've simplified to just have one overload, registered with the function name (matching the other functions).

internal/conditions/cerbos_lib_benchmarks_test.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2021-2022 Zenauth Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+package conditions
```
haines, line 4: I made `program` private to avoid issues with reusing a program instance, but this meant that I had to move the benchmarks to a whitebox test - they deliberately reuse the program instance in order to only measure the expression evaluation time without the (small) overhead of constructing the program.

internal/conditions/cerbos_lib.go
charithe: Yeah, I think that was a leftover from a botched refactoring. Good catch!

charithe (approved):
LGTM 👍🏽

PR #672, opened 2022-02-23
enhancement: Use camel case for custom functions

### LINKED ISSUE #669: Use camelCase for custom functions
The CEL function naming convention seems to be camelCase (e.g. `endsWith`, `getDayOfMonth`).

We have `inIPAddrRange` and `timeSince`, but also `has_intersection` and `is_subset`.

I think we should deprecate the snake_case versions and introduce `hasIntersection` and `isSubset`.

haines:
#### Description

Fixes #669 

This PR renames the `has_intersection` and `is_subset` CEL functions to `hasIntersection` and `isSubset` for consistency with the built-in (and other custom) functions.

The old `has_intersection` and `is_subset` functions are kept for backwards compatibility, but are no longer documented.


charithe (approved):
LGTM 👍🏽

PR #673, opened 2022-02-23
chore(ci): Fix linter issues when running `make lint`, but not in CI

haines:
#### Description

I accidentally introduced a formatting error in #670, but CI didn't catch it. The golangci-lint GitHub Action does not override the `fix` setting from the config file, so when auto-fixable linter issues are found in CI they were fixed and not reported. This PR switches to using the `--fix` flag in the Makefile instead.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -117,8 +117,6 @@ linters:
 issues:
   max-same-issues: 30
 
-  fix: true
-
   exclude-rules:
     - path: _test\.go
       linters:
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ clean-tools:
 
 .PHONY: lint
 lint: $(GOLANGCI_LINT) $(BUF)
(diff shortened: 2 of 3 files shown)


charithe (commented):
`gofmt` was not enabled because `gofumpt` is enabled -- which is a stricter formatter based on gofmt. I think the issue is that the `fix` setting in the config file is ineffective. I took your Makefile modification and used it with the old golangci configuration. It fixed the formatting issue.

.golangci.yaml
```diff
@@ -58,6 +58,7 @@ linters:
     - goconst
     - gocritic
     - godot
+    - gofmt
```
charithe, line 61:
```suggestion
    
```

charithe (dismissed):

haines:
TIL about `gofumpt` - very nice! Thanks 🙂

charithe (approved):

PR #676, opened 2022-02-25
feat: Introduce `cerbosctl put`

### LINKED ISSUE #674: Add `put` subcommand to cerbosctl
Currently, there is no way to put policies or schemas with `cerbosctl`.
A subcommand `put` could solve this problem.

Example;
```
cerbosctl put policy <path>
cerbosctl put schema <path>
```

- These commands must be able to retrieve multiple paths.
- The path might point to a file or a directory.
  - If the path is a directory, the tool has to traverse all the files under the directory and submit the policies/schemas with the Admin API.
  - If the path is a file, the tool has to submit the policy/schema with the Admin API.
 - `--recursive` (`-R`) for visiting the subdirectories of the given directory.

oguzhand95:
#### Description

Fixes #674


client/admin.go
```diff
@@ -20,13 +20,17 @@ import (
...
+	addPolicyBatchSize = 10
+	addSchemaBatchSize = 10
+)
 
 type AdminClient interface {
 	AddOrUpdatePolicy(context.Context, *PolicySet) error
 	AuditLogs(ctx context.Context, opts AuditLogOptions) (<-chan *AuditLogEntry, error)
 	ListPolicies(ctx context.Context) ([]string, error)
 	GetPolicy(ctx context.Context, ids ...string) ([]*policyv1.Policy, error)
+	AddOrUpdateSchema(ctx context.Context, schemas []*schemav1.Schema) error
```
charithe, line 33: Because we have the `PolicySet` helper to make it easier to work with `AddOrUpdatePolicy`, I think we should also do something similar for the `AddOrUpdateSchema` call as well to maintain symmetry. It will also hide away the implementation detail of `schemav1.Schema` and let users just deal with files. I think something like:

```go
schemaSet.AddFile("path/to/file.json")
schemaSet.AddReader("id_of_fle", reader)
```

cmd/cerbosctl/put/policy.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	return nil
+}
+
+func (pc *PolicyCmd) Help() string {
+	return policyCmdHelp
+}
+
+func (pc *PolicyCmd) findFiles(paths []string, recursive bool) ([]*policyv1.Policy, error) {
```
charithe, line 65: Why not simply return a `PolicySet` from this instead of creating a separate array?

cmd/cerbosctl/put/policy.go
```diff
@@ -0,0 +1,134 @@
...
+
+				if d.IsDir() && recursive {
+					return nil
+				} else if d.IsDir() && !recursive && d.Name() != "." {
+					return fs.SkipDir
+				}
+
+				if d.IsDir() {
+					return nil
+				}
```
charithe, line 89:
```suggestion
				if d.IsDir() {
				    switch {
				        case recursive:
				            return nil
				        case d.Name() != ".":
				            return fs.SkipDir
				    }
				 }
					
```

cmd/cerbosctl/put/policy.go
```diff
@@ -0,0 +1,134 @@
...
+
+				p := &policyv1.Policy{}
+				if err := util.LoadFromJSONOrYAML(fsys, path, p); err != nil {
+					return nil //nolint:nilerr
+				}
+
+				if err := policy.Validate(p); err != nil {
+					return nil //nolint:nilerr
+				}
+				policies = append(policies, p)
```
charithe, line 103: If you use a `PolicySet`, you can just replace this part with `policySet.AddFile`

cmd/cerbosctl/put/policy.go
```diff
@@ -0,0 +1,134 @@
...
+			p, err := policy.ReadPolicy(f)
+			if err != nil {
+				return nil, fmt.Errorf("failed to read policy file %q: %w", path, err)
+			}
+
+			if err := policy.Validate(p); err != nil {
+				return nil, fmt.Errorf("failed to validate policy file %q: %w", path, err)
+			}
+
+			policies = append(policies, p)
```
charithe, line 129: You can replace this with `policySet.AddFile`

cmd/cerbosctl/put/schema.go
```diff
@@ -0,0 +1,142 @@
...
+					return nil
+				} else if d.IsDir() && !recursive && d.Name() != "." {
+					return fs.SkipDir
+				}
+
+				if d.IsDir() {
+					return nil
+				}
+
+				if !util.IsSupportedFileType(d.Name()) {
```
charithe, line 88: I think this would let YAML files through -- which we don't support?

client/model.go
```diff
@@ -512,6 +515,52 @@ func (ps *PolicySet) Validate() error {
...
+// AddSchemaFromReader adds a schema from the given reader to the set.
+func (ss *SchemaSet) AddSchemaFromReader(r io.Reader, id string) *SchemaSet {
+	s, err := schema.ReadSchema(r, id)
+	if err != nil {
+		ss.err = multierr.Append(ss.err, fmt.Errorf("failed to add schema from reader: %w", err))
+		return ss
+	}
+	ss.schemas = append(ss.schemas, s)
+
+	return nil
```
charithe, line 585:
```suggestion
	return ss
```

cmd/cerbosctl/put/policy.go
```diff
@@ -0,0 +1,110 @@
...
+
+				policies.AddPolicyFromFile(path)
+
+				return nil
+			})
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			if !util.IsSupportedFileType(fileInfo.Name()) {
```
charithe, line 97: If the user has provided a direct path to a file, I think we should try to read it regardless of the extension and see whether it contains a valid policy.

cmd/cerbosctl/put/policy.go
```diff
@@ -0,0 +1,110 @@
...
+					}
+				}
+
+				if !util.IsSupportedFileType(d.Name()) {
+					return nil
+				}
+
+				policies.AddPolicyFromFile(path)
+
+				return nil
```
charithe: Rather than scanning the entire directory and failing at the end, we can exit early at the first error.

```suggestion
				return policies.AddPolicyFromFile(path).Err()

```

cmd/cerbosctl/put/policy.go
```diff
@@ -0,0 +1,110 @@
...
+			})
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			if !util.IsSupportedFileType(fileInfo.Name()) {
+				return nil, fmt.Errorf("unsupported file type %q: %w", path, err)
+			}
+
+			policies.AddPolicyFromFile(path)
```
charithe, line 101:
```suggestion
			policies.AddPolicyFromFile(path)
			if err := policies.Err(); err != nil {
			    return nil, err
			}
```

cmd/cerbosctl/put/schema.go
```diff
@@ -0,0 +1,107 @@
...
+	return nil
+}
+
+func (sc *SchemaCmd) Help() string {
+	return schemaCmdHelp
+}
+
+func (sc *SchemaCmd) findFiles(paths []string, recursive bool) (*client.SchemaSet, error) {
+	schemas := client.NewSchemaSet()
+	for _, path := range paths {
```
charithe, line 62: A lot of this code is the same as the code for the put policy call. I think we can make it into a single function that invokes a callback for each file found. The callback function can take care of the special policy or schema handling code. That way we don't have to maintain two copies of the traversal code.

```go
func findFiles(paths []string, recursive bool, callback func(filePath string) error) error 
```

cmd/cerbosctl/put/policy.go
```diff
@@ -59,12 +59,21 @@ func (pc *PolicyCmd) Run(log *zap.SugaredLogger, put *Cmd, ctx *cmdclient.Contex
 		return fmt.Errorf("failed to add or update the policies: %w", err)
 	}
 
-	log.Infof("Uploaded: %d - Ignored: %d", policies.Size(), len(errs))
+	_, err = fmt.Fprintf(k.Stdout, "Uploaded: %d - Ignored: %d\n", policies.Size(), len(errs))
+	if err != nil {
+		return fmt.Errorf("failed to print: %w", err)
+	}
```
charithe, line 65: Adding newline between uploaded and ignored to make it easier to parse. We can ignore the error from `Fprint` because it would be really exceptional to fail writing to stdout and there's nothing we can do to recover from it.

```suggestion
	_, _ = fmt.Fprintf(k.Stdout, "Uploaded: %d\nIgnored: %d\n", policies.Size(), len(errs))
```

cmd/cerbosctl/put/policy.go
```diff
@@ -59,12 +59,21 @@ func (pc *PolicyCmd) Run(log *zap.SugaredLogger, put *Cmd, ctx *cmdclient.Contex
...
 	if len(errs) != 0 {
-		log.Infof("Errors for the ignored files;")
+		_, err = fmt.Fprintln(k.Stdout, "Errors for the ignored files;")
+		if err != nil {
+			return fmt.Errorf("failed to print: %w", err)
+		}
```
charithe, line 70:
```suggestion
		_, _ = fmt.Fprintln(k.Stdout, "Errors:")
```

cmd/cerbosctl/put/policy.go
```diff
@@ -59,12 +59,21 @@ func (pc *PolicyCmd) Run(log *zap.SugaredLogger, put *Cmd, ctx *cmdclient.Contex
...
 	}
 	for _, putErr := range errs {
-		log.Errorf("- %s", putErr.Error())
+		_, err = fmt.Fprintf(k.Stdout, "- %s\n", putErr.Error())
+		if err != nil {
+			return fmt.Errorf("failed to print: %w", err)
+		}
```
charithe, line 76:
```suggestion
		_, _ = fmt.Fprintf(k.Stdout, "- %s\n", putErr.Error())
```

cmd/cerbosctl/put/schema.go
```diff
@@ -59,12 +59,21 @@ func (sc *SchemaCmd) Run(log *zap.SugaredLogger, put *Cmd, ctx *cmdclient.Contex
 		return fmt.Errorf("failed to add or update the schemas: %w", err)
 	}
 
-	log.Infof("Uploaded: %d - Ignored: %d", schemas.Size(), len(errs))
+	_, err = fmt.Fprintf(k.Stdout, "Uploaded: %d - Ignored: %d\n", schemas.Size(), len(errs))
+	if err != nil {
+		return fmt.Errorf("failed to print: %w", err)
+	}
```
charithe, line 65:
```suggestion
	_, _ = fmt.Fprintf(k.Stdout, "Uploaded: %d\nIgnored: %d\n", schemas.Size(), len(errs))
```

cmd/cerbosctl/put/schema.go
```diff
@@ -59,12 +59,21 @@ func (sc *SchemaCmd) Run(log *zap.SugaredLogger, put *Cmd, ctx *cmdclient.Contex
...
 	if len(errs) != 0 {
-		log.Infof("Errors for the ignored files;")
+		_, err = fmt.Fprintln(k.Stdout, "Errors for the ignored files;")
+		if err != nil {
+			return fmt.Errorf("failed to print: %w", err)
+		}
```
charithe, line 70:
```suggestion
		_, _ = fmt.Fprintln(k.Stdout, "Errors:")
```

cmd/cerbosctl/put/schema.go
```diff
@@ -59,12 +59,21 @@ func (sc *SchemaCmd) Run(log *zap.SugaredLogger, put *Cmd, ctx *cmdclient.Contex
...
 	}
 	for _, putErr := range errs {
-		log.Errorf("- %s", putErr.Error())
+		_, err = fmt.Fprintf(k.Stdout, "- %s\n", putErr.Error())
+		if err != nil {
+			return fmt.Errorf("failed to print: %w", err)
+		}
```
charithe, line 76:
```suggestion
		_, _ = fmt.Fprintf(k.Stdout, "- %s\n", putErr.Error())
```

charithe (commented):
Just a couple of small nits to address -- otherwise LGTM 👍🏽

charithe (approved):
LGTM!

PR #678, opened 2022-02-25
docs: Fix YAML syntax in testing documentation

haines:
#### Description

This PR removes spurious trailing commas from YAML in the "Testing policies" section of the docs.


charithe (approved):
Good catch!

PR #685, opened 2022-03-01
fix: Configure GoReleaser to generate Homebrew formula in tap's `Formula` subdirectory

haines:
#### Description

Relates to https://github.com/cerbos/homebrew-tap/issues/1, see also https://github.com/cerbos/homebrew-tap/pull/2

This PR updates the GoReleaser configuration to generate the `cerbos` Homebrew formula in the `Formula` subdirectory of https://github.com/cerbos/homebrew-tap. Currently, it's being generated in the root of the tap repository, and `brew install` is picking up the older version from the `Formula` directory.


charithe (approved):
Good catch. Thanks!

PR #690, opened 2022-03-01
fix: Require at least one action in every rule

### LINKED ISSUE #687: It's possible to define a rule that applies to no actions
### Current Behavior

A rule with `actions: []` is valid.

### Expected Behavior

A validation error is produced when the `actions` are empty.

haines:
Fixes #687


charithe (approved):
LGTM

PR #692, opened 2022-03-01
feat: Add healthcheck command

### LINKED ISSUE #666: Add healthcheck subcommand
[Docker health checks](https://docs.docker.com/engine/reference/builder/#healthcheck) rely on running a command _inside_ the container and some container orchestration services (e.g. ECS) make use of this feature for their own health checks as well.
 
Shipping a third-party binary like curl or [grpc-health-probe](https://github.com/grpc-ecosystem/grpc-health-probe) with the Cerbos image is not an option because it would increase the container size needlessly and also require extra work to ensure that the external binary has no security vulnerabilities. 

Instead, we can introduce another subcommand named `healthcheck` to the Cerbos binary that can perform the [gRPC health check](https://github.com/cerbos/cerbos/blob/50a428a9dd8d6155210038c62cf80d2eaf6170f9/client/testutil/server.go#L320) by itself. It should support setting the request timeout and the Cerbos gRPC port through flags.

charithe:
Adds a `healthcheck` subcommand to the `cerbos` binary so that
in-container healthchecks (Docker HEALTHCHECK) can be performed without
resorting to an external binary.

Fixes #666

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,9 +3,11 @@ RUN apk add -U --no-cache ca-certificates && update-ca-certificates
 
 FROM scratch
 EXPOSE 3592 3593
+ENV CERBOS_CONFIG="/conf.default.yaml"
 VOLUME ["/policies"]
 ENTRYPOINT ["/cerbos"]
-CMD ["server", "--config=/conf.default.yaml"]
+CMD ["server"]
+HEALTHCHECK --interval=1m --timeout=3s CMD ["/cerbos", "healthcheck"]
 COPY --from=base /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY cerbos /cerbos
 COPY conf.default.yaml /conf.default.yaml
--- a/cmd/cerbos/healthcheck/healthcheck.go
+++ b/cmd/cerbos/healthcheck/healthcheck.go
@@ -1,0 +1,192 @@
+// Copyright 2021-2022 Zenauth Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+package healthcheck
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/alecthomas/kong"
+	svcv1 "github.com/cerbos/cerbos/api/genpb/cerbos/svc/v1"
+	"github.com/cerbos/cerbos/internal/util"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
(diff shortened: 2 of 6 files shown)


cmd/cerbos/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,192 @@
...
+`
+)
+
+type Cmd struct {
+	Kind      string        `help:"Healthcheck kind (${enum})" default:"grpc" enum:"grpc,http,https" env:"CERBOS_HC_KIND"`
+	HostPort  string        `help:"Host and port to check. Defaults to 127.0.0.1:3592 for http/https and 127.0.0.1:3593 for grpc" env:"CERBOS_HC_HOST"`
+	CACert    string        `help:"CA Certificate to use to verify the server certificate" type:"existingfile"`
+	Insecure  bool          `help:"Do not verify server certificate" default:"false"`
+	Plaintext bool          `help:"No TLS (gRPC only)" default:"false"`
+	Timeout   time.Duration `help:"Healthcheck timeout" default:"10s" env:"CERBOS_HC_TIMEOUT"`
```
haines, line 60: I personally wouldn't automatically associate "HC" with "healthcheck", is it widely adopted? My preference would be to use `CERBOS_HEALTHCHECK_` as the prefix for the environment variables.

charithe: I have seen HC being used to refer to the healthcheck in some places. But, mainly, I just want to keep the environment variable name short.

haines: I'd imagine it would typically be used in things like Kubernetes config files rather than typed at the command line, so I would favour clarity over brevity. But there's an argument either way so if you'd rather keep it as `HC`, no worries.

Dockerfile
```diff
@@ -3,9 +3,11 @@ RUN apk add -U --no-cache ca-certificates && update-ca-certificates
 
 FROM scratch
 EXPOSE 3592 3593
+ENV CERBOS_CONFIG="/conf.default.yaml"
 VOLUME ["/policies"]
 ENTRYPOINT ["/cerbos"]
-CMD ["server", "--config=/conf.default.yaml"]
+CMD ["server"]
+HEALTHCHECK --interval=1m --timeout=3s CMD ["/cerbos", "healthcheck"]
```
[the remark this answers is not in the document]
charithe: Ah! Good catch.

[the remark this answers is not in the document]
charithe: I just set the default timeout to 2s because it should never take 10s anyway. If users want to override it, they can do that in `docker run` by setting `--health-timeout=<value>` and `-e CERBOS_HC_TIMEOUT=<value>`.

haines (approved):

PR #694, opened 2022-03-02
enhancement: Allow resource policies to be empty

haines:
#### Description

Discussed in https://github.com/cerbos/cerbos/discussions/688

This PR removes the validation on resource policies' `rules` fields, to allow empty policies.

This is useful for scoped policies, where you may wish to
- have a "child" policy inherit all rules from the "parent"
- have a "parent" policy that doesn't provide any rules, leaving them up to its "children"


charithe (approved):
LGTM 👍🏽

PR #700, opened 2022-03-03
enhancement: Allow compile and schema cache sizes to be configured

charithe:
Adds configuration options to set the cache size of internal compiled
policy cache and schema cache.


haines (approved):
LGTM! It would be really nice if we could expose cache metrics to help operators tune these parameters (assuming we don't already?).

charithe:
That's a good idea. We don't have cache metrics at the moment. Created an issue to add them: #703

PR #701, opened 2022-03-04
feat!: Add matrix tests

haines:
#### Description

Currently, policy tests are table tests for a multiple principals and a single resource - you specify a resource in the input, and principals in the expectations.

This PR adds the possibility to write tests for
- a single principal and a single resource
- a single principal and multiple resources
- multiple principals and resources

It works by requiring test inputs to include either `principal` or `principals`, and either `resource` or `resources`. Expectations specify `principal` and `resource` to identify the subject under test - these can be omitted when using the singular `principal`/`resource` in the input.
An expectation for a principal-resource pairing can be omitted, in which case it's treated as implicitly expecting `EFFECT_DENY` for all actions for that pair.

There is a breaking change here - existing tests will have to be updated to include the `principals` list in `input`. This allows us to be strict about only including expectations for things that were included in the input. Attempting to keep this backwards compatible introduced too much complexity (both in the implementation and in the documentation - it is difficult to justify being allowed to expect something to be output that wasn't specified in the inputs!).


haines:
The proto check is failing - but is it necessary to maintain backwards-compatibility on those messages, since we don't expose an API for that?

charithe:
Very nice! One question: can't we represent the single resource/principal case by just having a single-item in the relevant list? Although it's nice to have explicit singular and plural forms, I wonder whether they can be gotchas for people who like to copy an example and modify it.

We can make an exception for the proto check in this case. It's mainly there to prevent inadvertent modifications to protos.

haines:
>can't we represent the single resource/principal case by just having a single entry in the relevant list?

Yep, we could do that. There is a difference in behaviour at the moment between using the singular vs plural though - the `principal` field in the expectation is not required if you used `principal` in the input. If you used `principals` in the input, you must include `principal` in the expectation (even if there is only one item in the input list).

If we remove the singular options, then we could either

1. allow the `principal` and `resource` fields to be omitted from the expectations if there is only one item in the corresponding input list. This is less verbose for the case when you have a zero- or one-dimensional table, but creates a bit of a footgun in that when you add another item to the input list the test will fail for all the existing cases until you add in the now-required field.
2. never allow the `principal` and `resource` fields to be omitted. This avoids footguns, but is arguably unnecessarily verbose for zero- or one-dimensional tables.

I would probably lean towards the "verbose but consistent" option 2, which is simpler to implement anyway. WDYT?

charithe:
IMO explicit is usually better than implicit. So, I am in favour of option 2 as well. Even if it's a little verbose, it's quite clear to anybody reading the test what it's meant to do.

haines:
Currently we have singular `auxData` in input - should that also be a list (so you can create a three-dimensional matrix over principal/resource/auxData triples)? Or is that just taking it too far 😅

charithe:
Gosh! 😬 

I'll let you be the judge. I don't have a strong opinion. I am leaning towards not doing it now and waiting to see if there's actually a need for it.

haines:
> I am leaning towards not doing it now and waiting to see if there's actually a need for it.

Agreed!

charithe (approved):
LGTM 👍🏽

charithe:
I'll do a sudo merge to override the proto check.

PR #724, opened 2022-03-08
fix: Validate that test files match schema

### LINKED ISSUE #723: Policy tests do not have their schema validated on load
We don't call `Validate` after loading test fixture files or test suite files, so the rules specified in the proto definition (e.g. a test suite requires at lease one test) aren't enforced.

haines:
#### Description

Fixes #723 

This PR calls `Validate` on the values read from test suite and fixture files, so that the rules specified in the proto message definition are enforced.


charithe (approved):
Good catch!

PR #727, opened 2022-03-10
feat: Provide JSON schemas

haines:
#### Description

This PR adds a custom `protoc` plugin, `protoc-gen-jsonschema`, which produces JSON schemas from our protocol buffer definitions, including (where possible) the validation rules implemented by `protoc-gen-validate`.

The goal of providing these schemas is to be able to use them to provide validation and autocomplete in the playground and IDEs.

#### Limitations

Currently, the schemas don't attempt to validate that only one field of a `oneof` is set; to do so efficiently whilst rejecting unknown fields requires [`unevaluatedProperties`](https://json-schema.org/understanding-json-schema/reference/object.html#unevaluated-properties) from JSON schema draft 2019-09. The generated schemas target draft 7, because 2019-09 is [not yet supported](https://github.com/microsoft/vscode/issues/98724) by Visual Studio Code's JSON language service.

#### Future enhancements

- IDE functionality like autocomplete and hover will be richer if we add a `title` (and optionally a `description`) to every field's schema; these could be generated from documentation comments in the protocol buffer definitions.
- ~We should aggregate the schemas intended to be used in playground/IDEs (policy, test) into single files so the validator doesn't have to fetch multiple small schemas.~ I changed the way it generates the schemas to avoid the need to do this - now the schema files are self-contained.


alexolivier:
Nice - is the plan to publish these to some known endpoint or just bundle in the release tar.gz?

tools/tools.mk
```diff
@@ -16,13 +16,18 @@ PROTOC_GEN_GO_GRPC := $(TOOLS_BIN_DIR)/protoc-gen-go-grpc
...
 PROTOC_GEN_OPENAPIV2 := $(TOOLS_BIN_DIR)/protoc-gen-openapiv2
 PROTOC_GEN_VALIDATE := $(TOOLS_BIN_DIR)/protoc-gen-validate
 
 GEN_DIR := api/genpb
+JSONSCHEMA_DIR := schema/jsonschema
 OPENAPI_DIR := schema/openapiv2
 MOCK_DIR := internal/test/mocks
 
+PROTOC_GEN_JSONSCHEMA_SRC_DIR := hack/tools/protoc-gen-jsonschema
+PROTOC_GEN_JSONSCHEMA_SRC_FILES := go.mod go.sum $(shell find $(PROTOC_GEN_JSONSCHEMA_SRC_DIR) -type f -name '*.go')
```
charithe, line 29: Let's create a separate, nested go module for `hack/tools/protoc-gen-jsonschema` so that it's self-contained. Then you won't need to worry about adding build tags to the files either.

charithe (approved):
Very cool! I think this would very useful to other people as well so later we should parameterise things like the schema URL and spin this off as a separate project of its own.

tools/tools.mk
haines: Good idea!

haines:
> Nice - is the plan to publish these to some known endpoint or just bundle in the release tar.gz?

@alexolivier yep it is - I was intending to make the URLs from the `$id` attributes actually valid (e.g. https://api.cerbos.dev/cerbos/policy/v1/Policy.schema.json).

Ideally I would like something similar to how the Docker images have `latest`, `dev`, and tagged releases, so the default would be to serve the version from the latest release, but it would be possible to retrieve the version corresponding to the latest commit on main and also previous releases (perhaps by providing a `?release=0.14.0` query string?).

We'd probably publish the OpenAPI definitions there too.

charithe (approved):
LGTM 👍🏽

PR #729, opened 2022-03-11
enhancement!: Add tree output functionality to the `cerbos compile` command

oguzhand95:

#### Description

- Refactor some parts of the `cerbos compile` command.
  - `!` `--format` flag is changed with `--output` to be in line with the `cerbosctl` commands.
  - `!` Adds `--color` flag to enable colored texts for the `cerbos compile` output. Default is true.
- Adds `--output=tree` as default output format and implements the tree display for the verification results (test results).


internal/verify/test_fixture.go
```diff
@@ -117,78 +121,77 @@ func loadFixtureElement(fsys fs.FS, path string, pb validatableMessage) error {
...
 		failed = true
-		sr.Tests = []TestResult{{
-			Name:    TestName{TableTestName: "Failed to load the test suite"},
-			Skipped: false,
-			Failed:  true,
-			Error:   err.Error(),
-		}}
-		return sr, failed
+
+		sn.Name = suiteNameWithErrors
```
charithe, line 140: This just prints `[ERROR]` for name. I think it should be `Unknown` or `file`

```json
{
  "suites": [
    {
      "file": "album_object_test.yaml",
      "status": "failed to load test suite: invalid TestSuite.Resources[alicia_flagged_album]: embedded message failed validation | caused by: invalid Resource.Id: value length must be at least 1 runes",
      "name": "[ERROR]",
      "failed": true
    }
  ]
}
```

cmd/cerbos/compile/internal/verification/display.go
```diff
@@ -0,0 +1,184 @@
...
+						tree = append(tree, pterm.LeveledListItem{
+							Level: 4, //nolint:gomnd
+							Text:  fmt.Sprintf("%s %s", colored.ErrorMsg("ERROR:"), action.Data.Error),
+						})
+						if verbose && action.Data.EngineTrace != "" {
+							tree = append(tree, pterm.LeveledListItem{
+								Level: 4, //nolint:gomnd
+								Text:  fmt.Sprintf("%s %s", colored.Trace("TRACE:"), action.Data.EngineTrace),
+							})
+						}
```
charithe, line 166: The tree kind of breaks down when there are traces to display. We should either print the traces at the end as a separate list OR format the traces as tree nodes so that they display nicely. I think the first option might be easier to implement and read.

```
Test results
└─┬AlbumObjectTestSuite (album_object_test.yaml)
  ├─┬alicia
  │ ├─┬alicia_private_album
  │ │ ├─┬view [FAILED]
  │ │ │ ├──ERROR:   string(
- 	"EFFECT_DENY",
+ 	"EFFECT_ALLOW",
  )

  │ │ │ └──TRACE: policy=cerbos.resource.album_object.vdefault > scope="" > derivedRole=owner > condition > expr=`request.resource.attr.owner == request.principal.id`
	result -> true

policy=cerbos.resource.album_object.vdefault > scope="" > derivedRole=owner
	activated -> true
```

internal/verify/test_fixture.go
```diff
@@ -117,78 +121,77 @@ func loadFixtureElement(fsys fs.FS, path string, pb validatableMessage) error {
...
+				testData.Failed = true
+				testData.Error = "Empty response from server"
+				failed = true
+				continue
+			}
 
-		actualResult := make(EffectsMatch, len(actual[0].Actions))
-		for key, actionEffect := range actual[0].Actions {
-			actualResult[key] = actionEffect.Effect
+			if diff := cmp.Diff(test.Expected[action].String(), actual[0].Actions[action].Effect.String()); diff != "" {
```
charithe, line 181: Some people find the diff format hard to understand so maybe we should make it simpler and have a string like `Expected: EFFECT_DENY Actual: EFFECT_ALLOW`. Also, the diff output breaks the tree display as well.

cmd/cerbos/compile/compile.go
```diff
@@ -60,26 +43,28 @@ cerbos compile --skip-tests /path/to/policy/repo
...
-	IgnoreSchemas bool   `help:"Ignore schemas during compilation"`
-	Verbose       bool   `help:"Verbose output on test failure"`
+	Dir           string               `help:"Policy directory" arg:"" required:"" type:"existingdir"`
+	Output        flagset.OutputFormat `help:"Output format (${enum})" default:"tree" enum:"tree,pretty,json" short:"o"`
+	Tests         string               `help:"Path to the directory containing tests. Defaults to policy directory." type:"existingdir"`
+	RunRegex      string               `help:"Run only tests that match this regex" name:"run"`
+	SkipTests     bool                 `help:"Skip tests"`
+	IgnoreSchemas bool                 `help:"Ignore schemas during compilation"`
+	Verbose       bool                 `help:"Verbose output on test failure"`
+	Color         bool                 `help:"Enable or disable colored output" default:"true"`
```
charithe, line 53: Since colour is on by default, I think we should name this `no-color`. It's more natural because boolean flags are "set" by  just adding them to the command (`cerbos compile --no-color` is better than `cerbos compile --color=false`)

Since there's a dedicated flag to turn colour off, maybe the default JSON output should coloured as well like we do in the `audit` output: https://github.com/cerbos/cerbos/blob/4f04ba056aa176c8daa4ce3166a2388ac2c5666c/cmd/cerbosctl/audit/audit.go#L162

cmd/cerbos/compile/compile.go
```diff
@@ -60,26 +43,28 @@ cerbos compile --skip-tests /path/to/policy/repo
...
 type Cmd struct {
-	Dir           string `help:"Policy directory" arg:"" required:"" type:"existingdir"`
-	Format        string `help:"Output format (${enum})" default:"pretty" enum:"pretty,plain,json" short:"f"`
-	Tests         string `help:"Path to the directory containing tests. Defaults to policy directory." type:"existingdir"`
-	RunRegex      string `help:"Run only tests that match this regex" name:"run"`
-	SkipTests     bool   `help:"Skip tests"`
-	IgnoreSchemas bool   `help:"Ignore schemas during compilation"`
-	Verbose       bool   `help:"Verbose output on test failure"`
+	Dir           string               `help:"Policy directory" arg:"" required:"" type:"existingdir"`
+	Output        flagset.OutputFormat `help:"Output format (${enum})" default:"tree" enum:"tree,pretty,json" short:"o"`
```
charithe, line 47: I think `pretty` output doesn't make sense anymore. It should probably be called `list`

charithe (commented):
Looks good! There are few little things to polish before merging.

cmd/cerbos/compile/internal/printer/printer.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (p *Printer) Println(args ...interface{}) {
+	fmt.Fprintln(p.stdout, args...)
+}
+
+func (p *Printer) Printf(format string, args ...interface{}) {
+	fmt.Fprintf(p.stdout, format, args...)
+}
+
+func (p *Printer) PrintJSON(val interface{}) error {
```
charithe, line 36: We should handle the `--no-color` case and output plain JSON as well

internal/verify/test_fixture.go
```diff
@@ -32,6 +31,9 @@ type validatableMessage interface {
 	Validate() error
 }
 
+// EffectsMatch is a type created to make the diff output nicer.
+type EffectsMatch map[string]effectv1.Effect
```
charithe, line 35: I think we don't need this anymore because there's no diff output

internal/verify/test_fixture.go
```diff
@@ -117,78 +119,77 @@ func loadFixtureElement(fsys fs.FS, path string, pb validatableMessage) error {
...
+			}
 
-		if diff := cmp.Diff(expectedResult, actualResult); diff != "" {
-			testResult.Failed = true
-			testResult.Error = diff
-			testResult.EngineTrace = traceBuf.String()
-			failed = true
+			if test.Expected[action].String() != actual[0].Actions[action].Effect.String() {
+				testData.Failed = true
+				testData.Error = fmt.Sprintf("Expected: %s Actual: %s", test.Expected[action].String(), actual[0].Actions[action].Effect.String())
```
charithe, line 181: This comes out as a literal string in the JSON output too (`{"error": "Expected: EFFECT_DENY Actual: EFFECT_ALLOW"}`). I think it would be better if it was a JSON object as well.

```json
{
    "error": {
         "expected": "EFFECT_DENY",  
         "actual": "EFFECT_ALLOW"
     }
}
```

internal/verify/verify.go
```diff
@@ -17,40 +17,99 @@ import (
...
+	}
+	rn.Actions = append(rn.Actions, a)
+
+	return a
 }
 
-func (r TestName) String() string {
-	return fmt.Sprintf("'%s' for resource '%s' by principal '%s'", r.TableTestName, r.ResourceKey, r.PrincipalKey)
+type ActionNode struct {
+	Data *DataNode `json:"data"`
```
charithe, line 104: Let's call this `details`

 ```suggestion
	Details *DetailsNode `json:"details"`
```

**Review by charithe — COMMENTED** [review]
Looking good! Just a few small tweaks to do. Mainly, honouring the `--no-color` flag for JSON output and renaming `DataNode` to `DetailsNode`.

File: cmd/cerbos/compile/internal/verification/internal/diff.go [hunk 10]
```diff
@@ -0,0 +1,24 @@
...
+
+package internal
+
+import (
+	"fmt"
+
+	"github.com/cerbos/cerbos/cmd/cerbos/compile/internal/colored"
+)
+
+func PrintDiff(diff map[string]string) (string, error) {
```
charithe, line 12: Wouldn't it be easier to use a struct to hold the expected and actual values?  :-)

```go
type Outcome struct {
    Expected effectv1.Effect `json:"expected"`
    Actual effectv1.Effect `json:"actual"`
}

func (o Outcome) Display() string {
    return fmt.Sprintf("Expected: %s Actual: %s", colored.FailedTest(expected), colored.FailedTest(actual))
}
```

PR #730, opened 2022-03-11
chore(test): Handle non-deterministic ordering of output directory deletion

haines:
#### Description

When generating data for the load tests, the output directories are deleted and recreated. This was flaky because the map iteration order is non-deterministic, meaning we could delete `policies` after recreating `policies/_schemas` (resulting in `no such file or directory` errors when generating schemas).


charithe (approved):
Good catch!

PR #742, opened 2022-03-15
chore(fix): Convert regular expressions to ECMAScript-compatible syntax in JSON schemas

haines:
#### Description

`protoc-gen-validate` allows Go's regexp syntax, but unfortunately ECMAScript doesn't support the same set of features (most notably, POSIX character classes like `[:word:]`, which we use extensively).

This PR parses the regexps and rewrites them with simplified character classes. Simplification of the character classes is done for us by [`syntax.Parse`](https://pkg.go.dev/regexp/syntax#Parse) - we just have to convert the resulting syntax tree back into a string without using any Perlisms (which the default `String()` implementation unfortunately does introduce).

### DIFF
--- a/schema/jsonschema/cerbos/audit/v1/DecisionLogEntry.schema.json
+++ b/schema/jsonschema/cerbos/audit/v1/DecisionLogEntry.schema.json
@@ -136,13 +136,13 @@
         },
         "policyVersion": {
           "type": "string",
-          "pattern": "^[[:word:]]*$"
+          "pattern": "^[0-9A-Z_a-z]*$"
         },
         "roles": {
           "type": "array",
           "items": {
             "type": "string",
-            "pattern": "^[[:word:]\\-\\.]+$"
+            "pattern": "^[\\--\\.0-9A-Z_a-z]+$"
           },
           "maxItems": 20,
           "minItems": 1,
@@ -150,7 +150,7 @@
         },
         "scope": {
           "type": "string",
-          "pattern": "^([[:alpha:]][[:word:]\\-]+(\\.[[:alpha:]][[:word:]\\-]*)*)*$"
+          "pattern": "^([A-Za-z][\\-0-9A-Z_a-z]+(\\.[A-Za-z][\\-0-9A-Z_a-z]*)*)*$"
         }
       }
     },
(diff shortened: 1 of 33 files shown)


schema/jsonschema/cerbos/audit/v1/DecisionLogEntry.schema.json
```diff
@@ -136,21 +136,21 @@
...
           "type": "string",
-          "pattern": "^[[:word:]]*$"
+          "pattern": "^[0-9A-Z_a-z]*$"
         },
         "roles": {
           "type": "array",
           "items": {
             "type": "string",
-            "pattern": "^[[:word:]\\-\\.]+$"
+            "pattern": "^[\\--\\.0-9A-Z_a-z]+$"
```
charithe, line 145: This seems wrong to me. It looks like a range definition from `-` to `.` unless I am mistaken?

schema/jsonschema/cerbos/audit/v1/DecisionLogEntry.schema.json
```diff
@@ -175,15 +175,15 @@
         "kind": {
           "type": "string",
           "minLength": 1,
-          "pattern": "^[[:alpha:]][[:word:]\\@\\.\\-/]*(\\:[[:alpha:]][[:word:]\\@\\.\\-/]*)*$"
+          "pattern": "^[A-Za-z][\\--9@-Z_a-z]*(:[A-Za-z][\\--9@-Z_a-z]*)*$"
```
charithe, line 178: Another odd one 🤔

charithe (commented):
LGTM!

Since this is a fix to an unreleased feature, I would use `chore(fix)` as the prefix so that it doesn't show up as a bug fix in the release notes.

charithe (approved):

schema/jsonschema/cerbos/audit/v1/DecisionLogEntry.schema.json
haines: Yeah, `syntax.Parse` simplifies character classes to the minimum set of non-overlapping ranges. No human would write that regexp (I hope 😂)

PR #743, opened 2022-03-16
enhancement: Add telemetry

charithe:
Adds ability to send anonymous usage data on server startup unless telemetry collection is disabled by config or environment variable. Only information about the Cerbos build, operating system, enabled features and aggregated statistics about policies are collected. 

Telemetry collection can be disabled by either setting `telemetry.disabled: true` in the configuration file, setting `CERBOS_NO_TELEMETRY=1`, setting `DO_NOT_TRACK=1` or using the command-line flag `--set=telemetry.disabled=true`.


internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,382 @@
...
+	return &reporter{
+		fsys:   fsys,
+		store:  store,
+		logger: logger,
+	}
+}
+
+func (r *reporter) report(ctx context.Context) bool {
+	state := r.readState()
+	if state == nil {
```
haines, line 125: Looking at `readState`, I don't think it's possible for `state` to be `nil` at this point.

haines, line 125: Looking at `readState`, I don't think it's possible for `state` to be `nil` at this point.

haines (approved):

PR #745, opened 2022-03-16
chore(ci): Upgrade to Go 1.18 on CI

charithe:
Note that this only upgrades CI workflows to use Go 1.18. The minimum Go
version in `go.mod` is still `1.17` because we don't use any new Go 1.18
features yet. Using the new Go compiler should still bring some of the
performance improvements to the builds. (20% improvement on ARM64
allegedly)

### DIFF
--- a/.github/workflows/dependabot.yaml
+++ b/.github/workflows/dependabot.yaml
@@ -15,7 +15,7 @@ jobs:
 
       - uses: actions/setup-go@v2
         with:
-          go-version: '1.17'
+          go-version: 1.18.x
 
       - name: Generate notice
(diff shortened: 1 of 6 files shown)


charithe:
Looks like the pre-built `golangci-lint` binary downloaded by the GH action can't work with Go 1.18: https://github.com/golangci/golangci-lint/pull/2438 (thanks @oguzhand95). I'll temporarily downgrade the linter Go version to 1.17 until the upstream is fixed.

oguzhand95 (approved):
LGTM. But, tests haven't been triggered.

PR #747, opened 2022-03-18
enhancement!: Make the test server fully configurable

### LINKED ISSUE #707: Add ability to use a config reader with the `testutil.StartServer` helper
In the `client` package we provide a test helper to start a Cerbos instance. It is currently limited to accepting only a few configuration parameters using the `ServerOpt` functional arguments. This artificially limits the capabilities of the test server because not all components can be configured (a good example is the database storage backends). We should allow users to provide an `io.Reader` for a Cerbos configuration and use that to initialise the test server.

charithe:
This change makes it possible to construct a test server using all the
possible Cerbos configuration options.

- Add `WithConfig` option to load configuration from a reader
- Add `WithConfigKeyValue` option to add individual config options
- Rename `WithPolicyRepositoryDatabase` to `WithPolicyRepositorySQLite3`
  because the former wasn't very useful for configuring other databases.
- Refactor config code to make it easier to test without affecting
  global state.
- Refactor some function names to maintain consistency.

Fixes #707


internal/config/config.go
```diff
@@ -56,79 +53,31 @@ func LoadMap(m map[string]interface{}) error {
...
+		return err
 	}
 
 	conf.replaceProvider(provider)
-
 	return nil
 }
 
-// LoadAndWatch automatically reloads configuration if the config file changes.
-func LoadAndWatch(ctx context.Context, confFile string, overrides map[string]interface{}) error {
```
haines, line 74: This was dead code?

haines (approved):

internal/config/config.go
charithe: Yes. It was added very early on but I never followed through with it. Now it doesn't make sense anymore because most components only read the configuration only once on startup and can't be reloaded easily.

PR #749, opened 2022-03-18
enhancement: Add protobufs for engine traces and test results

haines:
#### Description

This PR introduces protobufs for engine traces so that the JSON output has machine-readable traces, and for test results as a first step towards being able to run tests in the playground.


cmd/cerbos/compile/internal/colored/colored.go
```diff
@@ -6,15 +6,21 @@ package colored
...
+	Header                  = color.New(color.FgHiWhite, color.Bold).SprintFunc()
+	FileName                = color.New(color.FgCyan).SprintFunc()
+	ErrorMsg                = color.New(color.FgRed).SprintFunc()
+	SkippedTest             = color.New(color.FgHiWhite).SprintFunc()
+	FailedTest              = color.New(color.FgRed).SprintFunc()
+	SuccessfulTest          = color.New(color.FgGreen).SprintFunc()
+	Suite                   = color.New(color.FgBlue, color.Bold).SprintFunc()
+	Principal               = color.New(color.FgCyan).SprintFunc()
+	Resource                = color.New(color.FgBlue).SprintFunc()
+	Action                  = color.New(color.FgYellow).SprintFunc()
```
haines, line 18: I felt like the high-intensity colours were a bit overwhelming, so I dropped them down to regular intensity.

internal/verify/verify_test.go
```diff
@@ -27,40 +28,33 @@ import (
...
-		require.Len(t, result.Suites, 4)
+		require.Len(t, result.Suites, 3)
 
 		for _, sr := range result.Suites {
 			t.Run(sr.File, func(t *testing.T) {
 				is := assert.New(t)
 				switch sr.File {
-				case "empty_test.yaml":
-					is.False(sr.Skipped)
-					is.Empty(sr.Principals)
```
haines, line 46: I removed this test case because while it technically passed, `sr.Failed` was `true`; an empty test suite is not actually valid, and we have coverage for the "invalid test" case below.

internal/engine/engine_test.go
```diff
@@ -41,10 +43,8 @@ func TestCheck(t *testing.T) {
 		tcase := tcase
 		t.Run(tcase.Name, func(t *testing.T) {
 			tc := readTestCase(t, tcase.Input)
-			buf := new(bytes.Buffer)
 
-			haveOutputs, err := eng.Check(context.Background(), tc.Inputs, WithWriterTraceSink(buf))
-			t.Logf("TRACE =>\n%s", buf.String())
+			haveOutputs, err := eng.Check(context.Background(), tc.Inputs, WithZapTraceSink(zaptest.NewLogger(t)))
```
charithe, line 47: Is there any way in which we could maybe reuse the printer used by the `compile` command to output the traces here as well? The logger output is a bit difficult to parse when trying to debug something during development.

charithe (approved):
Excellent! You have made it much better 🙇🏽

internal/engine/engine_test.go
haines: Good call, done!

charithe (dismissed):
👍🏽

charithe (approved):

PR #754, opened 2022-03-18
enhancement: Add `--color` flag to `cerbos compile`

### LINKED ISSUE #752: `--output=tree` does not respect `--no-color`
`cerbos compile --output=tree --no-color` only turns off color within the tree labels, not in the tree box drawing.

### LINKED ISSUE #753: Disable color by default when stdout is not a TTY
We should toggle color as

- `--color`: force enable color
- `--no-color`: force disable color
- (neither): enable color if stdout is a TTY

haines:
#### Description

Fixes #752 (the color level is now set for `pterm` as well as the `color` library, so `--no-color` turns off all coloring in the tree output)
Fixes #753 

This PR introduces a complimentary `--color` flag in addition to the existing `--no-color`.

With the default of `--color=auto`, the output color level is detected from the environment by checking if stdout is a TTY and looking for various environment variables (see https://pkg.go.dev/github.com/jwalton/go-supportscolor for details).

Otherwise, the output color level is set to:
- none for `--no-color`, `--color=false`, or `--color=never`;
- basic for `--color`, `--color=true`, or `--color=always`;
- 256-color for `--color=256`; and
- truecolor for `--color=16m`, `--color=full`, or `--color=truecolor`

This is basically a translation of `go-supportscolor`'s ["sniff flags"](https://pkg.go.dev/github.com/jwalton/go-supportscolor#SniffFlagsOption) functionality to make it compatible with Kong (which barfs on unknown flags, so we need to explicitly define them).

### DIFF
--- a/cmd/cerbos/compile/compile.go
+++ b/cmd/cerbos/compile/compile.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/alecthomas/kong"
 	"github.com/fatih/color"
+	"github.com/pterm/pterm"
 
 	policyv1 "github.com/cerbos/cerbos/api/genpb/cerbos/policy/v1"
 	internalcompile "github.com/cerbos/cerbos/cmd/cerbos/compile/internal/compilation"
@@ -21,6 +22,7 @@ import (
 	"github.com/cerbos/cerbos/cmd/cerbos/compile/internal/verification"
 	"github.com/cerbos/cerbos/internal/compile"
 	"github.com/cerbos/cerbos/internal/engine"
+	"github.com/cerbos/cerbos/internal/outputcolor"
 	"github.com/cerbos/cerbos/internal/printer"
 	"github.com/cerbos/cerbos/internal/schema"
 	"github.com/cerbos/cerbos/internal/storage/disk"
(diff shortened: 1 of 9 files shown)


charithe (commented):
LGTM. I am wondering whether we need the `--color` flag with all these options though. Wouldn't it be simpler to just have `--no-color` and if it's not defined, try to detect the colours supported by the TTY? In my mind, getting the colours absolutely right is not necessary for this use case. Do you think it's important to let users tune it?

haines:
>Do you think it's important to let users tune it?

I think it's important to at least be able to force color to be on even if stdout is not a tty. I think the colors enhance the legibility of the output, and I regularly find myself wanting tools to support this so that I can have color turned on for CI pipelines or within Docker builds.

Is it important to be able to precisely specify whether you have basic, 256, or true color support? Perhaps less so. I initially wanted to just do

```go
Color *bool `negatable:""`
```

so that you'd have `--color` to force color, `--no-color` to force no color, and neither to detect it based on stdout being a TTY. But Kong doesn't support pointer fields yet (https://github.com/alecthomas/kong/pull/280).

Ultimately I was swayed by the fact that we already were detecting the precise color support for the colored JSON output, so it seemed reasonable to have an escape hatch for when that detection failed. This CLI flag style is widely supported in the Node ecosystem (it originally comes from [supports-color](https://www.npmjs.com/package/supports-color), which is used by [chalk](https://www.npmjs.com/package/chalk)).

charithe (approved):
LGTM

PR #767, opened 2022-03-21
chore(fix): Restrict set of JSON schema enums to match validation rule

### LINKED ISSUE #766: protoc-gen-jsonschema: Add support for `enum.in` annotation
A good example is `effect` in the policy definitions. Even though the full enum contains four values, only two are allowed to be used in the policy definitions. The rest are for internal use only. This is encoded in the validation rule as `[(validate.rules).enum = {in: [1, 2]}]`.

charithe:
In protoc-gen-jsonschema, when `enum.*` validation rules are detected, use those rules to determine the allowed set of enum values.

Also makes the base URL for schemas configurable.

Fixes #766

### DIFF
--- a/hack/tools/protoc-gen-jsonschema/module/enum.go
+++ b/hack/tools/protoc-gen-jsonschema/module/enum.go
@@ -21,9 +21,10 @@ func (m *Module) defineEnum(enum pgs.Enum) *jsonschema.StringSchema {
 
 func (m *Module) schemaForEnum(enum pgs.Enum, rules *validate.EnumRules) (jsonschema.Schema, bool) {
 	required := false
-	schemas := []jsonschema.NonTrivialSchema{m.enumRef(enum)}
 
 	if rules != nil {
+		var schemas []jsonschema.NonTrivialSchema
+
 		if rules.Const != nil {
 			schemas = append(schemas, m.schemaForEnumConst(enum, rules.GetConst()))
 			required = true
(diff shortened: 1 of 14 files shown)


hack/tools/protoc-gen-jsonschema/module/enum.go
```diff
@@ -21,9 +21,10 @@ func (m *Module) defineEnum(enum pgs.Enum) *jsonschema.StringSchema {
 
 func (m *Module) schemaForEnum(enum pgs.Enum, rules *validate.EnumRules) (jsonschema.Schema, bool) {
 	required := false
-	schemas := []jsonschema.NonTrivialSchema{m.enumRef(enum)}
```
haines, line 24: I think this needs to be here. At the moment, we have that any validation rules are additional to the base rule - which sets that the value can only be one of the known enum values.

If we remove this, it does simplify the `const` and `in` rules, but it means that the `not_in` case is no longer correct (it only validates that it's not one of the specified values, but does not validate that it is still one of the known enum values).

We could add the `enumRef` just to the `not_in` branch if we wanted to simplify the `const` and `in` cases - in which case I would say we should just `return` directly from each `if` statement to save checking `len(schemas)`, because these rules are all mutually exclusive.

hack/tools/protoc-gen-jsonschema/module/enum.go
```diff
@@ -36,10 +37,15 @@ func (m *Module) schemaForEnum(enum pgs.Enum, rules *validate.EnumRules) (jsonsc
 
 		if len(rules.NotIn) > 0 {
 			schemas = append(schemas, jsonschema.Not(m.schemaForEnumIn(enum, rules.NotIn)))
+			required = true
```
haines, line 40: I don't think `not_in` generally implies `required` - the logic I've been applying for `required` is "does the rule exclude the zero value". Arguably we could loop over the `rules.NotIn` and set `required = true` if one of the values is zero - we should probably similarly do that for the `in` case so that `required = false` if zero is included.

charithe: Good point.

haines (approved):

PR #768, opened 2022-03-22
enhancement: Reduce verbosity of test results

### LINKED ISSUE #751: Only output passing tests when `--verbose`
The test output is currently very noisy, which makes it difficult to spot failing tests. I think by default we should just display each suite with a summary of results, followed by a list of failures, like

```
purchase_order (purchase_order_test.yaml) [146 OK] [3 SKIPPED] [1 FAILED]
  regional_apac_customer
    regional_apac_thingamabobs_purchase_order
      prepareForDelivery [FAILED]
        OUTCOME: expected: EFFECT_ALLOW, actual: EFFECT_DENY
    vanilla_widgets_purchase_order
      prepareForDelivery [SKIPPED]
      sendInvoice [SKIPPED]
      view [SKIPPED]
```

With `--verbose` we can revert back to the existing format (probably still keeping the summary counts for each suite, though).

### LINKED ISSUE #770: Check backwards compatibility with the latest release rather than the latest snapshot
Currently we use [`buf breaking`](https://docs.buf.build/breaking/usage#git) to check backwards compatibility with the `main` branch. However, this means we can't make breaking changes to unreleased features.

Instead, we should compare with the latest release, which will require an extra step in the action to find the tag to compare against.

haines:
#### Description

Fixes #751 
Fixes #770 

Even for [a small demo](https://github.com/cerbos/demo-multitenant-saas/) I had 150 tests, which can mean a lot of scrolling through the `cerbos compile` output to find which one failed.

This PR omits passing tests from the `cerbos compile` output unless the `--verbose` flag is present, and instead includes a summary of each test suite and the aggregate results showing how many tests were executed and the tally of results.

I also refactored the `list` output to reuse the `tree` code, because the only difference is the box drawing of the tree structure.

I'd still like to tweak the output a bit - I really like that principals, resources, and actions are color-coded, but it feels like there's a bit too much going on at the moment. I don't love that the colors clash with the suite name/filename, or that actions are yellow (which tends to signal "warning"). Ideally we would have a slightly subtler color palette for true- and 256-color output, with a fallback for basic.

#### With `--output=tree`
Screenshot 2022-03-22 at 09 49 35

#### With `--output=list`
Screenshot 2022-03-22 at 09 49 46


haines:
I feel like the proto check should be against the previous release rather than `main` - it seems ok to break backwards-compat on unreleased features?

charithe:
> I feel like the proto check should be against the previous release rather than `main` - it seems ok to break backwards-compat on unreleased features?

I agree. Unfortunately, I don't think there's a URL that always points to the latest release that we can use. We will have to manually construct it using an additional step in the workflow. https://docs.buf.build/breaking/usage#git

haines:
>We will have to manually construct it using an additional step in the workflow.

Cool, I'll tackle that (https://github.com/cerbos/cerbos/issues/770).

charithe:
I wonder if we can use the theme colours defined by the [Chroma package](https://pkg.go.dev/github.com/alecthomas/chroma/v2@v2.0.0-alpha4/styles). Like you, I was also thinking that it would be nice to have some kind of a theme with some UX considerations built into it to handle cases like displaying errors and warnings.

charithe (approved):
LGTM 👍🏽

PR #769, opened 2022-03-22
feat: Add immediate reloading from the store with Admin API

### LINKED ISSUE #726: Add Admin API call to force an immediate reload from store
When using a polling backend (disk, git, blob) allow users to force an immediate refresh via an API call.

oguzhand95:

#### Description

Fixes #726


charithe (commented):
This is a tough one. It needs a little bit more work because unlike `blob` and `git`, it's particularly tricky to reload the `disk` store. It requires
(a) doing a full scan to rebuild the index 
(b) replacing the index on both the store and the directory watcher

I think we'll need to push down the reload call to the index to address those issues.

We also need to consider the fact that `Reload` can be expensive on large repos. So if someone accidentally (or maliciously) calls reload too many times in close succession, we should try to coalesce those calls together (something like https://pkg.go.dev/golang.org/x/sync/singleflight might work) to avoid using up all the resources on the server. 

The issue with watchers is pretty similar to the above problem too. The key point here is that we should try to prevent multiple threads from simultaneously trying to reload the same data set. All those operations should be queued up and deduped if possible.

internal/storage/disk/disk.go
```diff
@@ -103,3 +103,16 @@ func (s *Store) LoadPolicy(ctx context.Context, file ...string) ([]*policy.Wrapp
...
+
+func (s *Store) Reload(ctx context.Context) error {
+	evts, err := s.idx.Reload(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to reload the index: %w", err)
+	}
+
+	for _, evt := range evts {
+		s.NotifySubscribers(evt)
+	}
```
charithe, line 115:
```suggestion
	s.NotifySubscribers(evts...)
```

internal/storage/disk/disk_test.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil
+	}
+}
+
+func mkAddFn(t *testing.T, storeDir string) internal.MutateStoreFn {
+	t.Helper()
+
+	policiesDir := test.PathToDir(t, "store")
+	return func() error {
+		err := filepath.WalkDir(policiesDir, func(path string, d fs.DirEntry, err error) error {
```
charithe, line 70: Is it not possible to use `test.FindPolicyFiles` here?

internal/storage/disk/disk_test.go
```diff
@@ -0,0 +1,124 @@
...
+		})
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}
+}
+
+func mkDirs(t *testing.T, storeDir string) {
```
charithe, line 111: Rather than hardcoding the directory structure of the `store` dir here, I think it would be better if we just do `os.MkDirAll(filepath.Dir(file))` before copying each file

internal/storage/index/index.go
```diff
@@ -58,7 +63,9 @@ type index struct {
 	dependents   map[namer.ModuleID]map[namer.ModuleID]struct{}
 	dependencies map[namer.ModuleID]map[namer.ModuleID]struct{}
 	modIDToFile  map[namer.ModuleID]string
+	rootDir      string
```
charithe, line 66: I think it would be better to store the set of `BuildOpt` here rather than just the root dir

internal/svc/admin_svc.go
```diff
@@ -233,6 +235,35 @@ func (cas *CerbosAdminService) DeleteSchema(ctx context.Context, req *requestv1.
...
+	log := ctxzap.Extract(ctx)
+	if err := cas.checkCredentials(ctx); err != nil {
+		return nil, err
+	}
+
+	_, err, shared := sfGroup.Do("reload", func() (interface{}, error) {
+		rs, ok := cas.store.(storage.ReloadableStore)
+		if !ok {
+			return nil, status.Error(codes.Unimplemented, "Configured store is not reloadable")
+		}
```
charithe, line 248: I think we can move this check out of the singleflight group. It's a cheaper to do this check before instantiating the machinery needed for singleflight.

internal/svc/admin_svc.go
```diff
@@ -233,6 +235,35 @@ func (cas *CerbosAdminService) DeleteSchema(ctx context.Context, req *requestv1.
 	return &responsev1.DeleteSchemaResponse{}, nil
 }
 
+func (cas *CerbosAdminService) ReloadStore(ctx context.Context, _ *requestv1.ReloadStoreRequest) (*responsev1.ReloadStoreResponse, error) {
+	log := ctxzap.Extract(ctx)
+	if err := cas.checkCredentials(ctx); err != nil {
+		return nil, err
+	}
+
+	_, err, shared := sfGroup.Do("reload", func() (interface{}, error) {
```
charithe, line 244: Shouldn't this key be different from the key used by the `index.Reload` function? Something like `admin_reload` perhaps?

charithe (commented):
Looking good. Just a couple of things:
- We should log an INFO message when the reload starts and another when it finishes. Otherwise there's no indication that it happened.
- Should we make the Admin API call asynchronous by default (i.e. it starts the process and returns immediately) and only wait for the process to finish if the user sends an additional `wait` query parameter? (`/admin/store/reload?wait`)
- We need to add this API call to the admin section of the docs
- Add the `reload` command to `cerbosctl` (maybe in a different PR)

oguzhand95:
I think making the API call async by default makes sense. Because the store might consist of many policies which will increase the time the client waits. 

I have changed the method from `POST` to `GET` and added this capability by using a boolean value in the request (`reload?wait=true`).

cmd/cerbosctl/root/root.go
```diff
@@ -39,6 +40,7 @@ type Cli struct {
 	Put       put.Cmd       `cmd:"" help:"Put policies or schemas"`
 	Decisions decisions.Cmd `cmd:"" help:"Interactive decision log viewer"`
 	Audit     audit.Cmd     `cmd:"" help:"View audit logs"`
+	Store     store.Cmd     `cmd:"" help:"Modify store"`
```
charithe, line 43:
```suggestion
	Store     store.Cmd     `cmd:"" help:"Store operations"`
```

cmd/cerbosctl/store/reload.go
```diff
@@ -0,0 +1,47 @@
...
+
+	err := ctx.AdminClient.ReloadStore(context.Background(), rc.Wait)
+	if err != nil {
+		return err
+	}
+
+	if rc.Wait {
+		_, _ = fmt.Fprint(k.Stdout, "Successfully reloaded the store\n")
+	} else {
+		_, _ = fmt.Fprint(k.Stdout, "Successfully transmitted the reload store call to the Admin API\n")
```
charithe, line 39:
```suggestion
		_, _ = fmt.Fprint(k.Stdout, "Reload request submitted\n")
```

cmd/cerbosctl/store/reload.go
```diff
@@ -0,0 +1,47 @@
...
+
+type ReloadCmd struct {
+	Wait bool `help:""`
+}
+
+func (rc *ReloadCmd) Run(k *kong.Kong, ctx *cmdclient.Context) error {
+	_, _ = fmt.Fprint(k.Stdout, "Initiated a store reload\n")
+	if rc.Wait {
+		_, _ = fmt.Fprint(k.Stdout, "Waiting until the reloading process completes\n")
+	}
```
charithe, line 29: Probably don't need this

docs/modules/api/pages/admin_api.adoc
```diff
@@ -267,3 +267,36 @@ curl -k -u cerbos:cerbosAdmin -X DELETE \
...
+[#store-management]
+== Store Management
+
+=== Reload store
+
+----
+GET /admin/store/reload
+----
+
+Issue a GET request to the endpoint to reload the store.
```
charithe, line 280:
```suggestion
Issue a GET request to the endpoint to force a reload of  the store.
```

docs/modules/cli/pages/cerbosctl.adoc
```diff
@@ -43,6 +43,7 @@ Available Commands:
   get         List or view policies and schemas
   help        Help about any command
   put         Put policies or schemas
+  store       Modify store
```
charithe, line 46:
```suggestion
  store       Store operations
```

docs/modules/cli/pages/cerbosctl.adoc
```diff
@@ -265,4 +266,24 @@ cerbosctl put schema ./dir/to/schemas ./other/dir/to/schemas
 ----
 cerbosctl put schema --recursive ./dir/to/schemas
 cerbosctl put schema -R ./dir/to/schemas
+----
+
+[#store]
+== `store`
+
+The store command consists subcommands related to the stores.
```
charithe, line 274:
```suggestion
Trigger operations on the policy store of the PDP
```

docs/modules/configuration/pages/storage.adoc
```diff
@@ -32,6 +33,7 @@ storage:
 CAUTION: On some platforms the automatic change detection feature can be inefficient and resource-intensive if the watched directory contains many files or gets updated frequently.
 
 [id="blob-driver"]
```
charithe, line 35: I think `id` is the long form of the `#` directive

internal/storage/disk/disk.go
```diff
@@ -34,6 +36,7 @@ func init() {
 
 type Store struct {
 	conf *Conf
+	log  *zap.SugaredLogger
```
charithe, line 39: I don't think we need to hold a long term reference to the logger here because it's only used in the `Reload` call. We can just grab a logger instance when the function is called.

internal/storage/index/index.go
```diff
@@ -436,7 +436,7 @@ func (idx *index) RepoStats(_ context.Context) storage.RepoStats {
 }
 
 func (idx *index) Reload(ctx context.Context) ([]storage.Event, error) {
-	log := ctxzap.Extract(ctx).Sugar()
+	log := zap.L().Named("index").Sugar()
```
charithe, line 439: Why not use the logger from the context?

I think I would have passed the store logger in the context to this function and added the start/end log call around it because this is where the meat of the logic is.

internal/svc/admin_svc.go
```diff
@@ -246,19 +244,19 @@ func (cas *CerbosAdminService) ReloadStore(ctx context.Context, _ *requestv1.Rel
...
 	}
 
-	_, err, shared := sfGroup.Do("admin_reload", func() (interface{}, error) {
-		if err := rs.Reload(ctx); err != nil {
-			log.Error("Failed to reload the store", zap.Error(err))
-			return nil, status.Errorf(codes.Internal, "Failed to reload the store")
+	if req.Wait {
+		err := storage.Reload(ctx, log, rs)
+		if err != nil {
+			log.Error("failed to reload the store due to an error: %w", err)
```
charithe, line 250: I think using `%w` to output errors is not correct (I am surprised the linter didn't catch it). Anyway, when logging errors we should do `log.Error("msg", zap.Error(err))` to make the error a component of the structured logging output. Otherwise some errors won't show up properly in log aggregators.

charithe, line 250:
```go
reload := func() error {
    if err := storage.Reload(ctx, rs); err != nil {
        log.Error("failed to reload store", zap.Error(err))
        return err
    }
    return nil
}

if !wait {
    go reload()
    return &responsev1.ReloadStoreResponse{}, nil
}

if err := reload(); err != nil {
    return nil, status.Error(codes.Internal, "failed to reload store")
}

return &responsev1.ReloadStoreResponse{}, nil
```

internal/svc/admin_svc.go
```diff
@@ -246,19 +244,19 @@ func (cas *CerbosAdminService) ReloadStore(ctx context.Context, _ *requestv1.Rel
 		return nil, status.Error(codes.Unimplemented, "Configured store is not reloadable")
 	}
 
-	_, err, shared := sfGroup.Do("admin_reload", func() (interface{}, error) {
-		if err := rs.Reload(ctx); err != nil {
-			log.Error("Failed to reload the store", zap.Error(err))
-			return nil, status.Errorf(codes.Internal, "Failed to reload the store")
+	if req.Wait {
+		err := storage.Reload(ctx, log, rs)
```
charithe, line 248: log can be passed in the context

internal/storage/reload.go
```diff
@@ -0,0 +1,31 @@
...
+	"context"
+	"fmt"
+
+	"go.uber.org/zap"
+	"golang.org/x/sync/singleflight"
+)
+
+var sfGroup singleflight.Group
+
+func Reload(ctx context.Context, log *zap.SugaredLogger, rs ReloadableStore) error {
```
charithe, line 16: We should be using the logger from context because then the logging calls are request-scoped -- which is useful to identify the sequence of events that led to something. Also, use `*zap.Logger` instead of `*zap.SugaredLogger` where possible because it's more efficient.

internal/svc/admin_svc.go
```diff
@@ -244,19 +244,26 @@ func (cas *CerbosAdminService) ReloadStore(ctx context.Context, req *requestv1.R
...
+		return nil
+	}
+
+	if !req.Wait {
 		go func() {
-			err := storage.Reload(context.Background(), log, rs)
+			err := reload(ctxzap.ToContext(context.Background(), ctxzap.Extract(ctx)))
 			if err != nil {
-				log.Error("failed to reload the store due to an error: %w", err)
+				log.Error("failed to reload store", zap.Error(err))
```
charithe, line 259: The `reload` function already handles this logging so this could be simplified to just `go reload()`

internal/storage/disk/disk.go
```diff
@@ -55,7 +56,7 @@ func NewStore(ctx context.Context, conf *Conf) (*Store, error) {
 	s := &Store{
 		conf:                conf,
 		idx:                 idx,
-		log:                 zap.S().Named(DriverName),
+		log:                 zap.L().Named("disk.store").With(zap.String("dir", conf.Directory), zap.Bool("watchForChanges", conf.WatchForChanges)),
```
charithe, line 59: We don't have to keep a pointer to the log in this struct because we don't use the logger anywhere else other than the `Reload` function. 

In the `Reload` function we can grab the logger from context. So we don't need a global logger.

internal/storage/disk/disk.go
```diff
@@ -113,13 +114,11 @@ func (s *Store) RepoStats(ctx context.Context) storage.RepoStats {
 }
 
 func (s *Store) Reload(ctx context.Context) error {
-	s.log.Info("Initiated a store reload")
-	evts, err := s.idx.Reload(ctx)
+	evts, err := s.idx.Reload(logging.ToContext(ctx, s.log))
```
charithe, line 117:
```suggestion
	evts, err := s.idx.Reload(ctx)
```

internal/svc/admin_svc.go
```diff
@@ -233,6 +233,42 @@ func (cas *CerbosAdminService) DeleteSchema(ctx context.Context, req *requestv1.
...
+			if err != nil {
+				log.Error("failed to reload store", zap.Error(err))
+			}
+		}()
+		return &responsev1.ReloadStoreResponse{}, nil
+	}
+
+	if err := reload(ctx); err != nil {
+		return nil, status.Error(codes.Internal, "failed to reload store")
+	}
```
charithe, line 267:
```suggestion
	if !req.Wait {
		go reload(ctxzap.ToContext(context.Background(), ctxzap.Extract(ctx))
		return &responsev1.ReloadStoreResponse{}, nil
	}

	if err := reload(logging.ToContext(ctx, ctxzap.Extract(ctx))); err != nil {
		return nil, status.Error(codes.Internal, "failed to reload store")
	}
```

internal/storage/git/store.go
```diff
@@ -155,6 +156,27 @@ func (s *Store) RepoStats(ctx context.Context) storage.RepoStats {
...
+func (s *Store) Reload(ctx context.Context) error {
+	changes, err := s.pullAndCompare(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to pull: %w", err)
+	}
+
+	if changes == nil {
+		s.log.Info("No new commits, ignoring store reload request")
+		return nil
+	}
```
charithe, line 168:
```suggestion
	if changes == nil {
	        logging.FromContext(ctx).Info("No new commits: reload not required")
		return nil
	}
```

internal/storage/git/store.go
```diff
@@ -155,6 +156,27 @@ func (s *Store) RepoStats(ctx context.Context) storage.RepoStats {
...
+
+	evts, err := s.idx.Reload(logging.ToContext(ctx, s.log.Desugar()))
```
charithe, line 170:
```suggestion
	evts, err := s.idx.Reload(ctx)
```

internal/storage/blob/store.go
```diff
@@ -330,3 +331,23 @@ func (s *Store) LoadPolicy(ctx context.Context, file ...string) ([]*policy.Wrapp
...
+
+func (s *Store) Reload(ctx context.Context) error {
+	changes, err := s.clone(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to clone: %w", err)
+	}
+
+	if failures := changes.failures(); failures > 0 {
+		s.log.Warnf("Failed to download (%d) files", failures)
+	}
```
charithe, line 343:
```suggestion
	if failures := changes.failures(); failures > 0 {
		logging.FromContext(ctx).Warn(fmt.Sprintf("Failed to download (%d) files", failures))
	}
```

internal/storage/blob/store.go
```diff
@@ -330,3 +331,23 @@ func (s *Store) LoadPolicy(ctx context.Context, file ...string) ([]*policy.Wrapp
...
+
+	evts, err := s.idx.Reload(logging.ToContext(ctx, s.log.Desugar()))
```
charithe, line 345:
```suggestion
	evts, err := s.idx.Reload(ctx)
```

internal/storage/disk/disk.go
charithe: Do we need access to `conf`? As far as I can see, it is being used to set the logger keys here. But, since we are getting the logger from the context in `Reload`, they are irrelevant.

oguzhand95: After a private chat, turns out I misunderstood the previous review. The resolution is to use the logger coming from `ctx` since the `Reload` function of the store is directly called by a service (as opposed to other store functions working independently from API calls)

charithe (approved):
LGTM 👍🏽

PR #772, opened 2022-03-22
chore: Make playground file limit consistent

haines:
#### Description

The playground file limit is (unfortunately) specified in two places, and we missed one of them in #750.


haines:
Raised https://github.com/cerbos/cerbos/issues/774 to avoid this problem in the future by making `validate.rules` the source of truth.

charithe (approved):

PR #775, opened 2022-03-22
feat: Add an API to execute tests from the playground

haines:
#### Description

This PR adds a playground API that executes policy tests and returns the results.


api/public/cerbos/request/v1/request.proto
```diff
@@ -272,6 +272,36 @@ message PlaygroundValidateRequest {
...
+        (validate.rules).repeated = { min_items: 1, max_items: 30 },
+        (google.api.field_behavior) = REQUIRED,
+        (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+            description: "List of policy files."
+            min_items: 1
+            max_items: 30
+        }
+    ];
+
+    repeated PolicyFile test_files = 3 [
```
charithe, line 294: What's the reason for sending test files separately?

The option to run tests from a separate directory is a bit of a legacy hangover. Since we now support the Go convention of defining tests in a file with the `_test` suffix, I am thinking that  we should enforce that in the playground. WDYT?

charithe (approved):
LGTM 👍🏽

api/public/cerbos/request/v1/request.proto
haines: Yup, makes sense.

charithe (approved):
👍🏽

haines:
@alexolivier note that we're making a backwards-incompatible change to the playground API here (renaming `policyFile` field and `PolicyFile` type to `file`/`File`) - we'll need to update the client when we bump the playground to the next release version.

charithe (approved):

PR #778, opened 2022-03-23
enhancement: Publish JSON schemas for nested messages

haines:
#### Description

I'd previously assumed we only needed to generate schemas for top-level messages, but actually the `TestFixture` doesn't reference its nested messages `AuxData`, `Principals`, and `Resources`, so those schemas weren't generated.

This PR generates standalone schemas for nested messages as well, so we have cerbos/policy/v1/TestFixture/AuxData.schema.json etc.


charithe (approved):
LGTM. 
I just noticed that we generate schemas for protos in `api/private` as well. Can we exclude them?

haines:
> Can we exclude them?

Done!

charithe (approved):
👍🏽

PR #794, opened 2022-03-28
chore: Add gci linter

oguzhand95:

#### Description

#658 disabled the gci linter a while ago due to having problems with `golangci-lint - gci` and `go fmt` working against each other. To solve this, this PR adds `gci` to the tools module and makes it run with `make lint` target.


charithe:
I think we disabled `gci` because the version shipped with golangci-lint was a bit buggy. If this new version has fixed those bugs, we should wait for golangci-lint to ship with that new version. Let's not create a special configuration that lives outside of `.golangci.yaml` and requires running `make lint` because that makes things difficult for devs. (They can't simply use the built-in golangci integration in their IDEs, for example).

PR #799, opened 2022-03-29
feat: REPL for conditions

charithe:
Introduces a REPL to help experimenting with CEL conditions in an interactive environment.

It is shipped as a sub command of the Cerbos binary (`cerbos repl`) to help with rapid prototyping.

repl


cmd/cerbos/repl/repl.go
```diff
@@ -0,0 +1,102 @@
...
+	r, err := internal.NewREPL(reader, internal.NewPrinterOutput(k.Stdout, k.Stderr))
+	if err != nil {
+		return fmt.Errorf("failed to initialize the REPL: %w", err)
+	}
+
+	return r.Loop()
+}
+
+func getHistoryFile(path string) string {
+	if path == "" {
```
haines, line 48: In this case, perhaps we could put the file in `${XDG_DATA_HOME:~/.local/share}/cerbos/repl_history` to avoid polluting the user's home directory?

cmd/cerbos/repl/repl.go
```diff
@@ -0,0 +1,102 @@
...
+		if err != nil {
+			return ""
+		}
+
+		return filepath.Join(homeDir, ".cerbos_history")
+	}
+
+	finfo, err := os.Stat(path)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return ""
```
haines, line 59: Should we output a warning in this case?

cmd/cerbos/repl/internal/repl.go
```diff
@@ -0,0 +1,415 @@
...
+			r.output.PrintErr("Failed to read input", err)
+			return ""
+		}
+
+		line = strings.TrimSpace(line)
+		if line == "" {
+			return input.String()
+		}
+
+		if strings.HasSuffix(line, "\\") {
```
haines, line 124: Clearly not trivial, but it would be really nice to be able to detect unfinished expressions and avoid a hard requirement to append the trailing `\`, so that you could paste multiline JSON directly into the REPL.

charithe: You read my mind 😄 

I got bogged down in edge cases like nested strings `"foo\"bar'x'baz\"quux"` while trying to implement that. The Participle parser is not much help here and I was writing a parser by hand at that point. It was taking much longer than I anticipated so I decided to park it for a while and just get the feature out for now.

haines (approved):

PR #803, opened 2022-03-31
fix: Handle comments at the beginning of YAML files

charithe:
If the first line of the YAML file contains a comment, the
multiple-document detector incorrectly classifies the file as containing
more than one YAML document and fails. This patch fixes that issue.

### DIFF
--- a/internal/util/serdeio.go
+++ b/internal/util/serdeio.go
@@ -79,14 +79,15 @@ func newYAMLDecoder(src *bufio.Reader) decoderFunc {
 		seenContent := false
 		for s.Scan() {
 			line := s.Bytes()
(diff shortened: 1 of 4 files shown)


haines (approved):

internal/util/serdeio.go
```diff
@@ -79,6 +80,11 @@ func newYAMLDecoder(src *bufio.Reader) decoderFunc {
 		for s.Scan() {
 			line := s.Bytes()
 
+			// ignore comments
+			if bytes.HasPrefix(line, yamlComment) {
```
haines, line 84: Perhaps unlikely, but I think it's valid for the # to be preceded by whitespace.

haines (approved):

PR #819, opened 2022-04-05
feat: Unified check API

### LINKED ISSUE #697: Unified check API
At the moment we have two check APIs - `check`, and `check_resource_batch`.

For the most part, `check` (which takes a homogeneous list of resources)  is just a special case of `check_resource_batch` (which takes a heterogeneous list of resources).

However, `check` accepts `includeMeta` but `check_resource_batch` does not. Also, the response formats are quite different, making it nontrivial to replace one call with the other.

There is a cognitive load for anyone consuming the API to decide which call they need. A single unified API would be simpler to understand.

We should introduce a new unified API based on `check_resource_batch`, which should also
- support `includeMeta`
- return `resource: { kind, id }` rather than `resourceId` (to account for cases where different kinds of resource have the same id, originally proposed in https://github.com/cerbos/cerbos/issues/695)

The existing `check` and `check_resource_batch` APIs should be gradually deprecated

- reduce their prominence in the documentation (a separate "deprecated" section/page?) and add a notice to point users to the new API
- add a deprecation warning to the PDP logs

charithe:
- Consolidate the various check APIs into a single one so that there's less confusion about which one to use. 
- Add deprecation notices to the old API endpoints
- Update documentation to use the new API
- Update the SDK and add a convenience helper to reuse a principal

Fixes #697


docs/modules/api/pages/index.adoc
```diff
@@ -40,10 +40,148 @@ Other languages coming soon
...
+[source,json,linenums]
+----
+{
+  "requestId": "test", <1>
+  "results": [ <2>
+    {
+      "resource": { <3>
+        "id": "XX125",
+        "kind": "leave_request",
+        "policyVersion": "20210210"
```
haines, line 126: Should this have `"scope": "acme.corp"` since that was in the request?

charithe: Good catch!

haines (approved):

PR #820, opened 2022-04-05
enhancement: Execute policy conditions in the CEL REPL

oguzhand95:


cmd/cerbos/repl/internal/repl.go
```diff
@@ -318,6 +355,112 @@ func (r *REPL) evalExpr(expr string) (ref.Val, *exprpb.Type, error) {
...
+				_, ok := rule.Condition.Condition.(*policyv1.Condition_Match)
+				if !ok {
+					continue
+				}
+				r.rules = append(r.rules, rule)
+			}
+		}
+
+		res := pt.ResourcePolicy.Resource
+		if pt.ResourcePolicy.Scope != "" {
```
charithe, line 383: Use the `namer` package to keep the policy names consistent everywhere.

cmd/cerbos/repl/internal/repl.go
```diff
@@ -318,6 +355,112 @@ func (r *REPL) evalExpr(expr string) (ref.Val, *exprpb.Type, error) {
...
+
+	condition, err := compile.Condition(cond)
+	if err != nil {
+		return fmt.Errorf("failed to compile condition: %w", err)
+	}
+
+	return r.evalCondition(condition)
+}
+
+func (r *REPL) evalCondition(condition *runtimev1.Condition) error {
```
charithe, line 433: Could we display this as a tree?

```
- all [false]
    |- expr: '"mark"+"eting" == "marketing"'  [true]
    |- exp: 1 == 2 [false]  
```

cmd/cerbos/repl/internal/repl.go
```diff
@@ -50,15 +56,18 @@ const (
 	directivePrefix = ':'
 	prompt          = "-> "
 	secondaryPrompt = "> "
+	rulePrefix      = "rule_"
```
charithe, line 59: Something like `#1` might be easier to type than `rule_1`

cmd/cerbos/repl/internal/repl.go
```diff
@@ -318,6 +355,112 @@ func (r *REPL) evalExpr(expr string) (ref.Val, *exprpb.Type, error) {
...
+	case *policyv1.Policy_DerivedRoles:
+		for _, def := range pt.DerivedRoles.Definitions {
+			if def.Condition != nil {
+				r.rules = append(r.rules, def)
+			}
+		}
+
+		r.output.Println(fmt.Sprintf("Derived roles '%s' loaded", pt.DerivedRoles.Name))
+	}
+
```
charithe, line 397: It would be good to print the formatted policy after loading it

charithe (commented):
Let's try to keep a single blank line between the output and the next prompt at all times.

```
blah blah

->
```

cmd/cerbos/repl/internal/repl.go
```diff
@@ -318,6 +350,137 @@ func (r *REPL) evalExpr(expr string) (ref.Val, *exprpb.Type, error) {
...
+	//nolint:gomnd
+	id, err := strconv.ParseInt(after, 10, 64)
+	if err != nil {
+		return fmt.Errorf("invalid id format")
+	}
+
+	if int(id) >= len(r.rules) {
+		return fmt.Errorf("failed to find the rule %s", name)
```
charithe, line 419:
```suggestion
		return fmt.Errorf("failed to find rule %s", name)
```

cmd/cerbos/repl/internal/directive.go
```diff
@@ -38,14 +40,25 @@ func NewParser() (*participle.Parser, error) {
...
 	Expr string `parser:"'=' @Any"`
 }
+
+type LoadDirective struct {
+	Path string `parser:"'load' @Path"`
+}
+
+type ExecDirective struct {
+	RuleName string `parser:"'exec' @Id"`
+}
```
charithe, line 64:
```suggestion
type ExecDirective struct {
	RuleID int `parser:"'exec' '#'@Int"`
}
```

cmd/cerbos/repl/internal/directive.go
```diff
@@ -6,14 +6,16 @@ package internal
...
 
-	participle "github.com/alecthomas/participle/v2"
+	"github.com/alecthomas/participle/v2"
 	"github.com/alecthomas/participle/v2/lexer"
 )
 
 func NewParser() (*participle.Parser, error) {
 	lex, err := lexer.New(lexer.Rules{
 		"Root": {
+			{Name: "Id", Pattern: `#[0-9]+`},
```
charithe, line 16:
```suggestion
			{Name: "Int", Pattern: `[0-9]+`},
```

charithe (commented):
It's coming along very nicely. There are a few things to address before we can merge it though. 

-  Variables are not automatically populated from the `variables` in the policy. For an example with variables, see `internal/test/testdata/store/resource_policies/policy_01.yaml`
- Condition evaluation doesn't seem to handle short-circuiting correctly. Some operators like `all` and `none` short-circuit when they encounter the first condition that is not false/true.
- There seems to be a bug in how the conditions are evaluated. Some of the expressions return conflicting results based on whether they are executed as part of rule or individually.
    ```
    -> :exec #2
    └─┬all [true]
      ├──"cerbos-jwt-tests" in request.aux_data.jwt.aud [true]
      └──"A" in request.aux_data.jwt.customArray [true]
    
    -> "A" in request.aux_data.jwt.customArray
    _ = false
    ```
- When printing rules, we should print them as YAML because that's how users usually work with them. Also they should be in colour to be consistent with other outputs from REPL directives.

cmd/cerbos/repl/internal/repl.go
```diff
@@ -68,10 +81,12 @@ func NewREPL(reader *liner.State, output Output) (*REPL, error) {
 	}
 
 	repl := &REPL{
-		reader:   reader,
-		parser:   parser,
-		output:   output,
-		toRefVal: conditions.StdEnv.TypeAdapter().NativeToValue,
+		reader:     reader,
+		parser:     parser,
+		mkRuleID:   newCounter(),
```
charithe, line 86: Do we need this now?

cmd/cerbos/repl/internal/repl.go
```diff
@@ -46,19 +53,25 @@ var (
...
 type REPL struct {
-	vars     variables
-	decls    map[string]*exprpb.Decl
-	reader   *liner.State
-	parser   *participle.Parser
-	output   Output
-	toRefVal func(any) ref.Val
+	output     Output
+	vars       variables
+	policyVars map[string]string
```
charithe, line 68: There's a disconnect between the "REPL vars" and the vars from the policy here. You can see this when you type `:vars`:

```
variables = 0

variables.pending_approval = "PENDING_APPROVAL"

variables.principal_location = ""
```

We should add the policy variables to REPL variables when a policy is loaded and remove them when another policy is loaded.

charithe (commented):
Nearly there! I mostly have nits about the formatting of things. However, the bug with the complex conditions still exists:

```
-> :let request = {"principal": {"id": "alice", "attr": {"ip_address": "10.1.1.1"}}, "auxData": {"jwt": {"aud": ["cerbos-jwt-tests"], "customArray": ["A", "B"]}}}

-> :exec #2
└─┬all [true]
  ├──"cerbos-jwt-tests" in request.aux_data.jwt.aud [true]
  └──"A" in request.aux_data.jwt.customArray [true]

-> :let request = {"principal": {"id": "alice", "attr": {"ip_address": "10.1.1.1"}}, "auxData": {"jwt": {"aud": ["cerbos-jwt-tests"], "customArray": ["C", "B"]}}}

-> :exec #2
└─┬all [true]
  ├──"cerbos-jwt-tests" in request.aux_data.jwt.aud [true]
  └──"A" in request.aux_data.jwt.customArray [true]

-> "A" in request.aux_data.jwt.customArray
_ = false
```


Nits about the output:

- It's hard to visually distinguish the rule ID and the definition when `:load` or `:rules` is executed. I think the YAML for the rule definition should be printed with an indent (or using the tree output) as follows:
    ```yaml
     [#0] # in bold
          actions:
          - approve
          condition:
            match:
              expr: request.resource.attr.status == V.pending_approval
          derivedRoles:
          - direct_manager
          effect: EFFECT_ALLOW
    ```
- I don't think we need to print `(resource_policies)` after the rule ID. When the user types `:rules` we should print the name of the policy first and then print out the rules (just like `:load` does) and that is enough context.

PR #823, opened 2022-04-06
fix: Use the branch name in git pull

charithe:
In some instances, pull operations fail with the error
`non-fast-forward update`. When the branch name is explicitly set in the
pull options, that error seems to go away.


charithe:
I am merging this without review to give the snapshot to Jan for testing.

PR #857, opened 2022-04-21
chore: Make playground test API response match validate and evaluate

haines:
Implementing the playground client for tests, I realised I goofed a bit when defining the API because the response doesn't have the same shape as the other playground APIs, which means the client can't handle the responses in a consistent fashion.

This PR moves the test API's `results` inside a `success` field to match the evaluate API's response. It's backwards-incompatible but nothing is consuming that API yet, so it's not a problem to make a breaking change.


charithe (approved):
LGTM

PR #858, opened 2022-04-22
enhancement: Add trace for denial due to no matching policies

haines:
Currently, when no policy matches, no engine trace is generated. This is problematic when a test fails because `--verbose` gives no additional information.

This PR adds a trace event when the default effect is applied due to no matching policies. I also removed the short-circuit for when there are no checks present in the evaluation context, because this condition is also checked within `(*evaluationCtx).evaluate`.

Now, the verbose output is more helpful for diagnosing the test failure:

```
purchase_order - regional_operations.regional_emea_widgets_purchase_order.sendInvoice
  action=sendInvoice
    activated
    effect → deny
    No matching policies
```

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -375,21 +375,24 @@ func (engine *Engine) evaluate(ctx context.Context, input *enginev1.CheckInput, 
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


charithe (approved):
LGTM

PR #859, opened 2022-04-22
enhancement: Auto-detect end of input in REPL

oguzhand95:

#### Description
This PR changes the behavior of the REPL in a way that instead of determining the end of the input only with the characters `\\`, it also uses the special characters (`{`, `}`, `[`, `]`, `(`, `)`, `"`, `"`).


charithe (commented):
Good work. From a quick glance, the algorithm seems OK to me. 
- Let's make it into a single function called `isUnterminated` or something similar. That makes the prompt handling code much simpler and also makes it easy to write tests for the logic.
- We need to handle single quotes (`'`) as well because they are valid string delimiters in JSON. That means we have to also think about nested strings (e.g. `"foo_'_bar_\"_baz"`)
- We should also handle the case where the newline is escaped (`\` at the very end of the line). This is a special case that we can use to short-circuit the logic of checking each rune one-by-one. At the start of the function, check whether the input ends in `\` and return `true` (meaning the string is unterminated) 
- This definitely needs unit tests

cmd/cerbos/repl/internal/repl.go
```diff
@@ -534,6 +541,48 @@ func (r *REPL) mkEnv() (*cel.Env, error) {
 	return conditions.StdEnv.Extend(cel.Declarations(decls...))
 }
 
+func isTerminated(line string, stack *runeStack) bool {
+	if line[len(line)-1] == '\\' {
```
charithe, line 545: We need to remove the `\` from the end of the string as well. Otherwise, the input is invalid.

```
-> :let request = {\
>   "principal":{\
>     "id":"john",\
>     "roles":["employee"],\
>     "attr":{"scope":"foo.bar.baz.qux"}\
>   },\
>   "resource":{\
>     "id":"x1",\
>     "kind":"leave_request",\
>     "attr":{"scope":"foo.bar"}\
>   }\
> }
Error: Failed to process input
 failed to unmarhsal JSON as "request": proto: syntax error (line 1:3): invalid value \
```

charithe (commented):
Looks good. I can paste long snippets now without escaping them 👍🏽 

There's a small bug where escaped newlines are not removed from the input. Otherwise it's good to merge.

charithe (approved):
LGTM

PR #920, opened 2022-05-13
enhancement: Allow rule match on any role

oguzhand95:

#### Description
This PR makes it possible to provide wildcards for the `role` field in the resource policy rules and the `parentRoles` field in the derived roles.


internal/engine/evaluator.go
```diff
@@ -391,9 +391,11 @@ func toSet(values []string) stringSet {
 }
 
 func setIntersects(s1 protoSet, s2 stringSet) bool {
```
charithe, line 393: Let's rename and move this function to the file with the glob functions in the `util` package so  that it is obvious that it uses glob matching.

charithe:
Implementation looks OK to me but the benchmarks show a non-trivial overhead. Let me think about this a little bit more to see if there are ways to reduce that.

haines (dismissed):

haines:
We probably need to add something to the derived role and resource policy docs to cover this feature.

haines (approved):

PR #924, opened 2022-05-16
chore: Add API usage stats to telemetry

charithe:
Add a telemetry event to periodically report how many requests were made to each RPC. Also reports the count of gRPC user agents to get an idea about SDK usage.


internal/telemetry/middleware.go
```diff
@@ -0,0 +1,138 @@
...
+
+	telemetryv1 "github.com/cerbos/cerbos/api/genpb/cerbos/telemetry/v1"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/protobuf/types/known/durationpb"
+)
+
+const collectorBufferSize = 64
+
+var totalReqCount uint64
```
haines, line 20: Would it make sense for this to be a field on the `statsInterceptors` struct similar to the tallies?

internal/telemetry/telemetry.go
```diff
@@ -169,42 +230,80 @@ func newState() *statev1.TelemetryState {
...
 }
 
-func (r *reporter) writeState(state *statev1.TelemetryState) error {
-	stateBytes, err := protojson.Marshal(state)
+func (r *segmentReporter) Stop() error {
+	var err error
+	r.closeOnce.Do(func() {
+		close(r.shutdownChan)
+		close(r.eventChan)
+		r.reportServerStop()
```
haines, line 284: Perhaps we could report the API activity here too? Otherwise if we stop the server after 59 minutes we lose a bunch of stats.

We might be able to move the request count onto the API activity event in that case.

internal/telemetry/middleware.go
```diff
@@ -0,0 +1,138 @@
...
+func (i *statsInterceptors) collectStats(ctx context.Context, method string) {
+	if strings.HasPrefix(method, "/grpc.") {
+		return
+	}
+
+	mInfo := methodInfo{name: method, userAgent: "unknown"}
+
+	md, ok := metadata.FromIncomingContext(ctx)
+	if ok {
+		if v := md.Get("user-agent"); len(v) > 0 {
```
haines, line 89: Does the HTTP gateway populate this metadata from the `User-Agent` header?

internal/telemetry/telemetry.go
charithe: The tricky part here is that interceptor runs in a different goroutine. When the system is shutting down, it's difficult to reliably  synchronize the reporter and the interceptor without resorting to some convoluted blocking code (which could hold up the process from shutting down properly). Because the API and UA stats are complex, they can't be shared between goroutines without locking -- which is not ideal for performance. That is why, as a compromise, the total request count is declared as a global variable. That way if the PDP is short-lived, we still get some idea of usage.

haines (approved):

PR #945, opened 2022-05-25
enhancement: Configurable request limits

charithe:
Allow the max batch size and max actions limit to be configured


api/public/cerbos/request/v1/request.proto
```diff
@@ -190,7 +185,7 @@ message CheckResourceBatchRequest {
       (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
         description: "List of actions being performed on the resource."
         min_items: 1
-        max_items: 10
+        max_items: 50
```
haines, line 188:
```suggestion
```

internal/server/conf.go
```diff
@@ -16,11 +16,14 @@ import (
...
-	defaultAdminUsername        = "cerbos"
-	defaultRawAdminPasswordHash = "$2y$10$VlPwcwpgcGZ5KjTaN1Pzk.vpFiQVG6F2cSWzQa9RtrNo3IacbzsEi" //nolint:gosec
+	confKey                       = "server"
+	defaultHTTPListenAddr         = ":3592"
+	defaultGRPCListenAddr         = ":3593"
+	defaultAdminUsername          = "cerbos"
+	defaultRawAdminPasswordHash   = "$2y$10$VlPwcwpgcGZ5KjTaN1Pzk.vpFiQVG6F2cSWzQa9RtrNo3IacbzsEi" //nolint:gosec
+	defaultMaxActionsPerResource  = 50
+	defaultMaxResourcesPerRequest = 50
+	requestItemsMax               = 500
```
haines, line 26: I'm not sure we really need to enforce this limit - e.g. if we'd allow 500×500, why wouldn't we allow 1000×10? Once you get into tuning these away from the defaults I think you're in the realm of "at your own risk", and it's clearly documented that it can affect performance.

charithe: Well, it's at the user's risk but they could have made a typo or made the change without really understanding the implications ("just stick a large number in" mentality) that could come back to bite them later. I feel that as we are more intimately familiar with the code, we should provide some guard rails here.

500x500 is different from 250x1000 or 10000x25 because the number of actions determine how many times the rules in a policy get evaluated and the number of resources determine how many policies need to be loaded from storage. So they all have different performance characteristics and that's why I chose to expose these two numbers rather than just their multiplication.

haines (approved):

PR #950, opened 2022-05-27
docs: Update audit examples

charithe:
Due to a change in how config is unmarshaled, the defaults are not
correctly being set. Until that issue is fixed, we need to update the
examples to instruct the users that they should explicitly set the
`accessLogsEnabled` and `decisionLogsEnabled` settings.


alexolivier:
👍

PR #963, opened 2022-06-01
enhancement: Expand aliases in PlanResources output

### LINKED ISSUE #874: The query planner may produce a suboptimal plan

### Current Behavior

The query planner might produce a plan containing precalculated subexpressions that can be eliminated:
```json
{
  "filter": {
    "kind": "KIND_CONDITIONAL",
    "condition": {
      "expression": {
        "operator": "and",
        "operands": [
          {
            "expression": {
              "operator": "eq",
              "operands": [
                {
                  "variable": "R.attr.owner"
                },
                {
                  "value": "john"
                }
              ]
            }
          },
          {
            "value": true
          }
        ]
      }
    }
  },
  "meta": {
    "filterDebug": "((R.attr.owner == \"john\") AND (true))",
    "matchedScope": "acme.hr.uk"
  }
}
```

Here the second subexpression can be removed.

### Expected Behavior

The query plan produces the minimal logical plan. In this case, the plan has to be reduced to the first subexpression: `R.attr.owner == "john"`.

### Steps To Reproduce

1. Run `make dev-server-insecure` in the repository to start the development server.
2. Run `make check-grpc-insecure`.

### Environment

_No response_

### Anything else?

_No response_

### LINKED ISSUE #962: Expand variable aliases in PlanResources AST
Currently the AST can contain either the full variable (e.g. `request.resource`) or the short alias (e.g. `R`), depending on what was used in the policy. It would be better to always return the full variable name so that you don't have to consider the alias as well when consuming the AST.

charithe:
Expands aliases like `R` into their full form (`request.resource`) so
that regardless of the expression used in the policy, the PlanResources
output is always predictable.

Also implements some simple optimisations like converting `X && true`
into `X` or `X || false` into `false`.

Fixes #962
Fixes #874

### DIFF
--- a/internal/conditions/cel.go
+++ b/internal/conditions/cel.go
@@ -5,6 +5,7 @@ package conditions
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/google/cel-go/cel"
 	"github.com/google/cel-go/checker/decls"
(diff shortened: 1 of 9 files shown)


dbuduev (approved):
LGTM 👍🏻 

Suggestion: cover `NOT true/false` cases.

dbuduev:
I've created a separate issue for `FilterDebug`. https://github.com/cerbos/cerbos/issues/964

charithe:
Let's cover `NOT true/false` case while fixing #964

PR #966, opened 2022-06-02
enhancement: Allow setting file mode of UDS

charithe:
Allow the file mode of the Unix domain socket to be set by the user. By
default Go creates the socket with 0o766 which makes it unusable if the
client is running as a different user.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -581,7 +581,7 @@ func toUDSFileMode(modeStr string) os.FileMode {
 	}
 
 	// Ignore everything but the last 9 bits which hold the user, group and world perms.
-	return os.FileMode(m & 0x1FF)
+	return os.FileMode(m & 0o777)
 }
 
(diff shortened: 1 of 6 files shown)


internal/server/server.go
```diff
@@ -564,6 +573,17 @@ func parseAndOpen(listenAddr string) (net.Listener, error) {
...
 
+//nolint:gomnd
+func toUDSFileMode(modeStr string) os.FileMode {
+	m, err := strconv.ParseInt(modeStr, 0, 32)
+	if err != nil || m <= 0 {
+		return 0o766
+	}
+
+	// Ignore everything but the last 9 bits which hold the user, group and world perms.
+	return os.FileMode(m & 0x1FF)
```
dbuduev, line 584: nit: I think `os.FileMode(m & 0o777)` is a bit easier to read.

dbuduev (dismissed):
LGTM

PR #968, opened 2022-06-03
fix: Handle `google.protobuf.Value` correctly in OpenAPI schemas

haines:
This PR updates `protoc-gen-openapiv2` to pull in https://github.com/grpc-ecosystem/grpc-gateway/pull/2719 which fixes `google.protobuf.Value` having a `{ "type": "object" }` schema, when in fact it can be any JSON-serializable value and should have an empty schema `{}`.


charithe (approved):
That was quick! 👍🏽

charithe:
Merging this instead of #979

PR #973, opened 2022-06-06
fix: Cerbosctl version should fail gracefully

### LINKED ISSUE #971: Cerbosctl version should fail gracefully
Running `cerbosctl version` fails when server credentials are not provided. This is because it tries to connect to Cerbos to get the remote version. When credentials are not provided or when the remote service is not accessible, it should still print the local version of `cerbosctl` instead of failing completely.

oguzhand95:

#### Description

Fixes #971

### DIFF
--- a/cmd/cerbosctl/version/version.go
+++ b/cmd/cerbosctl/version/version.go
@@ -16,14 +16,19 @@ import (
 type Cmd struct{}
 
 func (c *Cmd) Run(k *kong.Kong, ctx *cmdclient.Context) error {
-	r, err := ctx.Client.ServerInfo(context.Background())
+	_, err := fmt.Fprintf(k.Stdout, "Client version %s; commit sha: %s, build date: %s\n", util.Version, util.Commit, util.BuildDate)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


cmd/cerbosctl/version/version.go
```diff
@@ -16,14 +16,19 @@ import (
...
-	_, err = fmt.Fprintf(k.Stdout, "Client version %s; commit sha: %s, build date: %s\n", util.Version, util.Commit, util.BuildDate)
+	r, err := ctx.Client.ServerInfo(context.Background())
 	if err != nil {
-		return err
+		_, errPrint := fmt.Fprintf(k.Stdout, "Server version unknown; commit sha: unknown, build date: unknown\n")
+		if errPrint != nil {
+			return errPrint
+		}
+
+		return fmt.Errorf("failed to retrieve version, commit sha and build date from server: %w", err)
```
charithe, line 31:
```suggestion
		return fmt.Errorf("failed to retrieve version information from the server: %w", err)
```

charithe (approved):
👍🏽

PR #985, opened 2022-06-07
fix: Ignore test files when updating index from git

### LINKED ISSUE #984: git store does not ignore test files in policy directories
### Current Behavior

When the git store [applies an index update](https://github.com/cerbos/cerbos/blob/8ad06a6edc7dd99c80d4811ec3013c407e90cae5/internal/storage/git/store.go#L407), it only checks if the changed file is a supported type - not whether it is a test file.

This means that we attempt to read test files as policies and fail to unmarshal.

### Expected Behavior

Index updates ignore test files

haines:
Fixes #984 

The git store currently only ignores test files on the initial load - index updates attempt to read tests as policies, causing an unmarshalling error.

This PR ignores test files in index updates as well.

### DIFF
--- a/internal/storage/git/store.go
+++ b/internal/storage/git/store.go
@@ -438,6 +438,11 @@ func (s *Store) applyIndexUpdate(ce object.ChangeEntry, eventKind storage.EventK
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


charithe (approved):
LGTM

PR #989, opened 2022-06-08
enhancement: Add store polling metrics

### LINKED ISSUE #987: Metrics for storage engine activity

### Feature description

When using git (or other storage engines) having metrics for successful and unsuccessful syncs would be useful for creating alerting around if PDP instances have got stuck pulling the latest policies and thus on an older version.

Produce a set of metrics for the success and failure counts of pulling policies from storage. Possibly expose the last-sync time in some manner.

### Anything else?

_No response_

charithe:
Increments a counter each time a store is polled for updates and records
the success or failure status.

Also adds counters for CRUD operations.

Fixes #987

### DIFF
--- a/internal/storage/disk/dirwatch.go
+++ b/internal/storage/disk/dirwatch.go
@@ -91,7 +91,6 @@ func (dw *dirWatch) handleEvents(ctx context.Context) {
 			dw.triggerUpdate()
 			_ = stats.RecordWithTags(context.Background(), []tag.Mutator{
 				tag.Upsert(metrics.KeyStoreDriver, DriverName),
-				tag.Upsert(metrics.KeyStorePollStatus, "success"),
 			}, metrics.StorePollCount.M(1))
 		}
 	}
@@ -130,6 +129,7 @@ func (dw *dirWatch) triggerUpdate() {
 		dw.eventBatch = make(map[string]struct{})
 		dw.mu.Unlock()
 
+		errCount := 0
 		for f := range batch {
 			fullPath := filepath.Join(dw.dir, f)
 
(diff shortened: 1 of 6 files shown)


internal/storage/disk/dirwatch.go
```diff
@@ -86,6 +89,10 @@ func (dw *dirWatch) handleEvents(ctx context.Context) {
 			dw.processEvent(evtInfo)
 		case <-ticker.C:
 			dw.triggerUpdate()
+			_ = stats.RecordWithTags(context.Background(), []tag.Mutator{
+				tag.Upsert(metrics.KeyStoreDriver, DriverName),
+				tag.Upsert(metrics.KeyStorePollStatus, "success"),
```
haines, line 94: It seems a bit strange that this is unconditionally successful? I think the error handling is different between disk and git - disk just logs warnings and continues to apply other changes, whereas git stops on the first error?

haines (approved):

internal/storage/disk/dirwatch.go
charithe: Yes, I noticed that myself whilst working on this PR. I think the reason they are different is that with the disk store, someone could potentially be in the middle of an operation (like copying an existing file and modifying it) and it would be better to just ignore that and continue rather than throwing an error and stopping. Whereas, with git, we only look at committed changes. If something is wrong, then it will always be wrong until there's a new commit.

I do agree that it's not great to have different behaviours but I didn't want to change it for this PR without fully understanding the implications.

haines: I wonder if a more useful metric would be the number of errors encountered within `triggerUpdate`?

haines (approved):

PR #992, opened 2022-06-10
fix: Respect `WithRootDir` option in `(Index).ListSchemaIDs`

haines:
The `ListSchemaIDs` method currently assumes that schemas are located in `./_schemas`, but the index's filesystem is not rooted at the policies directory.

This PR refactors the method to delegate to the `SchemaLoader`, which is aware of the policies directory set by `WithRootDir`.


charithe (approved):
LGTM. Nice catch!

PR #993, opened 2022-06-10
enhancement!: Ignore hidden directories and files

### LINKED ISSUE #969: Don't traverse into hidden directories when scanning for policies
https://cerboscommunity.slack.com/archives/C02A364JYMQ/p1653990422896079

When mounting a configmap as a volume, Kubernetes creates a folder structure like

```
..2022_06_03_16_20_35.2211801369
..data -> ..2022_06_03_16_20_35.2211801369
foo.yaml -> ..data/foo.yaml
```

(the rationale being to allow atomic updates of the configmap by switching the `..data` symlink to a new target)

This causes an issue when mounting Cerbos policies from a configmap, because the disk driver will find both the symlink and the target of the symlink, treating them as separate files, and erroring due to duplicate policies.

As a workaround Cerbos can be pointed at `..data` rather than the mounted directory.

~Ideally, if Cerbos finds a symlink and its target, it would only load the file once.~ - this is too complex

Cerbos shouldn't enter hidden directories (name starting with `.`) when walking the filesystem.

haines:
Fixes #969 

This PR prevents Cerbos from looking in hidden directories or at hidden files when searching for policies and schemas.

I've refactored the handling of the policies directory in all the relevant storage drivers so the behaviour should be more consistent across drivers. The most notable issues I uncovered were:

- blob and disk: index updates attempt to load test files and files in subdirectories of `_schemas` as policies
- git: index updates don't delete the old file on rename (so renaming policies causes duplicate policy error, and renaming schemas leaves the old schema in the index)

The small breakages:
- users with hidden directories/files will need to restructure their policies directory (seems unlikely);
- schemas will no longer be loaded unless their file extension is `.json`; and
- `cerbosctl put schema --recursive` will now preserve relative paths in the schema IDs, so that `cerbosctl put schema --recursive path/to/policies/_schemas` will result in the same IDs as if the schemas were loaded by a storage driver pointing at `path/to/policies` (currently, the IDs are just set to the basename of the files, which seemed counterintuitive to me if the `_schemas` directory contains subdirectories).


internal/storage/git/store.go
```diff
@@ -389,63 +387,66 @@ func (s *Store) updateIndex(ctx context.Context) error {
...
+func (s *Store) normalizePath(path string) (string, util.IndexedFileType) {
+	if path == "" {
+		return path, util.FileTypeNotIndexed
 	}
 
-	if util.IsSupportedTestFile(ce.Name) {
-		s.log.Infow("Ignoring test file", "path", ce.Name)
-		return nil
+	if s.conf.SubDir != "" {
+		relativePath := strings.TrimPrefix(path, strings.TrimSuffix(s.conf.SubDir, "/")+"/")
```
charithe, line 433: Would this break on Windows where the separator is `\`? Shouldn't we call `filepath.ToSlash` first?

charithe (approved):
LGTM. Nice work catching all the inconsistencies and bugs between different drivers 🙌🏽

internal/storage/git/store.go
haines: Yeah, good point. `path` is sourced from `ChangeEntry.Name` which [is slash-separated](https://pkg.go.dev/github.com/go-git/go-git/v5/plumbing/object?utm_source=gopls#ChangeEntry.Name), but `s.conf.SubDir` could conceivably be backslash-separated.

charithe (approved):
👍🏽

PR #1003, opened 2022-06-13
enhancement!: Normalise query plan filterDebug field

### LINKED ISSUE #964: PlanResources API: "FilterDebug" should reflect normalised AST
An AST debug string representation is currently built before the AST is normalised. Hence in the string representation variables are not expanded, and logical expressions are not optimised.

Ideally, the AST debug string should reflect the final AST.

Additionally,  `NOT true` and `NOT false` expressions have to be optimised.

charithe:
Generates the `filterDebug` field from the normalised filter instead of
the raw AST.

This change makes the `filterDebug` output a bit more abstract by
producing S-expressions instead of the more CEL-like output before. The
previous output wasn't pure CEL anyway, so this change shouldn't be too
disruptive.

Fixes #964

### DIFF
--- a/api/genpb/cerbos/private/v1/hashpb_helpers.pb.go
+++ b/api/genpb/cerbos/private/v1/hashpb_helpers.pb.go
@@ -1302,6 +1302,29 @@ func cerbos_private_v1_IndexBuilderTestCase_hashpb_sum(m *IndexBuilderTestCase, 
 	}
 }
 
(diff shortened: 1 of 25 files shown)


dbuduev (approved):
LGTM 👍🏻 
Infix notation is more familiar to many users, but S-expressions can also help make sense of AST. 

P.S. I've noticed that later we can also add optimisation of merging nested expressions like: `(AND (AND x y) z)` to `(AND x y z)`.

PR #1009, opened 2022-06-15
fix: Combine conditions from principal and resource policies in query plans

### LINKED ISSUE #1004: PlanResources doesn't fall back to resource policies when principal policy is present
### Current Behavior

If a principal policy exists, PlanResources returns an "always denied" plan for any action not specified in the principal policy.

### Expected Behavior

PlanResources should fall back to the resource policies for any action not specified in the principal policy.

Further, if an action is conditionally allowed/denied in the principal policy, then the plan needs to be combined with the plan from the resource policy.

### Steps To Reproduce

https://play.cerbos.dev/p/h3sAARrHxkY62a86a75ycxx6mcVlNjlb

The `view` action is always allowed in the resource policy, and not overridden in the principal policy for `user@example.com`, so it should still be always allowed for this principal. CheckResources correctly allows the action but PlanResources produces an "always denied" plan.

(example using Ruby SDK)

```ruby
client = Cerbos::Client.new(
  "demo-pdp.cerbos.cloud",
  tls: Cerbos::TLS.new,
  playground_instance: "h3sAARrHxkY62a86a75ycxx6mcVlNjlb"
)

client.plan_resources(
  principal: { id: "user@example.com", roles: ["USER"] },
  resource: { kind: "example_resource" },
  action: "view"
).always_denied? # => true

client.allow?(
  principal: { id: "user@example.com", roles: ["USER"] },
  resource: { kind: "example_resource", id: "123" },
  action: "view"
) # => true
```

### Environment

```markdown
Cerbos version: 0.17.0
```

### Anything else?

The problem is here:

https://github.com/cerbos/cerbos/blob/8a2be23cbfd7715af5b6f9a677003587f7b7be40/internal/engine/engine.go#L308-L310

`plan` can never be `nil` because `EvaluateResourcesQueryPlan` returns an always-denied plan if nothing matches, so we always return.

We need to check if `plan` is conditional at this point, and the condition needs to be combined with the plan from the resource policies. I don't think we can do this combination at the AST level because we have lost the information about which nodes came from allow rules and which came from deny; we need an intermediate representation that preserves that information so we know whether to `AND` or `OR` the conditions.

haines:
Fixes #1004 

This PR changes the query planner to handle the combination of principal and resource policies correctly.

The logic is essentially to nest the query plan derived from the resource policy as one of the allow conditions in the query plan derived from the principal policy, i.e.

```
(and
  (not principal_policy_deny_rule_condition_1)
  (not principal_policy_deny_rule_condition_2)
  ...
  (or
    principal_policy_allow_rule_condition_1
    principal_policy_allow_rule_condition_2
    ...
    (and
      (not resource_policy_deny_rule_condition_1)
      (not resource_policy_deny_rule_condition_2)
      ...
      (or
        resource_policy_allow_rule_condition_1
        resource_policy_allow_rule_condition_2
        ...
      )
    )
  )
)
```

I've also changed the behaviour in the face of missing keys in expressions (i.e. attempting to evaluate a condition like `P.attr.foo.bar` where the input principal attributes don't include `foo.bar`) and expressions that don't return booleans. Currently these cause errors in PlanResources but not in CheckResources (where the conditions are just treated as not matching). To mimic the CheckResources behaviour, these cases will now produce `false` nodes in the AST rather than returning an error.


internal/engine/planner.go
```diff
@@ -32,105 +31,152 @@ type (
...
+	if principalPolicyPlan.Empty() {
+		return resourcePolicyPlan
+	}
+
+	if resourcePolicyPlan.Empty() {
+		return principalPolicyPlan
+	}
+
+	return &PolicyPlanResult{
+		Scope:       fmt.Sprintf("principal: %q; resource: %q", principalPolicyPlan.Scope, resourcePolicyPlan.Scope),
```
haines, line 52: I'm not 100% happy with packing information into a string like this rather than using a struct, but it's only used in the optional response metadata so it doesn't seem worth a breaking API change? 🤷‍♂️

charithe: Aaargh! This is something I hadn't considered before. Usually there would be only one applicable scope but that changes here, doesn't it?

I think what you've done here is the best way to work around it. It's mostly for informational purposes anyway.

charithe (approved):
LGTM 👍🏽

PR #1010, opened 2022-06-15
enhancement: Allow multiple rules per action in principal policies

### LINKED ISSUE #1008: Principal policies should allow multiple rules for the same action
Resource policies allow multiple rules for the same action. If more than one rule matches, `EFFECT_DENY` wins.

Principal policies do not have the same behaviour. If more than one rule is defined for an action, the last rule overrides all the previous ones.

This is because we compile the rules to a map:

https://github.com/cerbos/cerbos/blob/8a2be23cbfd7715af5b6f9a677003587f7b7be40/internal/compile/compile.go#L389

We should instead compile the rules to a list and apply the same logic as with resource policies.

haines:
Fixes #1008


charithe (approved):
LGTM 👍🏽

PR #1012, opened 2022-06-15
enhancement: Deduplicate operands to and/or operations

haines:
I noticed a couple of test cases with redundant `(and x x)` clauses, thought it might be nice to optimise those out.


charithe (approved):
Very nice! 🙌🏽

PR #1013, opened 2022-06-16
enhancement: Validate schemas when planning queries

### LINKED ISSUE #961: Validate principal and partial resource with schemas in PlanResources
PlanResources should validate the principal and resource attributes if there are schemas configured in the policy.

The resource attributes are a partial description of a resource, so we should probably override `required` if it is set on the schema to avoid returning a validation error when attributes are deliberately omitted because they should be part of the query plan.

haines:
Fixes #961 

This PR adds schema validation capability to PlanResources.

If schema validation fails and Cerbos is configured to reject requests on validation failure, then PlanResources returns an always-deny plan.

The validation errors are now recursively added to the list, so that you get full details of failures in nested objects.

**Before**
```
/billing_address: doesn't validate with 'cerbos:///address.json#'
```

**After**
```
/billing_address: missing properties: 'state'
```

For the resource schemas, `required` failures are ignored because the missing attributes will be included in the query plan.


dbuduev (approved):
LGTM 👍🏻

PR #1019, opened 2022-06-17
chore(ci): Split tests across multiple runners

haines:
Running tests currently takes ~11min when we miss the build cache and ~5min on cache hit. This is the rate-limiting step for the workflow

This PR speeds up CI by splitting the test suite across multiple runners, and by adding restore keys to allow partial cache hits. Afterward this change, the total duration is ~3min.

We'll have to change the branch protection rules to require the "Upload test times" job rather than "Test" because GitHub doesn't handle matrix jobs very nicely in the protection rules (you have to add every matrix entry separately).


charithe (approved):
🤯

PR #1032, opened 2022-06-20
fix: Transform maps to lists for membership test  

dbuduev:
#### Description

Membership test operation `in` can be used to test if a key exists in a map. 
The produced AST currently contains the entire struct, which is confusing. It is expected the map to be converted to a list of keys.

#### Checklist 

### DIFF
--- a/internal/engine/ast.go
+++ b/internal/engine/ast.go
@@ -240,6 +240,28 @@ func convert(expr *enginev1.PlanResourcesAst_Node, acc *enginev1.PlanResourcesFi
 	return nil
 }
 
+func mkConstStringExpr(s string) *exprpb.Expr {
+	return &exprpb.Expr{
+		ExprKind: &exprpb.Expr_ConstExpr{
+			ConstExpr: &exprpb.Constant{
+				ConstantKind: &exprpb.Constant_StringValue{
+					StringValue: s,
+				},
+			},
+		},
+	}
+}
+
+func mkListExpr(elems []*exprpb.Expr) *exprpb.Expr {
+	return &exprpb.Expr{
+		ExprKind: &exprpb.Expr_ListExpr{
+			ListExpr: &exprpb.Expr_CreateList{
+				Elements: elems,
+			},
+		},
+	}
+}
+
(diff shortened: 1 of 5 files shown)


internal/engine/ast.go
```diff
@@ -357,19 +383,43 @@ func buildExpr(expr *exprpb.Expr, acc *enginev1.PlanResourcesFilter_Expression_O
 		}
 	case *exprpb.Expr_StructExpr:
 		x := expr.StructExpr
+		if c := parent.GetCallExpr(); c != nil {
+			// convert struct to a list, if it is a key membership test operation
+			const nArgs = 2
+			const rhsIndex = 1 // right-hand side arg index
+			if c.Function == operators.In && len(c.Args) == nArgs && c.Args[rhsIndex] == cur {
```
charithe, line 390: Are there other functions similar to `in` that we should be handling here?

charithe (approved):
This seems to fix the original issue with `in`. I am not sure it handles other cases though. For example, if I change the test expression to `P.attr.teams.exists(t, t.startsWith(R.attr.teamId))` then the generated filter doesn't seem quite right to me.

```yaml
condition:
  expression:
    operands:
    - expression:
        operands:
        - expression:
            operands:
            - value: team1
            - value: employee
            operator: set-field
        - expression:
            operands:
            - value: team2
            - value: user
            operator: set-field
        operator: struct
    - expression:
        operands:
        - expression:
            operands:
            - variable: t
            - variable: request.resource.attr.teamId
            operator: startsWith
        - variable: t
        operator: lambda
    operator: exists
```

I appreciate that this is a complicated problem to address. I am OK with merging this as a quick fix but we probably need to spend a bit more time fixing the other cases as well.

internal/engine/ast.go
dbuduev: Good catch. From the top of my head, other map-related functions/macros include `all` and `exists_one`, which can be transformed similarly.

dbuduev: I prefer to implement them in a separate PR.

PR #1038, opened 2022-06-21
enhancement: Output detailed index build errors when the number of errors is low

### LINKED ISSUE #441: Output detailed index rebuild errors when the number of errors is low
When the server rebuilds the policy index and detects any errors, it only outputs a summary of the errors detected (`failed to build index: missing imports=1, duplicate definitions=0, load failures=2`). This is for two reasons:
- We expect users to always run `cerbos compile` to catch any problems _before_ pushing any policy changes live. 
- If there are too many errors, reporting them all in the server logs will just add a lot of noise and make it harder to read. 

We should improve this error message as follows:
- If the number of errors is reasonably low (say, less than 5) output them to the logs directly
- If the number of errors is too high, print the summary as before and add a note to inform the users that they should run `cerbos compile` to get the full error list

haines:
Fixes #441 

When no more than 5 errors occur during index build, or when debug logging is enabled, the full error details will now be printed instead of just the summary.

e.g. when starting server the output is now
```
2022-06-21T10:59:32.594+0100    ERROR   cerbos.index    Index build failed      {"missing": [{"importingFile":"policy.yaml","desc":"Import 'wat' not found"}]}
cerbos: error: failed to create store: failed to build index: missing imports=1, missing scopes=0, duplicate definitions=0, load failures=0
```

or if more than 5 errors occur

```
2022-06-21T11:38:49.002+0100    ERROR   cerbos.index    Index build failed      {"error": "too many errors; run `cerbos compile` to see a full list"}
cerbos: error: failed to create store: failed to build index: missing imports=3, missing scopes=0, duplicate definitions=0, load failures=3
```

### DIFF
--- a/cmd/cerbos/compile/compile.go
+++ b/cmd/cerbos/compile/compile.go
@@ -14,6 +14,7 @@ import (
 	"github.com/alecthomas/kong"
 	"github.com/fatih/color"
 	"github.com/pterm/pterm"
+	"go.uber.org/zap"
 
 	policyv1 "github.com/cerbos/cerbos/api/genpb/cerbos/policy/v1"
 	internalcompile "github.com/cerbos/cerbos/cmd/cerbos/compile/internal/compilation"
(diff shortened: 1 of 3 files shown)


charithe (approved):
🙌🏽

PR #1044, opened 2022-06-23
fix: Transform maps to lists in comprehension expressions

dbuduev:
#### Description

Expressions like `exists`, `all`, and `exists_one` can be applied to maps, but only map keys are used in the expressions. The resource plan API hence can contain a list instead of a map in the produced AST. That makes the AST much easier to parse.

#### Checklist 

### DIFF
--- a/internal/engine/ast.go
+++ b/internal/engine/ast.go
@@ -297,6 +297,7 @@ func mkListExpr(elems []*exprpb.Expr) *exprpb.Expr {
 		},
 	}
 }
+
 func mkExprOpExpr(op string, args ...*enginev1.PlanResourcesFilter_Expression_Operand) *enginev1.PlanResourcesFilter_Expression_Operand_Expression {
 	return &enginev1.PlanResourcesFilter_Expression_Operand_Expression{
 		Expression: &enginev1.PlanResourcesFilter_Expression{Operator: op, Operands: args},
@@ -452,7 +453,7 @@ func buildExprImpl(cur *exprpb.Expr, acc *enginev1.PlanResourcesFilter_Expressio
 		}
 		acc.Node = mkExprOpExpr(Struct, operands...)
 	case *exprpb.Expr_ComprehensionExpr:
-		lambdaAst, err := BuildLambdaAST(expr.ComprehensionExpr)
+		lambdaAst, err := buildLambdaAST(expr.ComprehensionExpr)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


internal/engine/ast.go
```diff
@@ -449,57 +447,33 @@ func buildExprImpl(cur *exprpb.Expr, acc *enginev1.PlanResourcesFilter_Expressio
...
-		case operators.Conditional:
-			switch x.AccuInit.ExprKind.(type) {
-			case *exprpb.Expr_ListExpr:
-				operator = Filter
-			case *exprpb.Expr_ConstExpr:
-				operator = ExistsOne
-			default:
-				return fmt.Errorf("expected loop-accu-init expression type ConstExpr or ListExpr, got: %T", x.AccuInit.ExprKind)
+		lambdaAst, err := BuildLambdaAST(expr.ComprehensionExpr)
+		iterRange := lambdaAst.IterRange
```
haines, line 451: I think this would panic if `BuildLambdaAST` returns `nil, err` - we should probably move the error handling up above this line.

internal/engine/ast.go
```diff
@@ -253,14 +253,39 @@ func mkConstStringExpr(s string) *exprpb.Expr {
...
+	sort.Slice(elems, func(i, j int) bool {
+		return elems[i].SortKey < elems[j].SortKey
+	})
+
+	exprs := make([]*exprpb.Expr, len(elems))
+	for i, elem := range elems {
+		exprs[i] = elem.Expr
 	}
+
+	return &exprpb.Expr_CreateList{Elements: exprs}
```
haines, line 288: Might make sense to move the construction of the list expression in here too? Both places we use `structKeys` we immediately wrap it like 

```go
&exprpb.Expr{
	ExprKind: &exprpb.Expr_ListExpr{
		ListExpr: structKeys(x),
	},
}
```

internal/engine/lambda.go
```diff
@@ -0,0 +1,57 @@
...
+package engine
+
+import (
+	"fmt"
+
+	"github.com/google/cel-go/common/operators"
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
+)
+
+type LambdaAST struct {
```
haines, line 13: Do this struct and `BuildLambdaAST` need to be public?

internal/engine/ast.go
dbuduev: I wanted to highlight the returning type of the structKeys function. I've refactored this snippet a bit.

haines (approved):

PR #1046, opened 2022-06-24
chore: Add function to compile a set of files

charithe:
Adds a utility function to compile a set of policy files and return the compiled artefacts.


private/compile/compile.go
```diff
@@ -0,0 +1,101 @@
...
+				}
+			} else {
+				log.Debug("Compilation succeeded")
+			}
+
+			log.Debug("Sending artefact")
+			select {
+			case outChan <- artefact:
+				log.Debug("Artefact sent")
+			case <-ctx.Done():
```
haines, line 93: We return `nil` for `err` in this case, should we return `context.Canceled`?

haines: The consumer could check ctx.Err (easy to forget, though). Otherwise maybe we could return two channels, the other being  `<-chan error` that receives nil or err?

charithe: Thinking about this a bit more, the context used here is passed by the consumer. So, if the context is cancelled, it has to be the consumer doing it. In that case, I don't think we need to worry about signalling context cancellation back to the consumer. WDYT?

haines: Yeah, I guess I'm thinking about if the context can be cancelled from outside the direct caller of this function... but in that case I suppose you should just do 

```go
artefacts, err := compile.Files(ctx, fsys)
select {
case <-artefacts:
	// ...
case <-ctx.Done:
	// ...
}
```

haines (approved):

PR #1068, opened 2022-07-08
chore(refactor): Split store interface

charithe:
Split the store interface by functionality so that we can support
different kinds of stores.


internal/storage/store.go
```diff
@@ -92,23 +93,35 @@ func NewFromConf(ctx context.Context, confWrapper *config.Wrapper) (Store, error
...
+	// GetCompilationUnits gets the compilation units for the given module IDs.
+	GetCompilationUnits(context.Context, ...namer.ModuleID) (map[namer.ModuleID]*policy.CompilationUnit, error)
+	// GetDependents returns the dependents of the given modules.
+	GetDependents(context.Context, ...namer.ModuleID) (map[namer.ModuleID][]namer.ModuleID, error)
 	// LoadPolicy loads the given policy from the store
 	LoadPolicy(context.Context, ...string) ([]*policy.Wrapper, error)
 }
 
+// BinaryStore is implemented by stores that have pre-compiled policies in binary format.
+type BinaryStore interface {
```
dbuduev, line 119: AFAIU the `BinaryStore` interface is used only as a policy loader, never as a store (binary store or not). Is it intended for future use?

internal/storage/internal/tests.go
```diff
@@ -18,9 +18,12 @@ import (
 // MutateStoreFn points to a function which mutates the store (ex: add, delete a policy).
 type MutateStoreFn func() error
 
-func TestSuiteReloadable(store storage.ReloadableStore, addFn, deleteFn MutateStoreFn) func(*testing.T) {
+func TestSuiteReloadable(store storage.Store, addFn, deleteFn MutateStoreFn) func(*testing.T) {
```
dbuduev, line 21: I'm wondering if the first parameter can be of type `storage.Reloadable`. The function seems to be used in storage tests which instantiate structs implementing `storage.Reloadable`.

dbuduev (approved):
LGTM. There's a minor comment.

internal/storage/store.go
charithe: It embeds `Store` so it's a store as well. It's just that the scope of `Store` has been reduced to include only the fundamental methods. Other "helper" methods are now in specialised interfaces.

PR #1091, opened 2022-07-19
chore(ci): Use pull request title to apply labels

### LINKED ISSUE #773: Use conventional commit type in PR title to apply corresponding label
e.g. `feat: ...` PRs should have the `feature` label

haines:
Fixes #773 

This PR adds a job to CI to apply label(s) to the PR depending on the Conventional Commits prefix in the title.

To avoid erroneously removing manually-applied labels, it doesn't attempt to remove labels, so if the prefix changes it might be necessary to manually remove outdated labels. I didn't feel like it was worth the extra complexity to attempt to remove labels because I don't think it's common for the prefix to swap between types.

I also removed `ci` and `test` from the allowed types, because we tend to use `chore(ci)` and `chore(test)` instead.

### DIFF
--- a/.github/workflows/pr-title.yaml
+++ b/.github/workflows/pr-title.yaml
@@ -32,13 +32,11 @@ jobs:
         with:
           types: |
             chore
-            ci
             docs
             enhancement
             feat
             fix
             revert
-            test
           requireScope: false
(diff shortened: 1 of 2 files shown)


charithe (approved):
👍🏽

PR #1125, opened 2022-08-02
enhancement: Use plan decorator to provide consistent time functions

### LINKED ISSUE #1123: Extending the CEL environment is expensive
Because the CEL library deprecated `programOptions.Functions` option (#1083), we switched to `env.Extend` to support session-consistent timestamps for `now` and `timeSince` functions. However, `env.Extend` seems to be quite expensive because it's copying the type registry each time.

cerbos cpu_2022-08-01_1550-to-2022-08-01_1555

charithe:
Using `env.Extend` to override the time functions on every program is
expensive because it needs to copy the type registry.

This patch uses a decorator to rewrite calls to `now` and `timeSince`
with a timestamp captured at the start of the program.

Fixes #1123

### DIFF
--- a/internal/conditions/cerbos_lib.go
+++ b/internal/conditions/cerbos_lib.go
@@ -10,12 +10,11 @@ import (
 	"time"
 
 	"github.com/google/cel-go/cel"
-	"github.com/google/cel-go/checker/decls"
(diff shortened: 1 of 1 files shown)


dbuduev (approved):
LGTM.

PR #1127, opened 2022-08-03
chore(ci): Use Go 1.19 in CI

charithe:


charithe:
Have to wait for Golangci-lint to update 😞

PR #1169, opened 2022-08-16
feat: Allow policy tests to use deterministic time functions

### LINKED ISSUE #1167: Allow `time.Now()` to be overriden during policy tests

### Feature description

Cerbos allows [timestamp conditions](https://docs.cerbos.dev/cerbos/latest/policies/conditions.html#_timestamps) as part of the rules in a Policy, which can not currently be tested deterministically because [time.Now](https://github.com/cerbos/cerbos/blob/33fccbfce60d077ee633a75294f218fba80399d6/internal/conditions/cerbos_lib.go#L156) is used.

[Testing policies](https://docs.cerbos.dev/cerbos/latest/policies/compile.html#_testing_policies) should allow the ability to override the time so that rule tests can be deterministically tested. Take the example `timestamp(P.attr.some_claim).timeSince() < duration("1h")` a passing test will fail after 1hr.

I'm assuming for now that this is dictated by https://github.com/cerbos/cerbos/blob/33fccbfce60d077ee633a75294f218fba80399d6/internal/conditions/cerbos_lib.go#L156

Be able to provide a specific timestamp to execute the test with, here I've supplied **now** as an RFC 3339 string.

```yaml
tests:
  - name: Check access was granted in the last hour
    now: "2022-08-15T14:49:03+00:00"
    input:
      principals:
        - some-principal
      resources:
        - some-resource
      actions:
        - some-action
    expected:
```

### Anything else?

_No response_

charithe:
Allows policy tests to specify a static timestamp that should be returned by the calls to the `now` function. This also affects the `timeSince` function because it uses the value of `now` to calculate the time elapsed.

Value of `now` can be set globally for the entire test suite as follows:

```yaml
name: TestSuite
description: Tests for verifying something
options:
  now: "2022-08-02T15:00:00Z"
tests:
...
```

Value can be set per test as well. If both global and local values are defined, local value takes precedence.

```yaml
tests:
  - name: With local now
    options:
      now: "2022-08-03T15:00:00Z"
    ...
```

Fixes #1167


haines (approved):
Nice!

PR #1171, opened 2022-08-18
enhancement: Warn if (and only if) using the default admin password

haines:
I noticed that the server warns about a security risk on the admin API if you continue to use `cerbos` as the username, even if you change the password. I don't think that should produce a warning on the basis that security comes from a strong password with a hard-to-crack hash, not the username being hard to guess.

Currently, if you change the password to an different hash of the same plaintext (e.g. by copy-pasting from the [docs](https://docs.cerbos.dev/cerbos/latest/configuration/server.html#_generating_a_password_hash)), it doesn't recognise that you're still using the default password, so you can avoid the warning without actually improving the security.

This PR changes the check to compare the password plaintext and ignore the username. The check is slower than the direct hash comparison due to the bcrypt cost, so it now happens in a separate goroutine to avoid delaying server start-up.

### DIFF
--- a/internal/server/conf.go
+++ b/internal/server/conf.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"go.uber.org/multierr"
+	"golang.org/x/crypto/bcrypt"
 
 	"github.com/cerbos/cerbos/internal/config"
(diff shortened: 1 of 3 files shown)


charithe (approved):
LGTM 👍🏽

PR #1182, opened 2022-08-29
chore: Refactor private/verify package

dbuduev:
Allow passing engine implementation to the private verifier.

### DIFF
--- a/internal/verify/verify.go
+++ b/internal/verify/verify.go
@@ -25,12 +25,12 @@ type Config struct {
 
 var ErrTestFixtureNotFound = errors.New("test fixture not found")
 
-type InputChecker interface {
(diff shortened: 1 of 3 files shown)


internal/verify/verify.go
```diff
@@ -24,8 +25,12 @@ type Config struct {
 
 var ErrTestFixtureNotFound = errors.New("test fixture not found")
 
+type InputChecker interface {
```
charithe, line 28: Let's call this just `Checker`.

charithe (commented):
LGTM. Let's call the interface `Checker` because `Input` is implied.

charithe (approved):
👍🏽

PR #1198, opened 2022-09-07
fix: Query planner partially evaluates a lambda body

### LINKED ISSUE #1192: Produce better error message when query plan is impossible
For example, given the condition `P.attr.permissions.folder.read.exists(f, R.attr.folder_path.contains(f))`, it's impossible to generate a query plan without knowing the value of `R.attr.folder_path`. This results in the following error message which is not very helpful.

```
error evaluating condition \"P.attr.permissions.folder.read.exists(f, R.attr.folder_path.contains(f))\": no such attribute: id: 4, attributes: [0xc000f2ad20 0xc000f2acc0 0xc000f2ac60]
```

dbuduev:
Fixes #1192

### DIFF
--- a/internal/engine/planner.go
+++ b/internal/engine/planner.go
@@ -581,7 +581,7 @@ func evalComprehensionBody(env *cel.Env, pvars interpreter.PartialActivation, e 
 		}
 		le = ResidualExpr(ast, det)
 		loopStep.CallExpr.Args[i] = le
-		err = evalComprehensionBody(env1, pvars, le)
+		err = evalComprehensionBody(env1, activation, le)
 		if err != nil {
 			return err
 		}
--- a/internal/engine/planner.go
+++ b/internal/engine/planner.go
@@ -569,17 +569,19 @@ func evalComprehensionBody(env *cel.Env, pvars interpreter.PartialActivation, e 
 		}
 		updateIds(le)
 		ast := cel.ParsedExprToAst(&exprpb.ParsedExpr{Expr: le})
-		partialVars, err := cel.PartialVars(pvars, cel.AttributePattern(ce.IterVar))
+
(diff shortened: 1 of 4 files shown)


internal/engine/planner.go
```diff
@@ -569,17 +569,19 @@ func evalComprehensionBody(env *cel.Env, pvars interpreter.PartialActivation, e
 		}
 		updateIds(le)
 		ast := cel.ParsedExprToAst(&exprpb.ParsedExpr{Expr: le})
-		partialVars, err := cel.PartialVars(pvars, cel.AttributePattern(ce.IterVar))
+
+		unknowns := append(pvars.UnknownAttributePatterns(), cel.AttributePattern(ce.IterVar))
```
haines, line 573: If I'm reading above correctly, we don't allow just any unknown variable, it should just be `request.resource` or `R`, right? I think the same should apply here - if you introduce a random `x` that should be an error, but accessing a resource attribute should return an AST.

dbuduev: Short version: Lambda parameter `x` is unknown in the context of lambda body evaluation. For example, `P.roles.exists(x, x.startsWith("admin"))`. Inside the inner expression `x.startsWith("admin")` the `x` is unknown.

Longer version:
When `evalComprehensionBody` is invoked the first time, the `pvars` contains `R`, `request`, `request.resource` in the unknowns list. In this code block, `ce` is a comprehension expression. 
For example `ce = P.roles.exists(x, x.startsWith("admin"))`, `le = x.startsWith("admin")`, `ce.IterVar = x`. In other words. `ce` is lambda, `x` is a lambda parameter, `le` is a lambda body.
We evaluate the lambda body here, so we must add `x` to the list of unknowns.

haines: Thanks, I was confused about the purpose of `pvars.UnknownAttributePatterns()` but it makes sense now 🙂

haines (approved):

PR #1202, opened 2022-09-09
enhancement: Allow specification of gRPC request metadata keys to be logged

### LINKED ISSUE #1201: Allow users to choose which metadata keys are logged to access logs
We currently log all the request metadata in the Audit log access entry. Most of this metadata is not useful and just adds noise to the logs. Instead, we should allow users to specify the metadata keys they are actually interested in capturing.

oguzhand95:
Adds `excludeMetadataKeys` and `includeMetadataKeys` to `audit` configuration.

```
audit:
  accessLogsEnabled: true
  excludeMetadataKeys: ['content-type', 'user-agent']
  includeMetadataKeys: ['content-type', 'user-agent']
```

Here is description of how it works:
- If both `excludeMetadataKeys` and `includeMetadataKeys` has a certain key, the key will *NOT* be logged.
- If only `excludeMetadataKeys` has a certain key, the key will *NOT* be logged.
- If only `includeMetadataKeys` has a certain key, the key will be logged.

resolves #1201

### DIFF
--- a/internal/audit/conf.go
+++ b/internal/audit/conf.go
@@ -22,6 +22,10 @@ type Conf struct {
 type confHolder struct {
 	// Backend states which backend to use for Audits.
 	Backend string `yaml:"backend" conf:",example=local"`
+	// IncludeMetadataKeys defines which grpc request metadata keys are going to be included in the logs.
+	IncludeMetadataKeys []string `yaml:"includeMetadataKeys" conf:",example=['content-type']"`
+	// ExcludeMetadataKeys defines which grpc request metadata keys are going to be excluded from the logs. Takes precedence over IncludeKeys.
+	ExcludeMetadataKeys []string `yaml:"excludeMetadataKeys" conf:",example=['content-type']"`
 	// Enabled defines whether audit logging is enabled.
 	Enabled bool `yaml:"enabled" conf:",example=false"`
 	// AccessLogsEnabled defines whether access logging is enabled.
--- a/internal/audit/conf.go
+++ b/internal/audit/conf.go
@@ -22,9 +22,9 @@ type Conf struct {
 type confHolder struct {
 	// Backend states which backend to use for Audits.
 	Backend string `yaml:"backend" conf:",example=local"`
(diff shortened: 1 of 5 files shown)


internal/audit/interceptor.go
```diff
@@ -47,12 +54,14 @@ func NewUnaryInterceptor(log Log, exclude ExcludeMethod) grpc.UnaryServerInterce
 				StatusCode: uint32(status.Code(err)),
 			}
 
-			md, ok := metadata.FromIncomingContext(ctx)
-			if ok {
-				entry.Metadata = make(map[string]*auditv1.MetaValues, len(md))
-				for key, values := range md {
-					if _, ok := excludeMetadataKeys[key]; !ok {
-						entry.Metadata[key] = &auditv1.MetaValues{Values: values}
+			if !(len(conf.ExcludeMetadataKeys) == 0 && len(conf.IncludeMetadataKeys) == 0) {
```
charithe, line 57: Let's do this check once and save the result to a variable so that it doesn't need to be done on every request.

internal/audit/conf.go
```diff
@@ -22,6 +22,10 @@ type Conf struct {
 type confHolder struct {
 	// Backend states which backend to use for Audits.
 	Backend string `yaml:"backend" conf:",example=local"`
+	// IncludeMetadataKeys defines which gRPC request metadata keys should be included in access logs.
+	IncludeMetadataKeys []string `yaml:"includeMetadataKeys" conf:",example=['content-type']"`
+	// ExcludeMetadataKeys defines which gRPC request metadata keys should be excluded from access logs. Takes precedence over includeMetadataKeys.
+	ExcludeMetadataKeys []string `yaml:"excludeMetadataKeys" conf:",example=['content-type']"`
```
charithe, line 28: I missed this earlier. The example is not quite correct if we use the same key for both include and exclude.  

```suggestion
	ExcludeMetadataKeys []string `yaml:"excludeMetadataKeys" conf:",example=['authorization']"`
```

charithe (commented):
Just a couple of nits to address. LGTM otherwise.

charithe (approved):

PR #1207, opened 2022-09-12
docs: Add clarification about who provides parentRoles

albcunha:
I suggest to making more clear that parent roles are provided by the identity provider. It would be interesting to have some white papers links on the subject. Some libraries do this, such as alibaba (https://antv.vision/en/docs/specification/reference), but I couldn´t find any free white paper that talks about static roles and derived roles.

Signed-off-by: albcunha <albcunha@users.noreply.github.com>

#### Description

Fixes #

#### Checklist 


charithe (approved):
Thanks for the PR. 

I am not aware of academic papers or specifications that discuss derived roles. It's a feature we came up with independently.

PR #1217, opened 2022-09-15
fix: Purge compile cache on store reload

### LINKED ISSUE #1215: Store reload should purge compile cache
Looks like the store reload (at least for disk store) doesn't flush the compile cache.

charithe:
Purges the compile cache when the store is reloaded using the Admin API.
Previously, the cache was only updated when there were new or deleted
files. If an existing file had changed, that wasn't detected.

This change also makes the database stores reloadable.

Fixes #1215


charithe:
Going to merge this so that E2E tests can run a couple of times before the release.

PR #1235, opened 2022-09-22
enhancement: Group compile requests

charithe:
Group simultaneous requests to compile a policy into a single call.

Also adds more trace spans to the engine to increase visibility.


oguzhand95 (approved):
LGTM

PR #1245, opened 2022-09-26
docs: Fix usage of CEL matches operator

oguzhand95:
#### Description

The global `matches(string, string)` operator is deprecated from the CEL spec in favor of member syntax `x.matches(string)`.


charithe (approved):
Nice work hunting down the root cause of the issue 👍🏽

PR #1248, opened 2022-09-27
chore(deps): Update alpine base image from 3.15 to 3.16

charithe:


oguzhand95 (approved):
LGTM.
Using dependabot for this could be an option too, though I am not sure if it could discover our dockerfiles correctly;
```yaml
package-ecosystem: docker
schedule:
  interval: weekly
labels:
  - "chore"
  - "ci"
  - "bots"
```

charithe:
Good point. From a quick glance, I couldn't find anything about how Dependabot discovers dockerfiles. I wonder whether it's smart enough to detect files that are not necessarily named `Dockerfile`.

PR #1262, opened 2022-10-05
fix: Validate required fields in `PlanResourcesRequest`

haines:
`PlanResourcesRequest` is missing a couple of validation rules, meaning that you get a nil pointer dereference when you omit the `resource` field instead of an invalid argument.


charithe (approved):
👍🏽

PR #1264, opened 2022-10-05
docs: added `Engineering` section with post on container usage

Sambigeara:
Signed-off-by: Sam Lock <sam@swlock.co.uk>

#### Description

Fixes #

#### Checklist 

### DIFF
--- a/docs/modules/engineering/pages/why_does_cerbos_use_containers.adoc
+++ b/docs/modules/engineering/pages/why_does_cerbos_use_containers.adoc
@@ -1,0 +1,26 @@
+include::ROOT:partial$attributes.adoc[]
+
+= Why do I need to run a container to use Cerbos?
+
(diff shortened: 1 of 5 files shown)


docs/modules/engineering/pages/why_does_cerbos_use_containers.adoc
```diff
@@ -0,0 +1,26 @@
+include::ROOT:partial$attributes.adoc[]
+
+= Why do I need to run a container to use Cerbos?
```
haines, line 3: I feel like the justification below is more "Why do I need to run a separate process (/binary/microservice) to use Cerbos?" rather than "Why do I need to run a container"... We offer other installation methods (brew, deb, rpm, pre-built binary, compile from source) where you can run the binary directly on your machine without a Docker container.

Sambigeara: That's a good point, thanks.

PR #1281, opened 2022-10-17
feat: Update list functions to support partial eval

dbuduev:
Update remaining list functions to support partial evaluation for the query planner.


charithe (approved):
LGTM

PR #1341, opened 2022-11-08
chore: Disable CGO for "all" make target

dbuduev:
Follow up to https://github.com/cerbos/cerbos/pull/1336.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ generate-notice: $(GO_LICENCE_DETECTOR)
 confdocs:
 	@ mkdir -p ./internal/confdocs
 	@ go run ./hack/tools/confdocs/confdocs.go > ./internal/confdocs/generated.go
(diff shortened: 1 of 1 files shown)


charithe (commented):
Why just this line and not the line above as well?

charithe:
Yes, and the Go file it generates is only used to generate a documentation page. It doesn't get compiled into the Cerbos binary. Even though there's no harm in disabling CGO, I don't think we need to do it in this instance.

dbuduev:
I understand that the generated utility doesn't get into the cerbos binary, but this is not clear to someone who runs `make all` and gets the deprecation warning because the generated utility imports cerbos:
```bash
$ go version -m generated | grep cerbos
	dep	github.com/cerbos/cerbos	(devel)
```

charithe (approved):
Ah, I see your point now.

PR #1342, opened 2022-11-08
enhancement: Clearer errors when trying to query audit logs

### LINKED ISSUE #1318: Clearer error when trying to query audit log with no backend enabled

### Feature description

If you try and query the audit log without having enabled an audit backend, the error logged from `cerbosctl` is `Audit log backend does not support querying`

It would be more informative to say that you have not enabled/defined an audit backend (and maybe linked to docs?)

### Anything else?

_No response_

oguzhand95:
Resolves #1318

### DIFF
--- a/internal/audit/conf_test.go
+++ b/internal/audit/conf_test.go
@@ -6,9 +6,12 @@ package audit_test
 import (
 	"testing"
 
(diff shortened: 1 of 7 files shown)


charithe (approved):
LGTM

PR #1343, opened 2022-11-08
enhancement: Support zip files in cerbosctl put

### LINKED ISSUE #1330: Support Zip files in `cerbosctl put`
Allow users to load policies from a zip file when they use `cerbosctl put`.

Zip readers already implement the `fs.FS` interface so this should be pretty easy to do.

oguzhand95:
Resolves #1330

### DIFF
--- a/client/model.go
+++ b/client/model.go
@@ -4,6 +4,7 @@
 package client
 
 import (
+	"archive/zip"
 	"errors"
 	"fmt"
 	"io"
@@ -584,6 +585,16 @@ func (ps *PolicySet) AddPolicyFromReader(r io.Reader) *PolicySet {
 	return ps
 }
 
+func (ps *PolicySet) AddPolicyFromZipFile(file *zip.ReadCloser, path string) (*PolicySet, error) {
+	r, err := file.Open(path)
+	if err != nil {
+		return nil, err
+	}
+
+	ps.AddPolicyFromReader(r)
+	return ps, nil
+}
+
 // AddPolicies adds the given policies to the set.
 func (ps *PolicySet) AddPolicies(policies ...*policyv1.Policy) *PolicySet {
 	ps.policies = append(ps.policies, policies...)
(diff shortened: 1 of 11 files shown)


cmd/cerbosctl/put/internal/files/find.go
```diff
@@ -71,6 +78,52 @@ func Find(paths []string, recursive bool, fileType util.IndexedFileType, callbac
 	return nil
 }
 
+func fromZip(path string, fileType util.IndexedFileType, callback callback) error {
```
charithe, line 81: Don't we already have this policy discovery logic defined elsewhere for a `fs.FS`? A `zip.Reader` is also a `fs.FS` so we should be able to avoid this duplication.

client/model.go
```diff
@@ -584,6 +585,16 @@ func (ps *PolicySet) AddPolicyFromReader(r io.Reader) *PolicySet {
 	return ps
 }
 
+func (ps *PolicySet) AddPolicyFromZipFile(file *zip.ReadCloser, path string) (*PolicySet, error) {
```
charithe, line 588: I think it'd be rare to want to pick just a single file from a zip file. So, we could just simplify this to `AddFromZipArchive` with just the path to the zip file as the argument. 

If someone wants to add a single file from a zip, they can open the zip file themselves, locate the file they want and then call `AddPolicyFromReader` on the policyset.

cmd/cerbosctl/put/internal/files/find.go
oguzhand95: I think the closest thing we have is [this method](https://github.com/cerbos/cerbos/blob/e7eda5e0787a566dabffa315f6c747edbe95f97b/internal/storage/index/builder.go#L82).
The problem with that method is it ignores the schemas, but in our case, we want to be aware of both policies and schemas.

charithe (approved):
LGTM

PR #1350, opened 2022-11-15
fix: Rewrite log level change without Go 1.19 features

charithe:
`atomic.Bool` is only available in Go 1.19 and users on older versions
of Go won't be able to download the SDK because of that. Since
`atomic.Bool` is not crucial for a correct implementation, we can revert
it this time.

### DIFF
--- a/internal/observability/logging/logging.go
+++ b/internal/observability/logging/logging.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"os/signal"
 	"strings"
-	"sync/atomic"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/observability/logging/logging.go
```diff
@@ -93,45 +92,55 @@ func doInitLogging(ctx context.Context, level string) {
...
 			select {
 			case <-ctx.Done():
 				return
 			case <-sigusr1:
-				if !inProgress.Load() {
-					inProgress.Store(true)
-					setLevelForDuration(zapcore.DebugLevel, originalLevel, tmpLogLevelDuration, &inProgress, atomicLevel)
+				if !inProgress {
+					inProgress = true
+					go setLogLevelForDuration(ctx, doneChan, originalLevel, atomicLevel)
```
haines, line 115: It might be quite nice if sending an additional USR1 while in progress would extend the time period rather than dropping the signal?

haines (dismissed):

dbuduev:
Do we need to change `Go` version in the `go.mod`?

charithe:
> Do we need to change `Go` version in the `go.mod`?

Good catch! Didn't realise that Renovate changed it to 1.19.

haines (approved):

PR #1353, opened 2022-11-17
enhancement(planner): Optimize membership test expression having index into struct

dbuduev:
Optimize membership test expression having index into a struct: `"ADMIN" in P.attr.roles[R.attr.location_id]`


charithe (approved):
LGTM

PR #1371, opened 2022-11-29
fix(planner): R.id unresolved

### LINKED ISSUE #1370: Query planner fails on R.id comparison
This is a regression because the problem does not exist in 0.22.0.

```
error evaluating condition \"P.attr.userQid == R.id\": no such attribute: id: 6, attributes: [0xc000a617d0 0xc000a61770]
```

This seems likely to happen because the `Resource` object in the `PlanResources_Input` message does not have an `id` field. But, it does work on 0.22.0 so my guess is that it's something to do with these new lines we introduced: https://github.com/cerbos/cerbos/blob/6988a238635e35851e3cb3535039d03e73b0bfda/internal/engine/planner/planner.go#L547

**Policy**:

```yaml
---
apiVersion: "api.cerbos.dev/v1"
resourcePolicy:
  version: "default"
  resource: "x"
  rules:
    - actions:
        - write
      effect: EFFECT_ALLOW
      roles:
        - USER
      condition:
        match:
          expr: |-
            P.attr.userQid == R.id
```

**Request**:

```shell
cat <<EOF | curl --silent "http://localhost:3592/api/plan/resources?pretty" -d @-
{
  "requestId": "query-plan",
  "resource": {
    "kind": "x"
  },
  "principal": {
    "id": "qid::link_token:link-token-foo",
    "roles": [
      "USER"
    ],
    "attr": {
      "userQid": "qid::user:user-foo"
    }
  },
  "action": "write",
  "includeMeta": true
}
EOF
```

dbuduev:
Fixes #1370 

Signed-off-by: Dennis Buduev [dbuduev@users.noreply.github.com](mailto:dbuduev@users.noreply.github.com)

### DIFF
--- a/internal/engine/planner/planner.go
+++ b/internal/engine/planner/planner.go
@@ -474,12 +474,9 @@ func evaluateConditionExpression(expr *exprpb.CheckedExpr, input *enginev1.PlanR
 		return nil, err
 	}
 
-	e := expr.Expr
-	if len(variables) > 0 {
-		e, err = replaceVars(e, variables)
-		if err != nil {
-			return nil, err
-		}
+	e, err := replaceVars(expr.Expr, variables)
+	if err != nil {
+		return nil, err
 	}
 
 	if m := input.Resource.GetAttr(); len(m) > 0 {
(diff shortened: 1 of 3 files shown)


charithe (approved):
Ah that was it! Thanks.

PR #1385, opened 2022-12-06
feat!: Include metadata in decision logs

charithe:
This change adds a `metadata` field to the decision log entries to store request metadata. Which metadata keys to include/exclude is based on the configuration values of `audit.includeMetadataKeys` and `audit.excludeMetadataKeys`. This is useful for users who don't have access logs enabled but still want to include some metadata fields in the audit log entry for context.

The include/exclude logic has been slightly modified from previous versions. The new logic is:
- no includes, no excludes: no metadata logged
- only includes: only the listed metadata keys logged
- only excludes: _all_ metadata keys except for the excluded ones are logged
- both includes and excludes: only the includes are logged iff they are not excluded as well


internal/audit/metadata.go
```diff
@@ -0,0 +1,107 @@
...
+	case len(exclude) > 0 && len(include) == 0:
+		shouldInclude = func(k string) bool {
+			if _, ok := alwaysExcludeMetadataKeys[k]; ok {
+				return false
+			}
+
+			if _, ok := exclude[k]; ok {
+				return false
+			}
+			return true
```
oguzhand95, line 51:
```suggestion
			_, ok = exclude[k]
			return !ok
```

oguzhand95 (dismissed):
LGTM

PR #1401, opened 2022-12-20
fix: Fix handling of the SIGUSR1 on Windows

### LINKED ISSUE #1400: Fails to run on windows due to USR1 signal handling
Cerbos expects `USR1` signal to temporarily switch the log level to `debug` level, however Windows doesn't have `syscall.SIGUSR1` which causes failure.

```
internal\observability\logging\logging.go:117:33: undefined: syscall.SIGUSR1
```

oguzhand95:
resolves #1400

### DIFF
--- a/internal/observability/logging/logging.go
+++ b/internal/observability/logging/logging.go
@@ -6,9 +6,7 @@ package logging
 import (
 	"context"
 	"os"
-	"os/signal"
 	"strings"
-	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


charithe (approved):
LGTM. 

Since we don't currently have a Windows build, I guess this is only useful for someone who builds from source. Does the Cerbos binary work  correctly on Windows? I think there were a few issues the last time we tried.

PR #1423, opened 2023-01-17
docs: Add note on resource-led policy design in `Best practices` section

Sambigeara:
Signed-off-by: Sam Lock <sam@swlock.co.uk>

### DIFF
--- a/docs/modules/policies/pages/best_practices.adoc
+++ b/docs/modules/policies/pages/best_practices.adoc
@@ -6,9 +6,9 @@ A collection of tips and code snippets designed to help you write cleaner, more 
 
 == Modelling policies
 
-The way you model your policies is up you -- you can achieve the same logical outcome in numerous ways. That said, some patterns will lend themselves more naturally to certain scenarios.
+With Cerbos, access rules are always resource oriented, and the policies you write map to these resources within your system. A _resource_ can be anything, and the way you model your policies is up you -- you can achieve the same logical outcome in numerous ways; action-led, role-led, attribute-led, or with combinations thereof.
 
-Consider this business model:
+That said, some patterns will lend themselves more naturally to certain scenarios -- let's take a look at some different approaches. Consider this business model:
 
 [%noheader,cols="2h,5*^"]
 |===
@@ -98,7 +98,7 @@ This might apply if your scenario doesn't strictly fall into one of the previous
 
 === Blanket allow, granular deny
 
-Finally, we can opt to explicitly state which actions a user *cannot* do:
+We can opt to explicitly state which actions a user *cannot* do:
 
 [source,yaml,linenums]
 ----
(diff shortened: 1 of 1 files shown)


charithe (commented):
I am struggling a bit to understand this new section. My understanding was that this page is about how to approach writing the rules for a resource, which you can do by thinking along the dimensions of actions, roles and so on. I don't think it can be resource-led because  that's a given (we are writing a resource policy, after all).

Would it be better to add an intro section to the document to say the following.
- With Cerbos, access rules are always resource oriented. You must write policies targeting each resource you want to secure.
- Within a policy, there are several approaches you can use to write access rules (action-led, role-led etc.)

Sambigeara:
Yeah, it certainly didn't seem _quite_ right putting it here. The intention was to solve [this user request](https://cerbos.slack.com/archives/C03JN3JM2LC/p1673947852199509), but perhaps this would be better placed elsewhere. We're currently waiting on feedback from the user in the community Slack - I'll wait for that and then consider better alternatives. I like your suggestion on some preliminary notes in an intro.

docs/modules/policies/pages/best_practices.adoc
```diff
@@ -119,6 +119,37 @@ Finally, we can opt to explicitly state which actions a user *cannot* do:
 
 This would suit scenarios where a principal can perform _nearly_ every action, and you want to explicitly list disallowed actions.
 
+=== Attribute-led
+
+Consider the following hypothetical scenario:
+
+****
+An organization models it's resources as specific _data sets_. Each data set is unique, as are the principals trying to access them. The organization uses JWTs extensively to manage and transmit identity/contextual information. The resource policies map 1:1 to each data set, and access is governed by arbitrary information (in this case, passed within the JWT).
```
charithe, line 127:
```suggestion
An organization models its resources as specific _data sets_. Each data set is unique, as are the principals trying to access them. The organization uses JWTs extensively to manage and transmit identity/contextual information. The resource policies map 1:1 to each data set, and access is governed by arbitrary information (in this case, passed within the JWT).
```

docs/modules/policies/pages/best_practices.adoc
```diff
@@ -119,6 +119,37 @@ Finally, we can opt to explicitly state which actions a user *cannot* do:
...
+****
+
+Given the changeable nature of the accessing audiences, it doesn't necessarily make sense to rely on roles for access. What we could do instead is to globally allow all roles and actions and then determine access based on attributes passed in the JWT. Take a look at the following example policy:
```
charithe, line 130:
```suggestion
Given the dynamic nature of audiences, it's not practical to enumerate all roles that have access. What we could do instead is to globally allow all roles and actions and then determine access based on attributes passed in the JWT. Take a look at the following example policy:
```

PR #1426, opened 2023-01-24
feat: Admin API endpoint to disable policy(s)

oguzhand95:
Changes
- Adds `Disable` endpoint to the Admin API.
- Adds `cerbosctl disable policies <id>...` command to `cerbosctl` CLI.
- Fixes the problem related to the `ListPolicies` endpoint not working with the SQLite store.
- Fixes `GetPolicy` responding with the disabled policies.

Planned for upcoming PRs
- Currently, the user must keep in mind the scoped policies with a hierarchy present.
Example: There are three policies with scopes set to `a.b.c`, `a.b`, `a`. If the user wants to disable the policy with the scope `a.b`, the policy with `a.b.c` should be disabled manually, too.
I added a note to the documentation for this.

TBD
- Since policy IDs consist of forward slashes (`/` - ex: `resource.purchase_order.vdefault/regional`), URLs should be encoded for REST calls. This is also true for the `GetPolicy` endpoint.


internal/storage/db/internal/db.go
```diff
@@ -38,6 +38,7 @@ type DBStorage interface {
 	ListPolicyIDs(ctx context.Context) ([]string, error)
 	ListSchemaIDs(ctx context.Context) ([]string, error)
 	AddOrUpdateSchema(ctx context.Context, schemas ...*schemav1.Schema) error
+	EnableOrDisable(ctx context.Context, disabled bool, policyKey ...string) error
```
charithe, line 41: Is the "enable" portion ever used? That would be taken care of by `AddOrUpdate` anyway, wouldn't it? So, I would just call this `Disable`

charithe (commented):
Have you tried this with scoped policies? Couple of things to be aware of with scoped polices are:
- The identifier is of the form `resource.x.v1/scope` which would require the `/` to be URL encoded in the REST request. I wonder whether we should just make this RPC the same as `AddOrUpdate` but let users post empty policies without any rules.
- Because scoped policies require an unbroken hierarchy, we should probably disable all the children of a scoped policy. For example, if the full scope hierarchy is `a.b.c.d` and the user disables `b`, we should automatically disable `c` and `d` as well or the policy repository will be in an invalid state.


Since you're loading the policies from the database and writing them back with the `disabled` field set, you can keep track of the actual policies that were disabled during the call and return them in the RPC response, can't you?

cmd/cerbosctl/disable/policy.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func (c *Cmd) Run(k *kong.Kong, ctx *client.Context) error {
+	if len(c.Policy.PolicyIds) == 0 {
+		return fmt.Errorf("no policy id(s) provided")
+	}
+
+	disabledPolicies, err := policy.Disable(ctx.AdminClient, c.Policy.PolicyIds...)
+	if err != nil {
+		return fmt.Errorf("failed to disable policy(s): %w", err)
```
charithe, line 36:
```suggestion
		return fmt.Errorf("failed to disable policies: %w", err)
```

cmd/cerbosctl/disable/policy.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	_, _ = fmt.Fprintf(k.Stdout, "Successfully disabled %d policy(s)", disabledPolicies)
```
charithe, line 39:
```suggestion
	_, _ = fmt.Fprintf(k.Stdout, "Number of policies disabled is %d", disabledPolicies)
```

docs/modules/api/pages/admin_api.adoc
```diff
@@ -156,6 +156,23 @@ curl -k -u cerbos:cerbosAdmin \
...
+NOTE: This endpoint requires a mutable storage driver such as xref:configuration:storage.adoc#sqlite3[sqlite3] to be configured.
+
+Issue a DELETE request to the endpoint with the list of IDs (the `id` query parameter can be repeated multiple times) to disable.
+
+----
+curl -k -u cerbos:cerbosAdmin -X DELETE \
+    'https://localhost:3592/admin/policy?id=principal.donald_duck.vdefault&id=derived_roles.my_derived_roles'
+----
+
+NOTE: If the policy being disabled has scope, its descendants must be disabled first. Otherwise, the storage index will
```
charithe, line 174:
```suggestion
CAUTION: Scoped policies must have unbroken scope chains. If you're disabling a scoped policy, make sure that its descendant policies are disabled as well. 
```


If a user accidentally breaks a scoped policy chain and then restarts the Cerbos instance, will it refuse to start? If so, they have no choice but to manually edit the database records, right?

docs/modules/cli/pages/cerbosctl.adoc
```diff
@@ -158,6 +159,27 @@ Use the arrow keys (or Vim keys kbd:[h], kbd:[j], kbd:[k], kbd:[l]) to scroll ho
...
+
+.Disable multiple policies
+----
+cerbosctl disable policies derived_roles.my_derived_roles resource.leave_request.default
+cerbosctl disable policy derived_roles.my_derived_roles resource.leave_request.default
+cerbosctl disable p derived_roles.my_derived_roles resource.leave_request.default
+----
+
+NOTE: If the policy being disabled has scope, its descendants must be disabled first. Otherwise, the storage index will
+be corrupt.
```
charithe, line 182:
```suggestion
CAUTION: Scoped policies must have unbroken scope chains. If you're disabling a scoped policy, make sure that its descendant policies are disabled as well.
```

internal/storage/db/internal/db.go
```diff
@@ -464,6 +478,55 @@ func (s *dbStorage) GetDependents(ctx context.Context, ids ...namer.ModuleID) (m
 	return out, nil
 }
 
+func (s *dbStorage) Disable(ctx context.Context, policyKey ...string) (int64, error) {
+	if len(policyKey) == 1 {
```
charithe, line 482: I think that this optimization is not necessary because in practice `disable` should be called very rarely. This just complicates the code for very little benefit.

internal/storage/db/internal/db.go
```diff
@@ -464,6 +478,55 @@ func (s *dbStorage) GetDependents(ctx context.Context, ids ...namer.ModuleID) (m
...
+	if err != nil {
+		return 0, err
+	}
+
+	affected, err := res.RowsAffected()
+	if err != nil {
+		return 0, fmt.Errorf("failed to discover whether the policies got disabled or not: %w", err)
+	} else if affected == 0 {
+		return 0, fmt.Errorf("failed to find the policies for disabling")
+	}
```
charithe, line 524: I don't think this is an error. If you issue the same call twice by accident, the second call would always return 0 rows.

charithe (commented):
LGTM apart from a few nits 👍🏽 

I think it's really important that we prevent scoped policies from being disabled or updated (both via `AddOrUpdate` and `DisablePolicy`) if that breaks the chain because that could make Cerbos unavailable. It's probably worth adding a condition to the `UPDATE` statements to avoid that scenario. Let's do that in this PR itself to make sure we don't forget to do that and accidentally ship this.

docs/modules/api/pages/admin_api.adoc
oguzhand95: Logs when a scoped policy is disabled;
```2023-01-31T11:44:38.126+0300    ERROR   cerbos.compiler Failed to recompile     {"id": {}, "error": "missing policy definition {13106758747738567535}: invalid compilation unit"}```

Cerbos doesn't refuse to start, but I think manual editing of the database records is a must.

charithe: If it still starts, we can use the CLI to fix the issue, can't we? Then that's not a big problem.

charithe (approved):

PR #1441, opened 2023-02-06
enhancement: Prevent scoped policies being disabled

oguzhand95:
store
---
```
{
    "policyIds": [
        "resource.purchase_order.vdefault",
        "resource.purchase_order.vdefault/regional",
        "resource.purchase_order.vdefault/vanilla"
    ]
}
```

requests & responses
---
```
DELETE /admin/policy?id=resource.purchase_order.vdefault
```
```json
{
    "code": 13,
    "message": "Failed to disable policies: removing the following scoped policies will break the scope chain: resource.purchase_order.vdefault"
}
```
---
```
DELETE /admin/policy?id=resource.purchase_order.vdefault&id=resource.purchase_order.vdefault/regional
```
```json
{
    "code": 13,
    "message": "Failed to disable policies: removing the following scoped policies will break the scope chain: resource.purchase_order.vdefault, resource.purchase_order.vdefault/regional"
}
```


internal/storage/db/internal/db.go
```diff
@@ -478,7 +480,97 @@ func (s *dbStorage) GetDependents(ctx context.Context, ids ...namer.ModuleID) (m
 	return out, nil
 }
 
+func (s *dbStorage) GetAncestorsAndDescendants(ctx context.Context, ids ...namer.ModuleID) (map[namer.ModuleID][]namer.ModuleID, error) {
```
charithe, line 483: I think we can make this a little bit more efficient. Right now, for every policy you need to disable, you have to execute two queries and map their values to Go structs. That's 2N queries. Instead, we can run just one query to figure out everything we need to know about the policy set to be disabled.

```sql
SELECT p.id, 
	EXISTS(select 1 FROM policy_ancestor pa1 JOIN policy p1 ON (pa1.policy_id = p1.id AND p1.disabled = false) WHERE pa1.ancestor_id = p.id) AS has_descendants 
FROM policy p 
WHERE p.id IN (?);
```

This will output something like the following:

```
id|has_descendants
-4874326380253481038|0
-311105230672970778|1
1057909666449662682|0
```

If the `has_descendants` column is `0`, we can disable that policy.

haines:
The `disabledPolicies` field in the response is currently `int64`. Would it make sense for this to be a) unsigned and b) smaller, i.e. `uint32`? 64-bit values are a bit of a pain to consume in JS (they have to be encoded as strings).

schema/openapiv2/cerbos/svc/v1/svc.swagger.json
```diff
@@ -2178,7 +2178,7 @@
       "type": "object",
       "properties": {
         "disabledPolicies": {
-          "type": "string",
+          "type": "integer",
           "format": "int64"
```
oguzhand95, line 2182: I would expect this to be `{"type": "integer", format: "uint32"}` 🤔

internal/storage/db/internal/db.go
```diff
@@ -478,7 +480,72 @@ func (s *dbStorage) GetDependents(ctx context.Context, ids ...namer.ModuleID) (m
...
+			),
+		).
+		Where(goqu.C(PolicyAncestorTblAncestorIDCol).Table("pa1").Eq(goqu.C(PolicyTblIDCol).Table("p")))
+
+	// SELECT p.id, EXISTS(<innerQuery>) AS has_descendants
+	// FROM policy p
+	// WHERE p.id IN (?);
+	query := s.db.Select(
+		goqu.C(PolicyTblIDCol).Table("p"),
+		goqu.L("EXISTS ?", innerQuery).As("has_descendants"),
```
charithe, line 505: Can we use [`Func`](https://pkg.go.dev/github.com/doug-martin/goqu/v9#Func) here instead? I think that would convey the idea better.


```suggestion
		goqu.Func("EXISTS", innerQuery).As("has_descendants"),
```

internal/storage/db/internal/db.go
```diff
@@ -478,7 +480,72 @@ func (s *dbStorage) GetDependents(ctx context.Context, ids ...namer.ModuleID) (m
...
+	}
+
+	return out, nil
+}
+
+func (s *dbStorage) Disable(ctx context.Context, policyKey ...string) (uint32, error) {
+	var brokenChainPolicies []string
+	for _, pk := range policyKey {
+		mID := namer.GenModuleIDFromFQN(namer.FQNFromPolicyKey(pk))
+		hasDescendants, err := s.HasDescendants(ctx, mID)
```
charithe, line 535: Instead of running the query for each policy to be disabled, build a list of  module IDs from the `policyKey` slice and call `HasDescendants` once with that list. Then iterate through the result of that call and print the error message for those policies that have descendants.

internal/storage/db/internal/db.go
oguzhand95: `goqu.Func` puts an extra pair of parentheses and postgres doesn't seem to like it.

internal/svc/admin_svc.go
```diff
@@ -180,6 +181,9 @@ func (cas *CerbosAdminService) DisablePolicy(ctx context.Context, req *requestv1
 	disabledPolicies, err := ms.Disable(ctx, req.Id...)
 	if err != nil {
 		ctxzap.Extract(ctx).Error("Failed to disable policies", zap.Error(err))
+		if errors.As(err, &db.ErrBreaksScopeChain{}) {
+			return nil, status.Error(codes.Internal, fmt.Sprintf("Failed to disable policies: %s", err.Error()))
```
charithe, line 185:
```suggestion
			return nil, status.Error(codes.InvalidArgument, err.Error())
```

charithe (approved):
LGTM 👍🏽 
Let's fix the error code returned by the service as commented below.

charithe (approved):
I think you probably need to rebase from main and run `make clean-tools generate`

oguzhand95:
Unfortunately it doesn't work, even though my branch is up-to-date. I even tried manually executing the commands from `generate-mocks` target with the latest version of the `mockery@v2.18.0`. Is there a chance this being a problem related to GitHub cache?

charithe:
Yep. Just zapped it.

PR #1444, opened 2023-02-08
chore: Upgrade Otel semconv version

charithe:
The semconv package from Otel has a secondary version encoded in the
package name. So, even if the library version is updated in go.mod, we
have to remember to manually update the import path too. If not, it
panics because apparently two different "versions" of semconv cannot
co-exist together :shrug:

Did I mention that Otel is the worst library I have ever worked with?


haines (approved):
yuck

PR #1457, opened 2023-02-10
chore(ci): Add merge_group to GitHub workflow

oguzhand95:
I think this is why the queued PRs aren't being merged into `main` branch.

### DIFF
--- a/.github/workflows/pr-test.yaml
+++ b/.github/workflows/pr-test.yaml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - 'v*'
+  merge_group:
 jobs:
   changes:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


charithe (approved):

haines:
I'm not sure that we'll be able to get this working until the DCO GitHub App is updated to support the `merge_group` event (unless we make this check optional)

https://github.com/dcoapp/app/issues/199

.github/workflows/pr-test.yaml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - 'v*'
+  merge_group:
```
Licenser, line 8: saw this as mentioned in the DCO issue, the recommendation from GH is to use `merge_group: ["main"]` :)

Issue #1461, opened 2023-02-13
Cerbosctl missing commands 


charithe:
- List disabled policies (#1463)
- Re-enable disabled policies (#1472)
- Delete schema (#1463)

PR #1463, opened 2023-02-15
enhancement: Add includeDisabled to Admin API and schema deletion to cerbosctl

oguzhand95:
- Added `includeDisabled` field to `GetPolicy` and `ListPolicies` RPCs.
- Added `cerbosctl delete schema <schemaIDs...>` command.
- Updated the documentation to reflect the changes.
- This PR partially handles #1461, namely the first and third points.


api/public/cerbos/request/v1/request.proto
```diff
@@ -541,6 +545,10 @@ message GetPolicyRequest {
       example: "\"principal.sarah.vdefault\""
     }
   ];
+  bool include_disabled = 2 [
+    (google.api.field_behavior) = OPTIONAL,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Include disabled policies"}
+  ];
```
charithe, line 551: Do we need this flag for `GetPolicy`? If the user gives us a valid ID that exists in the database we can simply return it, can't we?

client/admin.go
```diff
@@ -28,8 +28,8 @@ const (
 type AdminClient interface {
 	AddOrUpdatePolicy(ctx context.Context, policies *PolicySet) error
 	AuditLogs(ctx context.Context, opts AuditLogOptions) (<-chan *AuditLogEntry, error)
-	ListPolicies(ctx context.Context) ([]string, error)
-	GetPolicy(ctx context.Context, ids ...string) ([]*policyv1.Policy, error)
+	ListPolicies(ctx context.Context, includeDisabled bool) ([]string, error)
```
charithe, line 31: This would be a breaking change to existing users. I think we should declare it like `ListPolicies(context.Context, ...ListPoliciesOption)`. 

There would be just one `ListPoliciesOption` and that would be the one to include disabled policies in the output.

client/admin.go
```diff
@@ -28,8 +28,8 @@ const (
...
+	GetPolicy(ctx context.Context, includeDisabled bool, ids ...string) ([]*policyv1.Policy, error)
```
charithe, line 32: Probably don't need the `includeDisabled` argument here. See above.

cmd/cerbosctl/del/schema.go
```diff
@@ -0,0 +1,45 @@
...
+cerbosctl delete schema principal.json
+cerbosctl delete s principal.json
+
+# Delete multiple schemas
+cerbosctl delete schemas principal.json leave_request.json
+cerbosctl delete schema principal.json leave_request.json
+cerbosctl delete s principal.json leave_request.json`
+
+type SchemaCmd struct {
+	SchemaIds []string `arg:"" name:"id" optional:"" help:"list of schema ids to delete"`
```
charithe, line 26: This is not `optional`, is it?

internal/storage/db/internal/db.go
```diff
@@ -603,16 +608,21 @@ func (s *dbStorage) Delete(ctx context.Context, ids ...namer.ModuleID) error {
 	return nil
 }
 
-func (s *dbStorage) ListPolicyIDs(ctx context.Context) ([]string, error) {
+func (s *dbStorage) ListPolicyIDs(ctx context.Context, includeDisabled bool) ([]string, error) {
 	var policyCoords []namer.PolicyCoords
+	var whereExprs []exp.Expression
+	if !includeDisabled {
+		whereExprs = append(whereExprs, goqu.C(PolicyTblDisabledCol).Neq(goqu.V(true)))
+	}
```
charithe, line 616: I think this can be simplified. Just always add a WHERE clause to the query.

```suggestion
goqu.C(PolicyTblDisabledCol).Eq(goqu.V(includeDisabled)))
```

client/admin.go
oguzhand95: It would be confusing for the user, wouldn't it? If we are keen on this, at least we could return an error in case more than one `opt` are provided to the function, WDYT?

charithe: Why do you think it would be confusing?

We don't need to worry about how many options are passed to the function. The last one would always win. That's the expected behaviour for functional options in Go.

internal/storage/db/internal/db.go
oguzhand95: Then we will have only disabled policies or enabled policies depending on the `includeDisabled` parameter. But, we actually want to have all policies when `includeDisabled == true`.

charithe: Good point! I completely messed that one up 🤦🏽

client/model.go
```diff
@@ -1225,3 +1225,16 @@ func (e *AuditLogEntry) DecisionLog() (*auditv1.DecisionLogEntry, error) {
 type PlanResourcesResponse struct {
 	*responsev1.PlanResourcesResponse
 }
+
+type (
+	ListPoliciesOption func(opts *ListPoliciesParams)
+	ListPoliciesParams struct {
+		IncludeDisabled bool
+	}
```
charithe, line 1233: I don't think you need this struct. Just define `ListPoliciesOption` as `func(*requestv1.ListPoliciesRequest)`.

client/model.go
```diff
@@ -1225,3 +1225,13 @@ func (e *AuditLogEntry) DecisionLog() (*auditv1.DecisionLogEntry, error) {
...
+
+type (
+	ListPoliciesOption func(*requestv1.ListPoliciesRequest)
+)
+
+func WithIncludeDisabled(value bool) ListPoliciesOption {
+	return func(request *requestv1.ListPoliciesRequest) {
+		request.IncludeDisabled = value
+	}
+}
```
charithe, line 1237: Sorry I missed this earlier. We don't need the boolean argument here because it's false by default anyway. Just name it `WithIncludeDisabled()`.

charithe (commented):
LGTM. Just have a nit about the option signature.

charithe (approved):

PR #1472, opened 2023-02-21
enhancement: Enable disabled policies

### LINKED ISSUE #1461: Cerbosctl missing commands

oguzhand95:
resolves #1461 

I've made `POST /admin/policy/enable` the endpoint, while the disable is `DELETE /admin/policy`. I think if we go forward with the `POST /admin/policy/enable`, we may change the disable endpoint to `POST /admin/policy/disable` to be consistent, not sure.


cmd/cerbosctl/enable/internal/policy/policy.go
```diff
@@ -0,0 +1,29 @@
...
+		if idx%internal.MaxIDPerReq == 0 {
+			idxEnd := internal.MinInt(idx+internal.MaxIDPerReq, len(ids))
+			var err error
+			enabled, err := c.EnablePolicy(context.Background(), ids[idx:idxEnd]...)
+			if err != nil {
+				return 0, fmt.Errorf("error while enabling policy: %w", err)
+			}
+			enabledPolicies += enabled
+		}
+	}
```
charithe, line 26: I see this batching pattern being repeated a lot. Now that Go generics support has landed, I wonder if it could be made into a utility function that can be reused.

cmd/cerbosctl/enable/enable.go
```diff
@@ -0,0 +1,8 @@
+// Copyright 2021-2023 Zenauth Ltd.
```
charithe, line 1: Probably can add this definition to `policy.go` or vice versa. It feels too Java-like to have a single file for a single type 😄

internal/storage/db/internal/db.go
```diff
@@ -570,6 +568,31 @@ func (s *dbStorage) Disable(ctx context.Context, policyKey ...string) (uint32, e
...
 
+func (s *dbStorage) Enable(ctx context.Context, policyKey ...string) (uint32, error) {
+	mIDs := make([]namer.ModuleID, len(policyKey))
+	events := make([]storage.Event, len(policyKey))
+	for idx, pk := range policyKey {
+		mIDs[idx] = namer.GenModuleIDFromFQN(namer.FQNFromPolicyKey(pk))
+		events[idx] = storage.Event{Kind: storage.EventAddOrUpdatePolicy, PolicyID: namer.GenModuleIDFromFQN(namer.FQNFromPolicyKey(pk))}
+	}
+
+	res, err := s.db.Update(PolicyTbl).Prepared(true).
```
charithe, line 579: Can we be sure that the `disabled` field of the policy definition is not `true`? It's possible that someone uses the old `AddOrUpdatePolicy` RPC to disable a policy and then later uses this RPC to enable it. In that case, the policy definition would still remain disabled as far as the Cerbos compiler is concerned.

charithe (commented):
It makes sense to have a corresponding `disable` endpoint as well for symmetry. 

I think the HTTP verb for the REST call should support both `PUT` and `POST` as well.

client/util.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"context"
+
+	"golang.org/x/exp/constraints"
+)
+
+const MaxIDPerReq = 25
+
+func BatchAdminClientCall[T constraints.Integer](fn func(context.Context, ...string) (T, error), ids ...string) (T, error) {
```
charithe, line 14: Looks like all the callbacks just return `uint32`? In that case, we don't even need the generic parameter here. Just define it as:

```suggestion
func BatchAdminClientCall(fn func(context.Context, ...string) (uint32, error), ids ...string) (uint32, error) {
```

cmd/cerbosctl/get/internal/policy/policy.go
```diff
@@ -51,8 +50,8 @@ func List(k *kong.Kong, c client.AdminClient, filters *flagset.Filters, format *
 	fd := newFilterDef(kind, filters.Name, filters.Version, filters.IncludeDisabled)
 
 	for idx := range policyIds {
-		if idx%internal.MaxIDPerReq == 0 {
-			idxEnd := internal.MinInt(idx+internal.MaxIDPerReq, len(policyIds))
+		if idx%client.MaxIDPerReq == 0 {
```
charithe, line 53: Oh we still have to duplicate this logic in some places, don't we? I wanted to keep it in one place so that it doesn't drift apart (i.e. someone modifies it in one place but forgets to do it in the other places). Looks like that's not possible 🤔

charithe (commented):
I think we discussed that the `disabled` column should be the source of truth. In that case, the places where we retrieve policies from the DB (`GetCompilationUnits`, `LoadPolicy`) need to override the stored policy's `disabled` field. I think `LoadPolicy` does that already but not so in `GetCompilationUnits`.

cmd/cerbosctl/get/internal/policy/policy.go
charithe: Fair enough. Even if we do find a way, that itself would become too complicated to make sense of. Let's leave it as it is.

charithe (approved):
LGTM

charithe (approved):
👍🏽

haines:
This is a breaking change for the JS SDK (and anyone else consuming the HTTP API) so perhaps as well as introducing `POST /admin/policy/disable` we should keep (and deprecate) the `DELETE /admin/policy` endpoint?

https://github.com/cerbos/cerbos-sdk-javascript/actions/runs/4292039774/jobs/7477987513#step:5:37

PR #1473, opened 2023-02-22
enhancement: Add support for reading archive files for the disk driver

Sambigeara:
Adds function to attempt retrieval of an `fs.FS` instance from `tar`, `zip` and `gzip` files if they have the matching extensions, otherwise defaults to `os.DirFS`.

TODO
- [x] Swop out for existing os.DirFS() calls where appropriate
- [x] Tidy file extension parsing
- [x] Fix [dirwatch](https://github.com/Sambigeara/cerbos/blob/e20b91705e28d41bb120b10102ef6b28d2779d94/internal/storage/disk/dirwatch.go#L51-L53)
- [x] Docs

### DIFF
--- a/internal/util/filesystem.go
+++ b/internal/util/filesystem.go
@@ -82,30 +82,30 @@ func IsArchiveFile(fileName string) bool {
 	return IsZip(fileName) || IsTar(fileName) || IsGzip(fileName)
 }
 
-func getFsFromTar(r io.Reader) fs.FS {
+func getFsFromTar(r io.Reader) (fs.FS, error) {
 	tfs, err := tarfs.New(r)
 	if err != nil {
-		panic(fmt.Errorf("failed to open tar file: %w", err))
+		return nil, fmt.Errorf("failed to open tar file: %w", err)
 	}
-	return tfs
+	return tfs, nil
 }
 
 // OpenDirectoryFS attempts to open a directory FS at the given location. It'll initially check if the target file is a zip,
 // and if so, will return a zip Reader (which implements fs.FS).
-func OpenDirectoryFS(path string) fs.FS {
+func OpenDirectoryFS(path string) (fs.FS, error) {
 	// We don't use `switch filepath.Ext(path)` here because it only suffixes from the final `.`, so `.tar.gz` won't be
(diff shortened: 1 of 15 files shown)


internal/util/filesystem.go
```diff
@@ -61,8 +66,51 @@ func IsHidden(fileName string) bool {
 	}
 }
 
-func IsZip(fileName string) bool {
-	return strings.HasSuffix(fileName, ".zip")
+func getFsFromTar(r io.Reader) fs.FS {
+	tfs, err := tarfs.New(r)
+	if err != nil {
+		panic(fmt.Errorf("failed to open tar file: %w", err))
```
charithe, line 72: I know it makes the `OpenDirectoryFS` function a bit awkward to use. But, because we are dealing with files and there's good probability that things could go wrong, we should return the errors back to the caller and exit gracefully rather than panicking.

Sambigeara:
Was hoping to wrap this up today, but manual testing just highlighted an issue with the `watchDir` function [here](https://github.com/Sambigeara/cerbos/blob/e20b91705e28d41bb120b10102ef6b28d2779d94/internal/storage/disk/dirwatch.go#L51-L53). Have to step away now, but will fix this (and do more thorough testing) when I pick this up again on Friday.

charithe:
We should probably disable directory watch if the path is not a directory because I don't think it makes sense in the context of an archive file.

Sambigeara:
The code coverage is confusing me a little, as my tests are covering the code paths that codecov seems to think I'm not 🤷‍♂️

charithe (dismissed):
Nice 👍🏽  

Let's add a note to the documentation to say that we assume the archive is laid out in the format of a policy directory (`_schemas` exists at the root) it must not contain any other non-policy YAML files either.

Sambigeara:
The force push to [4f9853b](https://github.com/cerbos/cerbos/commit/4f9853b2319d7d546816826ba8b4297e7d914b14) was just a rebase on `main`

charithe (approved):
The cache of the linter gets out of sync sometimes and that's why you get that failure. I have cleared it now.

PR #1476, opened 2023-02-28
chore: Deprecate old policy disable endpoint

oguzhand95:
Added `DELETE /admin/policy` endpoint back for the `DisablePolicy` RPC. ([related comment](https://github.com/cerbos/cerbos/pull/1472#issuecomment-1447957235))
Since we can't `deprecate` the endpoint, I only changed the docs to indicate `DELETE /admin/policy` is deprecated.
Also, there are some documentation updates.


haines (approved):
Thanks!

Issue #1403, opened 2022-12-24
Grouping test results by principal does not update the tally


charithe:
Consider the following test suite definition:

```yaml
---
name: Foo
resources:
  my_resource_1:
    id: my_resource_1
    kind: my_resource
principals:
  some_user:
    id: some_user
    roles: ["test"]
  admin_user:
    id: admin_user
    roles: ["admin"]
tests:
  - name: Test1
    input:
      principals:
        - some_user
        - admin_user_1
      resources:
        - my_resource_1
      actions:
        - create
    expected:
      - principal: some_user
        resource: my_resource_1
        actions:
          create: EFFECT_DENY

  - name: Test2
    input:
      principals:
        - admin_user
      resources:
        - my_resource_1
      actions:
        - create
    expected:
      - principal: admin_user
        resource: my_resource_1
        actions:
          create: EFFECT_ALLOW
```

The output of this test looks like the following:

```
Test results
└─┬Foo (x_test.yaml) [2 OK] [1 FAILED]
  ├─┬some_user
  │ └─┬my_resource_1
  │   └──create [OK]
  └─┬admin_user
    └─┬my_resource_1
      └──create [OK]

3 tests executed [2 OK] [1 FAILED]
```

`Test1` creates an **implicit** test case for `admin_user` for `create` action with the expectation of `EFFECT_DENY` because it is listed in `principals` but has no corresponding `expected` entry. However `Test2` is an **explicit** test case for the same principal and action but with the expectation of `EFFECT_ALLOW`.

When the test suite is run, the implicit test fails because the expectation of `EFFECT_DENY` does not match the actual effect of `EFFECT_ALLOW`. Then the explicit test (`Test2`) is run and succeeds because the expectation matches the actual result. Because we group the results by principal and action, this result overwrites the previous failure entry. However, the running tally is not updated to reflect that fact and we end up with the rather confusing `3 tests executed [2 OK] [1 FAILED]` message.

There are a couple ways in which we could handle this:
1. Warn the users when they have conflicting `principal:action` tests
1. Group the test results by test name first (we may still need to detect duplicate test names)
1. Instead of keeping a running tally, do the tally at the end

My personal preference is for grouping by test name first because that would produce the least surprising/confusing output for users.

haines:
[...]

PR #1498, opened 2023-03-22
enhancement!: Group test results by test name

### LINKED ISSUE #1411: Test output mode where results are grouped by test name first
Follows from #1403

We currently don't display the name of the test case in the output produced by the test runner. It would be useful to have that information displayed to help users link the results back to a particular test case they wrote. 

One way to do this would be to group the results by test name first. However, if the JSON output also has to change to accommodate this (I haven't checked), then that would be a breaking change. In that case, let's evaluate whether we should go ahead with that breaking change or come up with an alternative.

### LINKED ISSUE #1490: Duplicate test detection false positive

### Current Behavior

I get the following message when I have a duplicate test but with different input auxData.

```
  └──Duplicate test: The combination [acme_ny_plugin|acme_ny_entry_1|create] in test "Plugin Read Scopes Test" was already exercised in test "Plugin Tenancy Test"
```

### Expected Behavior

The test is not duplicate b/c the auxData jwt I'm passing in is different so it should not trigger the duplicate detection.

### Steps To Reproduce

LMK if you need this, happy to provide but I think its pretty self explanatory.

### Environment

```markdown
- OS: N/A
- Cerbos version: 0.25.0
- Docker version: N/A
- Go version: N/A
```

### Anything else?

_No response_

oguzhand95:
resolves #1411 
fixes #1490


internal/verify/test_fixture.go
```diff
@@ -142,19 +142,21 @@ func (tf *testFixture) runTestSuite(ctx context.Context, eng Checker, shouldRun
 		}
 
 		for _, action := range test.Input.Actions {
-			testKey := fmt.Sprintf("%s|%s|%s", test.Name.PrincipalKey, test.Name.ResourceKey, action)
+			testKey := fmt.Sprintf("%s|%s|%s|%s", test.Name.TestTableName, test.Name.PrincipalKey, test.Name.ResourceKey, action)
```
charithe, line 145: We don't need to include the principal, resource and action here anymore. This check was added because of #1403. Because now the results are grouped by test name, that problem goes away. We only need to check that the test name is not duplicated.

internal/verify/verify_test.go
```diff
@@ -42,6 +42,12 @@ type TestCase struct {
 	want    *policyv1.TestResults
 }
 
+func TestReloadGolden(t *testing.T) {
+	testCases := test.LoadTestCases(t, filepath.Join("verify", "cases"))
+	eng := mkEngine(t)
+	updateGoldenFiles(t, eng, testCases)
+}
```
charithe, line 49: I don't think this is needed? Just run `go test -v -tags=tests ./internal/verify/... --args -updateGolden` to update the golden files.

api/public/cerbos/policy/v1/policy.proto
```diff
@@ -291,11 +291,16 @@ message TestResults {
   message Suite {
     string file = 1;
     string name = 2;
-    repeated Principal principals = 3;
+    repeated TestCase test_cases = 3;
```
charithe, line 294: I think we should keep the previous `principals` field at number 3 and mark it as deprecated. That way we don't break the old versions horribly (i.e. there won't be a parsing error). They would just see an empty principal list.

charithe (commented):
I think we should add a test for #1490

internal/verify/test_fixture.go
oguzhand95: Given a test file like this;
```yaml
tests:
  - name: Test 1
    input:
      principals:
        - admin
        - user
      resources:
        - contact
      actions:
        - create

    expected:
      - principal: admin
        resource: contact
        actions:
          create: EFFECT_ALLOW
      - principal: user
        resource: contact
        actions:
          create: EFFECT_ALLOW
```
We have two tests (separated into two: one with principal `admin`, another one with `user`) having the same test name (`Test 1`). Since both have the `Test 1` name after separation, I didn't remove the logic.

charithe: I see. In that case, the error message would look a bit strange because `testKey` already contains the test name in it. Maybe strip out the test name prefix from the `testKey` when producing the error message.

internal/test/testdata/verify/cases/case_010.yaml.input
```diff
@@ -0,0 +1,59 @@
...
+    expected:
+      - principal: john
+        resource: john_leave_request
+        actions:
+          view:public: EFFECT_ALLOW
+
+  - name: Test 1
+    input:
+      principals:
+        - john
```
charithe, line 50: If we change the principal here to someone else, would it still be detected as a duplicate test? We should probably disallow two test cases to have the same name regardless of whether they exercise completely different principals/resources/actions.

charithe (commented):
LGTM but I think we need to check for duplicate test case names and warn the user as well.

internal/test/testdata/verify/cases/case_010.yaml.input
oguzhand95: I modified the relevant part. Now, we;
- continue even if we have the same `principal|resource|action` pair in two test cases. (`case_08`)
- continue even if we have the same `principal|resource|action` pair with different aux data. (`case_09`)
- fail the test if there are any duplicate test names present. (`case_10`)

Does this sound good?

charithe (approved):
LGTM

PR #1499, opened 2023-03-23
feat: Kafka audit log

rcrowe:
#### Description

Adds a new audit backend that publishes access & decision log entries to Kafka.

#### Checklist 

- [ ] ~PR is linked to the corresponding issue~


internal/audit/kafka/publisher.go
```diff
@@ -0,0 +1,187 @@
...
+
+func (p *Publisher) write(ctx context.Context, msg *kgo.Record) error {
+	if !p.async {
+		return p.Client.ProduceSync(ctx, msg).FirstErr()
+	}
+
+	p.Client.Produce(ctx, msg, func(r *kgo.Record, err error) {
+		if err != nil {
+			// TODO: Handle via interceptor
+			ctxzap.Extract(ctx).Warn("failed to write audit log entry", zap.Error(err))
```
rcrowe, line 120: This felt fine for the first version, but I'd like to introduce metrics requiring this to be refactored eventually.
Hopefully, you're ok with a little duplication for now 🙏🏻

internal/audit/kafka/conf.go
```diff
@@ -0,0 +1,57 @@
...
+
+const (
+	EncodingJSON     = "json"
+	EncodingProtobuf = "protobuf"
+)
+
+// Conf is optional configuration for kafka Audit.
+type Conf struct {
+	// Seed brokers Kafka client will connect to
+	Brokers []string `yaml:"brokers" conf:",example=localhost:9092,localhost:9093"`
```
charithe, line 24:
```suggestion
	Brokers []string `yaml:"brokers" conf:",example=['localhost:9092', 'localhost:9093']"`
```

internal/audit/kafka/publisher.go
```diff
@@ -0,0 +1,187 @@
...
+func (m recordMarshaller) record(key, payload []byte) (*kgo.Record, error) {
+	return &kgo.Record{
+		Key:   key,
+		Value: payload,
+		Headers: []kgo.RecordHeader{
+			{
+				Key:   encodingHeaderKey,
+				Value: []byte(m.Encoding),
+			},
+		},
```
charithe, line 185: It might be helpful to include the kind of record (access or decision) as a header as well.

internal/audit/kafka/conf.go
```diff
@@ -0,0 +1,57 @@
...
+	// Name of the topic audit entries are written to
+	Topic string `yaml:"topic" conf:",example=cerbos.audit.log"`
+	// Data format written to Kafka, accepts either json (default) or protobuf
+	Encoding string `yaml:"format" conf:",example=protobuf"`
+	// Improve performance by not waiting for confirmation from Kafka
+	Async bool `yaml:"async" conf:",example=true"`
```
charithe, line 30: I'd rename this to `ProduceSync`. Then, if the config value is not set explicitly, we'll default to async -- which is better for performance..

charithe (commented):
Thank you. This is brilliant. For completeness sake, I think it'd be good to have the ability to configure the following as well:

- [Client ID](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#ClientID) -- default to `cerbos`
- [MaxBufferedRecords](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#MaxBufferedRecords) -- The kgo default of 10,000 is probably too high. I'd set it to something like 256.
- [RequiredAcks](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#RequiredAcks)

It'd be really useful to configure the [logger](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#Logger) with a log level taken from the environment so that we can help debug issues on customer premises. (See https://github.com/cerbos/cerbos/blob/06c514da1aa61c605d274f58e1d2034b625aa6b6/internal/storage/db/internal/db.go#L56-L62)

internal/audit/kafka/publisher.go
```diff
@@ -0,0 +1,248 @@
...
+		return nil, fmt.Errorf("failed to marshal entry: %w", err)
+	}
+
+	return &kgo.Record{
+		Key:   callID.Bytes(),
+		Value: payload,
+		Headers: []kgo.RecordHeader{
+			{
+				Key:   HeaderKeyEncoding,
+				Value: []byte(m.Encoding),
```
charithe, line 204: Since these are byte array constants, we could declare them as package vars and potentially save an allocation on each write.

internal/audit/kafka/publisher.go
```diff
@@ -0,0 +1,248 @@
...
+			},
+			{
+				Key:   HeaderKeyKind,
+				Value: []byte(KindAccess),
```
charithe, line 208: Same here. Declare as package var.

internal/audit/kafka/publisher.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+	})
+	return nil
+}
+
+type recordMarshaller struct {
+	Encoding string
+}
+
+func (m recordMarshaller) MarshalAccessLogEntry(rec *auditv1.AccessLogEntry) (*kgo.Record, error) {
```
charithe, line 178: We could reuse this function for both access and decision entries by declaring it like this:

```suggestion
func (m recordMarshaller) Marshal(kind []byte, rec proto.Message) (*kgo.Record, error) {
```

rcrowe: At first, I had it like this, but need access to `msg.GetCallId()` & wanted to reduce the work type switching on an interface.
That said, would simplify the code & we can worry about performance a bit later given you're taking on a new dependency to manage, so manageable code makes sense 👍🏻

charithe: Oh I didn't think about that. Maybe declare a new interface type for it then?

```go
type auditEntry interface {
  proto.Message
  GetCallId() []byte
}
```

rcrowe: Refactored things, see if you like the setup now.
The allocations have been removed on the headers & while there's likely room for improvements it felt best to land the PR before optimising much further, though I've included a benchmark to aid with that.

```
> benchstat old.txt new.txt
goos: darwin
goarch: arm64
pkg: github.com/cerbos/cerbos/internal/audit/kafka
                                                  │   old.txt   │               new.txt               │
                                                  │   sec/op    │   sec/op     vs base                │
RecordMarshaller_AccessLog/encoding_json-10         672.5n ± 0%   656.1n ± 1%   -2.45% (p=0.000 n=10)
RecordMarshaller_AccessLog/encoding_protobuf-10     163.4n ± 0%   140.9n ± 0%  -13.77% (p=0.000 n=10)
RecordMarshaller_DecisionLog/encoding_json-10       693.1n ± 3%   697.9n ± 0%        ~ (p=0.183 n=10)
RecordMarshaller_DecisionLog/encoding_protobuf-10   162.6n ± 1%   150.7n ± 1%   -7.29% (p=0.000 n=10)
geomean                                             333.6n        314.0n        -5.86%

                                                  │  old.txt   │              new.txt              │
                                                  │    B/op    │    B/op     vs base               │
RecordMarshaller_AccessLog/encoding_json-10         600.0 ± 0%   584.0 ± 0%  -2.67% (p=0.000 n=10)
RecordMarshaller_AccessLog/encoding_protobuf-10     304.0 ± 0%   288.0 ± 0%  -5.26% (p=0.000 n=10)
RecordMarshaller_DecisionLog/encoding_json-10       600.0 ± 0%   584.0 ± 0%  -2.67% (p=0.000 n=10)
RecordMarshaller_DecisionLog/encoding_protobuf-10   304.0 ± 0%   288.0 ± 0%  -5.26% (p=0.000 n=10)
geomean                                             427.1        410.1       -3.97%

                                                  │  old.txt   │              new.txt               │
                                                  │ allocs/op  │ allocs/op   vs base                │
RecordMarshaller_AccessLog/encoding_json-10         15.00 ± 0%   13.00 ± 0%  -13.33% (p=0.000 n=10)
RecordMarshaller_AccessLog/encoding_protobuf-10     6.000 ± 0%   4.000 ± 0%  -33.33% (p=0.000 n=10)
RecordMarshaller_DecisionLog/encoding_json-10       15.00 ± 0%   13.00 ± 0%  -13.33% (p=0.000 n=10)
RecordMarshaller_DecisionLog/encoding_protobuf-10   6.000 ± 0%   4.000 ± 0%  -33.33% (p=0.000 n=10)
geomean                                             9.487        7.211       -23.99%
```

rcrowe:
@charithe Great idea, not sure what I was thinking here. Trying to keep them independent & use the tests to catch changes, but it's clear the audit record always has access to both & this is a lot cleaner 👍🏻

charithe (approved):
LGTM. Thank you!

Issue #1504, opened 2023-03-27
Add E2E test for Kafka audit log


charithe:
Add an E2E test that exercises the Kafka audit log.

PR #1516, opened 2023-04-04
fix: Kafka async produce context cancellation

rcrowe:
#### Description

Fixes the `context canceled` audit log message when producing via the async mechanism.

I've first committed some integration tests to show `sync` working, followed by `async` failing. The second commit then shows that async test is passing.

Fixes: https://github.com/cerbos/cerbos/issues/1504

#### Checklist 

- [ ] ~PR is linked to the corresponding issue~

### DIFF
--- a/internal/audit/kafka/kafka_test.go
+++ b/internal/audit/kafka/kafka_test.go
@@ -6,17 +6,20 @@ package kafka_test
 import (
 	"context"
 	"fmt"
+	"net"
+	"strconv"
 	"testing"
 	"time"
 
+	"github.com/ory/dockertest/v3"
+	"github.com/ory/dockertest/v3/docker"
 	"github.com/stretchr/testify/require"
-	"github.com/testcontainers/testcontainers-go"
-	"github.com/testcontainers/testcontainers-go/wait"
 	"github.com/twmb/franz-go/pkg/kadm"
 	"github.com/twmb/franz-go/pkg/kgo"
 
 	auditv1 "github.com/cerbos/cerbos/api/genpb/cerbos/audit/v1"
 	"github.com/cerbos/cerbos/internal/audit"
+	_ "github.com/cerbos/cerbos/internal/audit/kafka"
 	"github.com/cerbos/cerbos/internal/config"
 )
 
@@ -28,13 +31,12 @@ const (
 )
 
 func TestSyncProduce(t *testing.T) {
+	t.Parallel()
+
 	ctx := context.Background()
 
 	// setup kafka
-	uri := newKafkaBroker(t)
-	err := newKafkaTopic(uri, defaultIntegrationTopic)
-	require.NoError(t, err)
-	// kafka audit backend with synchronous publishing
+	uri := newKafkaBroker(t, defaultIntegrationTopic)
 	log, err := newLog(map[string]any{
 		"audit": map[string]any{
 			"enabled": true,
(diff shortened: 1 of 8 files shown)


rcrowe:
Failing test to show it not working as expected, as observed in our dev environments while we've been testing Kafka audit support when running with async producer...

https://github.com/cerbos/cerbos/actions/runs/4611974520/jobs/8152336438?pr=1516#step:5:57

internal/audit/kafka/publisher.go
```diff
@@ -170,18 +168,12 @@ func (p *Publisher) WriteDecisionLogEntry(ctx context.Context, record audit.Deci
...
+		return p.Client.ProduceSync(ctx, msg).FirstErr()
 	}
 
 	// detach the context from the caller so the request can return
 	// without cancelling any async kafka operations
-	produceCtx, produceCancel := context.WithTimeout(context.Background(), p.produceTimeout)
-	defer produceCancel()
-
-	p.Client.Produce(produceCtx, msg, func(r *kgo.Record, err error) {
+	p.Client.Produce(context.Background(), msg, func(r *kgo.Record, err error) {
```
rcrowe, line 176: This is the actual fix, completely detaching the context cancellation from the parent.

The other changes in the PR cover the integration tests & dropping the produce timeout config, which is no longer required.

charithe (commented):
Thank you for investigating and fixing the issue 🙌🏽 

Do you mind trying to swap out testcontainers with dockertest? It's what we use everywhere else so that means we don't have to introduce a new dependency. Here's one of the places where we use dockertest:

https://github.com/cerbos/cerbos/blob/80072f5ffbf4c4711f27c1c5df27254dd9cfe655/internal/storage/db/mysql/mysql_test.go#L35-L45

internal/audit/kafka/kafka_test.go
```diff
@@ -0,0 +1,201 @@
...
+		records, err := fetchKafkaTopic(uri, defaultIntegrationTopic)
+		require.NoError(t, err)
+		return len(records) == 2
+	}, 10*time.Second, 100*time.Millisecond, "expected to see audit log entries in kafka")
+}
+
+func newKafkaBroker(t *testing.T, topic string) string {
+	t.Helper()
+
+	hostPort, err := freePort()
```
charithe, line 123: You can use [`util.GetFreePort()`](https://github.com/cerbos/cerbos/blob/583155638574716a139bb272950d96f9f0b26179/internal/util/network.go#L52)

However, in this case, I don't think you even need to figure the free port yourself. You can get the random port assigned to the container by using `resource.GetPort("9092/tcp")`. You don't have to do the host port mapping on the container using `PortBindings` either. :)

charithe (commented):
LGTM 👍🏽 

Have a tiny nit about determining the port used by Kafka. Otherwise it's good to go.

internal/audit/kafka/kafka_test.go
rcrowe: I tried that, but I struggled to get it working with the advertised port & Kafka admin client

charithe: Ah OK. I reckon it's the Kafka client trying to be clever and discover the brokers. That's fine then. Do you mind switching to `util.GetFreePort`?

rcrowe: Yeah, that's what it looked to be. On connect, the metadata back from the broker was using the local address which was not the same outside the container. 

Will migrate to that util function 👍🏻

charithe (approved):
Thanks!

PR #1520, opened 2023-04-06
enhancement: Add principal, resource and action as properties of JUnit output

### LINKED ISSUE #1519: Add principal, resource and action as properties of JUnit output
Add properties for the principal, resource and action for each test case in the JUnit output so that they are easier to parse.

oguzhand95:
resolves #1519


charithe (approved):
👍🏽

PR #1526, opened 2023-04-12
feat: Validate policy and test files with JSON schema

oguzhand95:


cmd/cerbos/compile/compile.go
```diff
@@ -85,6 +88,15 @@ func (c *Cmd) Run(k *kong.Kong) error {
 		return err
 	}
 
+	policySchema, err := jsonschema.CompileString("Policy.schema.json", string(schema.Policy))
```
charithe, line 91: Can't we integrate this step into the `ValidatePolicies` function?

internal/jsonschema/jsonschema.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		if err != nil {
+			return err
+		}
+
+		if d.IsDir() && d.Name() == util.TestDataDirectory {
+			if d.Name() == util.TestDataDirectory {
+				return fs.SkipDir
+			}
```
charithe, line 80: Isn't this redundant?

charithe (commented):
The validation step needs to be a part of the index build process. As the index builder goes through each file, it should validate them using the appropriate schema and collect any validation errors to be returned back as index build errors.

internal/jsonschema/jsonschema.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to open file %s: %w", path, err)
+	}
+
+	data, err := io.ReadAll(f)
+	if err != nil {
+		return fmt.Errorf("failed to read file %s: %w", path, err)
+	}
+
+	var y interface{}
```
charithe, line 55:
```suggestion
	var y any
```

internal/test/testdata/index/incomplete_files.yaml
```diff
@@ -1,9 +1,9 @@
 ---
 wantErrList:
   loadFailures:
-    - error: "invalid Policy.DerivedRoles: embedded message failed validation | caused by: invalid DerivedRoles.Definitions: value must contain at least 1 item(s)"
+    - error: "jsonschema: '/derivedRoles' does not validate with https://api.cerbos.dev/cerbos/policy/v1/Policy.schema.json#/allOf/0/properties/derivedRoles/$ref/required: missing properties: 'definitions'"
```
charithe, line 4: We should clean up this error message like we do in https://github.com/cerbos/cerbos/blob/b6fdc81fb3ab0245ff920107ea9bc4b8afe20c5f/internal/schema/error.go#L43. The user doesn't care which schema was used because they can't control it. Just need to tell them the actual problem.

schema/schema.go
```diff
@@ -21,6 +21,12 @@ var svcSwaggerRaw []byte
 //go:embed assets/ui.html
 var rapidocHTML []byte
 
+//go:embed jsonschema/cerbos/policy/v1/Policy.schema.json
+var PolicyJSONSchema []byte
```
charithe, line 25: Because `jsonschema` only reads strings, there's no need to read this file as a `[]byte` and then cast it back to a string.

```suggestion
var PolicyJSONSchema string
```

schema/schema.go
```diff
@@ -21,6 +21,12 @@ var svcSwaggerRaw []byte
...
+
+//go:embed jsonschema/cerbos/policy/v1/TestSuite.schema.json
+var TestSuiteJSONSchema []byte
```
charithe, line 28:
```suggestion
var TestSuiteJSONSchema string
```

go.mod
```diff
@@ -95,6 +95,7 @@ require (
 	google.golang.org/grpc v1.54.0
 	google.golang.org/protobuf v1.30.0
 	gopkg.in/yaml.v2 v2.4.0
+	gopkg.in/yaml.v3 v3.0.1
```
charithe, line 98: I don't think we need both `yaml.v2` and `yaml.v3`. The `yaml.v2` package is imported by `internal/audit/conf.go`. Let's change it to import `yaml.v3`.

internal/jsonschema/jsonschema.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+}
+
+// ValidatePolicy validates the policy in the fsys with the JSON schema.
+func ValidatePolicy(fsys fs.FS, path string) error {
+	return validate(policySchema, fsys, path)
+}
+
+// ValidateTest validates the test in the fsys with the JSON schema.
+func ValidateTest(fsys fs.FS, path string) error {
```
charithe, line 40: This isn't used anywhere. I assume you still haven't gotten around to it?

oguzhand95: It's actually used in `internal/verify/verify.go:101` ([link](https://github.com/cerbos/cerbos/pull/1526/files#diff-7285c90794dc46994c4a2590334d2c9da31710e20af9a125dd479e997ae0e5bdR101))

charithe: My bad. Sorry.

internal/jsonschema/jsonschema.go
```diff
@@ -0,0 +1,136 @@
...
+
+type validationErrorList []validationError
+
+func (e validationErrorList) ErrOrNil() error {
+	if len(e) == 0 {
+		return nil
+	}
+
+	return e
+}
```
charithe, line 119: Unused

```suggestion
```

internal/jsonschema/jsonschema.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (e validationErrorList) Error() string {
+	return fmt.Sprintf("[%s]", strings.Join(e.ErrorMessages(), ", "))
```
charithe, line 122:
```suggestion
	return fmt.Sprintf("file is not valid: [%s]", strings.Join(e.ErrorMessages(), ", "))
```

charithe (approved):
LGTM

PR #1541, opened 2023-04-19
fix: Introduce custom temp directory creation function with pathname length limit for uds

### LINKED ISSUE #1540: Tests within `internal/server` fail on creation of Unix domain sockets with address `t.TempDir()`
On my system (`macOS Monterey 12.4`), running tests (`make test`) causes the following:

```
panic: failed to create listener at 'unix:/var/folders/yt/50sw1ks16gg2xgr_hzyr5d6c0000gn/T/TestServerstore=diskapiwith_tlsuds3243008092/001/grpc.sock': listen unix /var/folders/yt/50sw1ks16gg2xgr_hzyr5d6c0000gn/T/TestServerstore=diskapiwith_tlsuds3243008092/001/grpc.sock: bind: invalid argument
```

This is likely related to the length of the address string (the above is potentially above the max allowed limit). Manually crafting a temp dir using a shorter path fixes the failing tests.

Sambigeara:
Fixes #1540 

Unix domain socket addresses are limited in length (generally) to 104-108, platform specific (on `MacOS Monterey 12.4`, this appears to be 104). Sometimes, the generation of the temp directory pathname (via `t.TempDir()`) can exceed this limit, so when attempting to create a uds listener, it would fail with `invalid argument`.

This PR creates a helper function with more explicit control over the length of the output pathname.

### DIFF
--- a/internal/server/server_test.go
+++ b/internal/server/server_test.go
@@ -177,7 +177,7 @@ func apiTests(tpg testParamGen) func(*testing.T) {
 }
 
 // createTempDirForUDS is used to generate a temporary directory with a pathname length below a defined limit.
-// This is to prevent the randonly generated directory path length exceeding platform limits (~100-108 ish, platform specific).
+// This is to prevent the randonly generated directory path length exceeding platform limits (104-108 ish, platform specific).
 func createTempDirForUDS(t *testing.T) string {
 	t.Helper()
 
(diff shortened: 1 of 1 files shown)


internal/server/server_test.go
```diff
@@ -172,6 +177,52 @@ func apiTests(tpg testParamGen) func(*testing.T) {
 	}
 }
 
+// createTempDirForUDS is used to generate a temporary directory with a pathname length below a defined limit.
+// This is to prevent the randonly generated directory path length exceeding platform limits (104-108, platform specific).
```
charithe, line 181:
```suggestion
// This is to prevent the randomly generated directory path length exceeding platform limits (104-108, platform specific).
```

charithe (approved):

PR #1547, opened 2023-04-26
enhancement(helm): Volume mounts for Cerbos Cloud

charithe:
If the user is using the default settings, automatically mount
`emptyDir` volumes for temp and cache directories.

Also adds an example values file for connecting to Cerbos Cloud.


oguzhand95 (approved):

deploy/charts/cerbos/templates/deployment.yaml
```diff
@@ -95,8 +103,15 @@ spec:
         - name: config
           configMap:
             name: {{ include "cerbos.fullname" . }}
+      {{- if $defaultBundleDriverEnabled }}
+        - name: bundletmp
+          emptyDir: {}
+        - name: bundlecache
+          emptyDir: {}
```
haines, line 110: What if the user wants to mount a persistent volume for cache?

haines: Ah got it, they have to specify the `tempDir` as well as a volume mount.

PR #1548, opened 2023-04-26
enhancement(helm): Upgrade to HPA v2 API

charithe:
The beta versions of `HorizontalPodAutoScaler` v2 will be dropped in
Kubernetes 1.26 and some distributions such as GKE are warning about it
now.


haines (dismissed):

haines:
Should we add a `kubeVersion` to our chart? I guess we require at least 1.23

haines (approved):

PR #1550, opened 2023-04-27
enhancement: Handle schemas folder being in wrong place with more verbose error

### LINKED ISSUE #1539: Provide more verbose error message if `_schemas` directory is not at storage root
As per #1536

oguzhand95:
resolves #1539

### DIFF
--- a/internal/storage/index/builder.go
+++ b/internal/storage/index/builder.go
@@ -23,6 +23,9 @@ import (
 	"github.com/cerbos/cerbos/internal/util"
 )
 
+// errSchemasInWrongDir signals that schemas folder is in the wrong place.
+var errSchemasInWrongDir = fmt.Errorf("%s directory must be under the root of the storage directory", util.SchemasDirectory)
+
 const maxLoggableBuildErrors = 5
 
 // BuildError is an error type that contains details about the failures encountered during the index build.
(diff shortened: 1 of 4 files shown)


internal/storage/index/builder.go
```diff
@@ -94,6 +98,9 @@ func build(ctx context.Context, fsys fs.FS, opts buildOptions) (Index, error) {
 				d.Name() == util.TestDataDirectory ||
 				util.IsHidden(d.Name()) {
 				return fs.SkipDir
+			} else if strings.HasSuffix(filePath, schema.Directory) {
```
charithe, line 101:
```suggestion
			} else if d.Name() == schema.Directory {
```

internal/storage/index/builder.go
```diff
@@ -23,6 +24,9 @@ import (
 	"github.com/cerbos/cerbos/internal/util"
 )
 
+// ErrSchemasInWrongDir signals that schemas folder is in the wrong place.
+var ErrSchemasInWrongDir = fmt.Errorf("%s directory must be under the root of the storage directory", util.SchemasDirectory)
```
charithe, line 28: This doesn't need to be a public error because we don't check for it anywhere.

internal/storage/index/builder.go
haines, line 101: Why `HasSuffix`? Wouldn't that potentially have false positives like `foo/bar_schemas`? What about 
```suggestion
			} else if d.Name() == schema.Directory {
```

charithe (approved):

PR #1560, opened 2023-05-02
feat: Storage overlay

### LINKED ISSUE #1106: Stacked storage driver
This is a special storage driver that stacks two or more existing storage drivers together. Stores are accessed in the priority order they are defined. First hit wins.  

```yaml
storage:
  driver: stack
  stack:
    drivers:
      - disk
      - blob
      - sqlite3
      
  disk:
    directory: /etc/cerbos/policies

  blob:
    bucket: "gs://my-bucket-name"
    workDir: ${HOME}/tmp/cerbos/work
    updatePollInterval: 10s

  sqlite3:
    dsn: "file:/tmp/cerbos.sqlite?mode=rwc&cache=shared&_fk=true"
```

Sambigeara:
Closes #1106 

Introduces the ability to define a second `fallback` storage driver via a configurable [circuit breaker pattern](https://learn.microsoft.com/en-us/previous-versions/msp-n-p/dn589784(v=pandp.10)?redirectedfrom=MSDN):

```yaml
storage:
  driver: "overlay"
  overlay:
    baseDriver: postgres
    fallbackDriver: disk
    fallbackErrorThreshold: 1 # number of errors allowed within the fallbackErrorWindow before failover occurs
    fallbackErrorWindow: 1s # the rolling window in which errors are aggregated
  disk:
    directory: policies
    watchForChanges: true
  postgres:
    url: "postgres://${PG_USER}:${PG_PASSWORD}@localhost:5432/postgres?sslmode=disable&search_path=cerbos"
```

I wasn't sure if this was a `feat` or an `enhancement` - can change.


Sambigeara:
Realised whilst raising PR that I need to add documentation - I'll get this done tomorrow morning.

internal/storage/overlay/conf.go
```diff
@@ -0,0 +1,50 @@
...
+	defaultFailoverInterval  = 5 * time.Minute
+)
+
+// Conf is required (if driver is set to 'overlay') configuration for overlay storage driver.
+// +desc=This section is required only if storage.driver is overlay.
+type Conf struct {
+	// BaseDriver is the default storage driver
+	BaseDriver string `yaml:"baseDriver" conf:"required,example=disk"`
+	// FallbackDriver is the secondary or fallback storage driver
+	FallbackDriver string `yaml:"fallbackDriver" conf:"required,example=git"`
```
charithe, line 25: For the example, I think it might make more sense if the base was `blob` and the failover was `disk`.

```suggestion
	BaseDriver string `yaml:"baseDriver" conf:"required,example=blob"`
	// FallbackDriver is the secondary or fallback storage driver
	FallbackDriver string `yaml:"fallbackDriver" conf:"required,example=disk"`
```

internal/storage/overlay/conf.go
```diff
@@ -0,0 +1,50 @@
...
+	// FailoverThreshold is the max number of errors we allow within the failoverInterval period
+	FailoverThreshold int `yaml:"failoverThreshold,omitempty" conf:",example=5"`
```
charithe, line 27: Shall we call this `FallbackErrorThreshold` so that it ties into `FallbackDriver`

internal/storage/overlay/conf.go
```diff
@@ -0,0 +1,50 @@
...
+	// FailoverInterval is the cyclic period within which we aggregate failures
+	FailoverInterval time.Duration `yaml:"failoverInterval" conf:",example=5m"`
```
charithe, line 29: Maybe `RollingWindow` or `FallbackErrorWindow` because it's really a rolling window rather than an interval?

internal/storage/overlay/store.go
```diff
@@ -0,0 +1,204 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("unknown storage driver [%s]", key)
+		}
+
+		store, err := cons(ctx, confW)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create overlay store: %w", err)
+		}
+
+		sourceStore, ok := store.(storage.SourceStore)
```
charithe, line 53: Why is it necessary for the store to be a `SourceStore`? Isn't it enough to implement the `Store` interface? 

In my mind, I was thinking that the `overlay` driver could implement all the relevant interfaces (`SourceStore`, `BinaryStore`, and `Reloadable`). Then, based on which driver was active, each of those methods could either call the corresponding method on the underlying driver or return an error.

internal/storage/store.go
```diff
@@ -68,6 +68,19 @@ func RegisterDriver(name string, cons Constructor) {
 	drivers[name] = cons
 }
 
+// GetDriverConstructor registers a storage driver.
+func GetDriverConstructor(name string) (Constructor, error) {
+	driversMu.Lock()
+	defer driversMu.Unlock()
```
charithe, line 74:
```suggestion
	driversMu.RLock()
	defer driversMu.RUnlock()
```

internal/storage/overlay/store.go
charithe: I have been thinking about this a bit more and, at the cost of complicating things a little bit, what if the `overlay` driver always exposed the same interfaces as the `base` driver?  When the fallback driver is active -- which should not happen quite so often anyway -- it's OK to return errors for methods that couldn't be invoked or emulated using it.

Sambigeara: My thinking was that we weren't likely to be configuring overlays with `BinaryStore`s (assuming the only implementations of them were bundle/cloud related). Asserting on `SourceStore` shows intent and simplifies the interface methods by negating the need for type assertions/switches in each. Is this likely to be used with `BinaryStore` drivers?

Omitting `Reloadable` was an arbitrary decision (similar to the decision to avoid any mutable operations) but I realise now that this is an oversight given all(most?) stores seem to support it!

> what if the overlay driver always exposed the same interfaces as the base driver?

This would make sense to me in the scenario where driver interfaces frequently varied. I think I'm struggling to see the pay-off now given I'm stuck in a "`SourceStore` only" mindset 🤔.

charithe: I think `BinaryStore` is a prime candidate for having a fallback driver configured in case of connectivity issues to the cloud. I expect that some users would want to configure `bundle` as the primary store but have `git` or `disk` as fallback drivers for redundancy.

Sambigeara: In that case, enforcing the base driver interface makes sense 👍

charithe (commented):
I just realised that we can simplify this quite a bit. Sorry I didn't think about it earlier.

Essentially, the overlay driver only needs to implement the `BinaryStore` interface. If one or both of the wrapped drivers is a `SourceStore`, we can easily convert them to `BinaryStore` anyway. Therefore, you only need to have two struct fields of type `BinaryStore` instead of the four that it has now. 

The `SourceStore` interface is normally used by the `CompileManager` when the engine is initialised but because we provide our special `PolicyLoader`, those methods will never be used. Similarly, because we don't support mutable operations, the admin service can work with just the `Store` interface which is already embedded in `BinaryStore`.

Sambigeara:
Aaah, I was getting a bit muddled about how updates to the underlying `SourceStore`s would propagate to the `PolicyLoader`s (e.g. polling for updates to `git`/`blob` storage etc), but on further reading, it seems that this'll all be handled per store in the background via subscriptions, and therefore `GetPolicySet` will always reflect up to date store state? If so, that all makes sense!

On that, would it not still make sense to implement the `Reloadable` interface? We don't support mutable ops, but if we already support updates via polling, it doesn't seem unreasonable to expose this manual trigger?

> Therefore, you only need to have two struct fields of type BinaryStore instead of the four that it has now.

I think we'll still need to maintain the four struct fields (two `Stores` and two `PolicyLoader`s), because there's a chicken-egg situation whereby the [`schemaMgr`](https://github.com/Sambigeara/cerbos/blob/767bec6f5a62487e267af7480b7257beca25f11a/internal/server/server.go#L139) expects a `Store` implementation, but then is required by the `compilation.Manager` to [build the `PolicyLoader`](https://github.com/Sambigeara/cerbos/blob/767bec6f5a62487e267af7480b7257beca25f11a/internal/server/server.go#L148). I'm not sure how to solve that tidily without separately instantiating the stores and the policy loaders in two stages.

Sambigeara:
I've addressed most of the smaller review items but I've left the `Reloadable` interface in there for now until the need is clarified.

internal/storage/overlay/store.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil, fmt.Errorf("failed to create fallback policy loader: %w", err)
+	}
+
+	return &Store{
+		log:                 logger,
+		conf:                conf,
+		baseStore:           baseStore,
+		fallbackStore:       fallbackStore,
+		circuitBreaker:      newCircuitBreaker(conf),
+		SubscriptionManager: storage.NewSubscriptionManager(ctx),
```
charithe, line 76: I don't think we need this?

internal/storage/overlay/store.go
```diff
@@ -0,0 +1,213 @@
...
+
+func (s *Store) LoadSchema(ctx context.Context, url string) (io.ReadCloser, error) {
+	return withCircuitBreaker(
+		s,
+		func() (io.ReadCloser, error) { return s.baseStore.LoadSchema(ctx, url) },
+		func() (io.ReadCloser, error) { return s.fallbackStore.LoadSchema(ctx, url) },
+	)
+}
+
+func (s *Store) Reload(ctx context.Context) error {
```
charithe, line 192: Shouldn't we just try to reload both stores in parallel? If any of them is not a `Reloadable` then it would just be a noop. I don't think we should use the circuit breaker here because that would result in only one store being reloaded.

I'd use a [`conc.Pool`](https://pkg.go.dev/github.com/sourcegraph/conc@v0.3.0/pool#Pool) with   `WithContext().WithErrors().WithFirstError()`.

charithe (commented):
I think it's reasonable to implement `Reloadable` because it allows users to reload the store through the Admin API. I left a comment down below about it. 

You're right. The `Store` struct needs four fields. I wasn't thinking about the initialisation stage for `SourceStores`.

charithe (approved):
LGTM 👍🏽

PR #1579, opened 2023-05-12
chore(test): Overlay e2e test

Sambigeara:
Closes #1571 

* Added an e2e test for the overlay storage driver
* Added new method to pre-compute arbitrary env vars for e2e tests 
* Also made overlay store instantiation concurrent for a slight optimisation

PR #1583, opened 2023-05-15
fix: Add `helmfile init` step to e2e GH actions setup

Sambigeara:
#1579 introduced a change which called `helmfile apply` instead of `helmfile sync`. The former requires `helm diff`, which was previously not installed on setup.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -37,6 +37,10 @@ jobs:
             trafficManagerAPI: 30s
           EOF
 
+      - name: Initialize Helmfile
+        run: |
+          helmfile init --force
+
       - name: Install KinD
         uses: helm/kind-action@v1.5.0
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yaml
```diff
@@ -19,12 +19,15 @@ jobs:
         uses: ./.github/actions/setup-go
 
       - name: Install Helm
-        uses: azure/setup-helm@v3.5
+        uses: azure/setup-helm@v3
+        with:
+          version: 'v3.11.1'
+        id: install
```
charithe, line 25: Apparently `helmfile init` installs `helm` too? If that's the case, we can remove this step.

charithe (commented):
Yeah, testing GH actions is a pain. For complicated workflows that are hard to setup `act` for, I go with pushing the branch to the Cerbos repo as well.

.github/workflows/e2e.yaml
Sambigeara: Well that's cool

PR #1584, opened 2023-05-15
enhancement: Verify required tables are present in the db

oguzhand95:


internal/storage/db/internal/db.go
```diff
@@ -720,3 +721,28 @@ func (s *dbStorage) Reload(context.Context) error {
 	s.NotifySubscribers(storage.NewReloadEvent())
 	return nil
 }
+
+// Verify verifies the tables required by cerbos are available.
+func (s *dbStorage) Verify(ctx context.Context) error {
```
charithe, line 726: `Verify` is a bit too vague. Let's call it `CheckSchema` instead.

internal/storage/db/internal/db.go
```diff
@@ -720,3 +721,28 @@ func (s *dbStorage) Reload(context.Context) error {
...
+			).
+			Executor().
+			ExecContext(ctx)
+		if err != nil {
+			failed = append(failed, table)
+		}
+	}
+
+	if len(failed) > 0 {
+		return fmt.Errorf("failed to the required database tables: %s", strings.Join(failed, ", "))
```
charithe, line 744:
```suggestion
		return fmt.Errorf("schema check failed: %s", strings.Join(failed, ", "))
```

internal/storage/db/internal/model.go
```diff
@@ -33,11 +33,15 @@ const (
...
 	PolicyAncestorTblAncestorIDCol = "ancestor_id"
 
+	PolicyRevisionTbl = "policy_revision"
+
 	SchemaTbl              = "attr_schema_defs"
 	SchemaTblIDCol         = "id"
 	SchemaTblDefinitionCol = "definition"
 )
 
+var tables = []string{PolicyTbl, PolicyDepTbl, PolicyAncestorTbl, PolicyRevisionTbl, SchemaTbl}
```
charithe, line 43:
```suggestion
var requiredTables = []string{PolicyTbl, PolicyDepTbl, PolicyAncestorTbl, PolicyRevisionTbl, SchemaTbl}
```

internal/storage/db/mysql/conf.go
```diff
@@ -18,6 +18,8 @@ type Conf struct {
 	ServerPubKey map[string]string      `yaml:"serverPubKey" conf:",example=\n  mykey: testdata/server_public_key.pem"`
 	// DSN is the data source connection string.
 	DSN string `yaml:"dsn" conf:"required,example=\"user:password@tcp(localhost:3306)/db?interpolateParams=true\""`
+	// Verify enables the pre-check for the required database tables.
+	Verify bool `yaml:"verify" conf:",example=false"`
```
charithe, line 22: I think we should default this to `true` because the main reason for introducing this feature is that some people don't RTFM. 

```suggestion
	SkipSchemaCheck bool `yaml:"skipSchemaCheck" conf:",example=true"`
```

internal/storage/db/mysql/mysql.go
```diff
@@ -25,7 +25,10 @@ import (
 	"github.com/cerbos/cerbos/internal/storage/db/internal"
 )
 
-const DriverName = "mysql"
+const (
+	DriverName      = "mysql"
+	urlToSchemaDocs = "https://docs.cerbos.dev/cerbos/latest/configuration/storage.html#_database_object_definitions_2"
```
charithe, line 30: Let's add an anchor ID to the documentation to make this link nicer rather than saying "...object_definitions_2"

internal/storage/db/mysql/mysql.go
```diff
@@ -59,12 +62,18 @@ func NewStore(ctx context.Context, conf *Conf) (*Store, error) {
...
-	storage, err := internal.NewDBStorage(ctx, goqu.New("mysql", db))
+	s, err := internal.NewDBStorage(ctx, goqu.New("mysql", db))
 	if err != nil {
 		return nil, err
 	}
 
-	return &Store{DBStorage: storage}, nil
+	if conf.Verify {
+		if err := s.Verify(ctx); err != nil {
+			return nil, fmt.Errorf("failed to verify mysql database schema (%s): %w", urlToSchemaDocs, err)
```
charithe, line 72:
```suggestion
			return nil, fmt.Errorf("schema check failed. Ensure that the schema is correctly defined as documented at %s: %w", urlToSchemaDocs, err)
```

internal/storage/db/sqlite3/sqlite3.go
```diff
@@ -69,12 +72,18 @@ func NewStore(ctx context.Context, conf *Conf) (*Store, error) {
...
 	}
 
-	storage, err := internal.NewDBStorage(ctx, goqu.New("sqlite3", db))
+	s, err := internal.NewDBStorage(ctx, goqu.New("sqlite3", db))
 	if err != nil {
 		return nil, err
 	}
 
-	return &Store{DBStorage: storage}, nil
+	if conf.Verify {
```
charithe, line 80: The SQLite schema is automatically created by Cerbos so there's no need to verify it here.

charithe (commented):
LGTM. Just a minor copy change and a couple of log lines to add.

internal/storage/db/internal/db.go
```diff
@@ -66,13 +68,15 @@ func NewDBStorage(ctx context.Context, db *goqu.Database, dbOpts ...DBOpt) (DBSt
...
 		SubscriptionManager: storage.NewSubscriptionManager(ctx),
 	}, nil
 }
 
 type dbStorage struct {
-	opts *dbOpt
-	db   *goqu.Database
+	opts   *dbOpt
+	db     *goqu.Database
+	logger *zap.Logger
```
charithe, line 79: This doesn't need to be a struct member because it's only used in the `CheckSchema` method. Just create the logger in the method itself.

internal/storage/db/internal/db.go
```diff
@@ -737,12 +743,15 @@ func (s *dbStorage) CheckSchema(ctx context.Context) error {
 			ExecContext(ctx)
 		if err != nil {
 			failed = append(failed, table)
+			s.logger.Error("Check failed for the table", zap.String(tableLogKey, table))
```
charithe, line 746:
```suggestion
			s.logger.Error("Check failed for the table", zap.String(tableLogKey, table), zap.Error(err))
```

charithe (approved):

PR #1592, opened 2023-05-18
fix: Return correct error code when store is invalid

### LINKED ISSUE #1590: Internal error when evaluating resource with a disabled scope hierarchy

### Current Behavior

Internal Server Error returned when evaluating a resource policy with a scope defined where one resource policy in the scope hierarchy is disabled. Example response from `CheckResources`:
```
{
  "code": 13,
  "message": "Policy check failed"
}
```

### Expected Behavior

If this setup is invalid, then I expect to see a 400-like error with an error message indicating that the resource access could not be evaluated due to disabled scope policy.

### Steps To Reproduce

1. Add scoped policies such that one policy is disabled (I observed that the level of the hierarchy does not matter). Example with default policy being disabled:
```
{
  "policies": [
    {
      "apiVersion": "api.cerbos.dev/v1",
      "disabled": true,
      "resourcePolicy": {
        "resource": "test:resource",
        "scope": "",
        "version": "default",
        "rules": [
          {
            "actions": [
              "read"
            ],
            "roles": [
              "TestRole"
            ],
            "effect": "EFFECT_DENY"
          }
        ]
      }
    },
    {
      "apiVersion": "api.cerbos.dev/v1",
      "resourcePolicy": {
        "resource": "test:resource",
        "scope": "a",
        "version": "default",
        "rules": [
          {
            "actions": [
              "read"
            ],
            "roles": [
              "TestRole"
            ],
            "effect": "EFFECT_ALLOW"
          }
        ]
      }
    }
  ]
}
```
2. CheckResources call to see internal error:
```
{
  "requestId": "c2db17b8-4f9f-4fb1-acfd-9162a02be42b",
  "includeMeta": true,
  "principal": {
    "id": "bugs_bunny",
    "policyVersion": "default",
    "roles": [
      "TestRole"
    ]
  },
  "resources": [
    {
      "actions": [
        "read"
      ],
      "resource": {
        "kind": "test:resource",
        "policyVersion": "default",
        "scope": "a",
        "id": "XX125"
      }
    }
  ]
}
```

### Environment

```markdown
- OS:
- Cerbos version: github.com/cerbos/cerbos v0.26.0
- Docker version: Docker version 20.10.11, build dea9396
- Go version: go version go1.19 darwin/amd64
```

### Anything else?

_No response_

charithe:
Instead of returning `Internal` error, return `FailedPrecondition` when
the requested policy has a compilation error. We already had this special case handled in the `CheckResourceSet` RPC but, for some reason, it had been omitted in the other RPCs.

Fixes #1590

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -182,6 +182,115 @@ func (engine *Engine) submitWork(ctx context.Context, work workIn) error {
 	}
 }
 
+func (engine *Engine) PlanResources(ctx context.Context, input *enginev1.PlanResourcesInput) (*enginev1.PlanResourcesOutput, error) {
+	output, err := measurePlanLatency(func() (output *enginev1.PlanResourcesOutput, err error) {
+		ctx, span := tracing.StartSpan(ctx, "engine.Plan")
+		defer span.End()
+
+		output, err = engine.doPlanResources(ctx, input)
+		if err != nil {
+			tracing.MarkFailed(span, http.StatusBadRequest, err)
+		}
+
+		return output, err
+	})
+
+	return engine.logPlanDecision(ctx, input, output, err)
+}
+
+func (engine *Engine) doPlanResources(ctx context.Context, input *enginev1.PlanResourcesInput) (*enginev1.PlanResourcesOutput, error) {
+	// exit early if the context is cancelled
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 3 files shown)


internal/engine/engine.go
charithe, line 1: Just moved the functions around on this file to make it easier to read it from top to bottom without jumping about.

haines (approved):

PR #1611, opened 2023-06-01
fix: Update e2e overlay test seed file with newly added output data

Sambigeara:
The standard e2e db storage driver tests rely on the mutable tests to run prior to `check` tests in order to populate the DB with expected data. The overlay driver doesn't support mutable operations, hence the need for the seed data (as the test uses a mutable `baseDriver`). I forgot to update this data when introducing the recent `output` changes. This fixes that by adding the additional data.


charithe (approved):
Would it be possible to have some kind of a script or a "Golden" output generator that we can run to generate this file when the policies change?

Sambigeara:
That's a good idea. I've thus far been running the `postgres` test and taking a data-only dump from the resultant db (then doing a bit of manual tidying). I'll raise a separate ticket to try to automate this somehow.

PR #1594, opened 2023-05-19
feat: User-defined output from policy evaluation

Sambigeara:
Added support to define custom `outputs` per rule for resource and principal policies:

Resource policy:
```yaml
# Resource policy
  rules:
    - actions: ["view:public"]
      roles: ["employee"]
      effect: EFFECT_ALLOW
      name: public-view
      output:
        expr: |-
          "view:public_allowed:%s".format([request.principal.id])

# Principal policy:
  rules:
    - resource: equipment_request
      actions:
        - action: "create"
          effect: EFFECT_ALLOW
          output:
            expr: |-
              'some_arbitrary_expression'
```

The outputs are listed within each `CheckResourcesResponse_ResultEntry`:

```yaml
[
  {
    ...
    "outputs": [
      { "src": "resource.expense.v1/acme#rule-001", "val": "view_allowed:alice" },
      { "src": "resource.expense.v1/acme#rule-002", "val": "foo" }
    ]
  },
  ...
] 
```

PR #1631, opened 2023-06-07
fix: Only produce outputs for activated rules

### LINKED ISSUE #1630: Outputs get evaluated for all rules -- not just the activated ones

### Current Behavior

The `outputs` field in the API response contains outputs from all the rules in the policy.

### Expected Behavior

The output expression of a policy should only be included in the response iff the rule is activated.  

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

_No response_

charithe:
Rule output expressions should only be evaluated and appended to the API
response if the rule is activated during the processing of the request.

Fixes #1630


Sambigeara (commented):
This makes sense 👍 . I opted for global evaluation because it gave the potential for more explicit visibility into rule activation, e.g.

```yaml
- action: "reject"
  effect: EFFECT_ALLOW
  name: reject-rule
  condition:
    match:
      exr: P.attr.foo == "bar"
  output:
    expr: |-
      "eval_result:%s".format([P.attr.foo == "bar"])

```

Due in part to `includeMeta` giving `policy:scope` granularity, which (IIUC) in the case of duplicate action definitions, might still leave some guess-work WRT which rule was triggered. But I didn't really consider the impact on performance or the general misdirection and noise in the response.

charithe:
The main use case is to detect when certain conditions are satisfied. Therefore, we do have to make sure that the output is produced only for rules that are activated. Unfortunately, when I was testing it during #1594 I just happened to use policies that didn't have `output` defined for rules that didn't activate 😞

Sambigeara (approved):

PR #1633, opened 2023-06-08
chore(test): Remove manual setup from overlay E2E test

### LINKED ISSUE #1612: Automate e2e overlay test seed data generation
Context in: https://github.com/cerbos/cerbos/pull/1611.

At the moment, we need to manually regenerate the e2e [overlay test seed script](https://github.com/Cerbos/cerbos/blob/main/e2e/overlay/data.sql) if the expected db policy set changes. The process looks roughly like:

1. Run the `postgres` e2e test, persisting the environment after: `E2E_NO_CLEANUP=true ./run.sh ./postgres/...`
2. Take a data-only dump: `pg_dump -h localhost -U postgres -f foo.sql -a --inserts`
3. Bit of manual cleaning and updating the script accordingly

It would be good to automate this process one way or another.

charithe:
Avoid having to manually generate the database script to bootstrap the
overlay test.

This is a bit of a hack. It starts a second Cerbos instance connected to
the same Postgres instance and runs the bootstrapping tests to load the
policies into the database. Then, by reusing the same `*testing.T`, it
prevents the resources from being torn down and runs the second set of
tests on the same database.

Fixes #1612


Sambigeara (approved):
👌

PR #1634, opened 2023-06-09
feat: Ability to verify outputs in policy tests

### LINKED ISSUE #1629: Ability to verify policy outputs in tests
Follows #1594

In policy tests, it should be possible to verify the outputs generated by the tested policies.

charithe:
Add ability to write tests that verify output values produced from
policy evaluation.

```yaml
---
name: TestSuite
description: Tests for equipment requests
tests:
  - name: John's request
    input:
      principals:
        - john
        - bev
        - terry_tibbs
      resources:
        - john_equipment_request
        - john_pending_equipment_request
      actions:
        - view:public
        - approve
        - create
    expected:
      - principal: john
        resource: john_equipment_request
        actions:
          view:public: EFFECT_ALLOW
          approve: EFFECT_DENY
          create: EFFECT_ALLOW
        outputs:
          - action: view:public
            expected:
              - src: "resource.equipment_request.vdefault#public-view"
                val:
                  id: john
                  keys: XX125
                  formatted_string: "id:john"
                  some_bool: true
                  some_list: ["foo", "bar"]
                  something_nested:
                    nested_str: foo
                    nested_bool: false
                    nested_list: ["nest_foo", 1.01]
                    nested_formatted_string: "id:john"

          - action: create
            expected:
              - src: "resource.equipment_request.vdefault/acme#rule-001"
                val: "create_allowed:john"
```

Fixes #1629


internal/test/testdata/verify/cases/case_011.yaml.input
```diff
@@ -0,0 +1,176 @@
...
+                  id: john
+                  keys: XX125
+                  formatted_string: "id:john"
+                  some_bool: true
+                  some_list: ["foo", "bar"]
+                  something_nested:
+                    nested_str: foo
+                    nested_bool: false
+                    nested_list: ["nest_foo", 1.01]
+                    nested_formatted_string: "id:john"
```
Sambigeara, line 91: This is more food for thought:

In this instance (by design), we're rejecting `approve` because the request doesn't have the [appropriate status](https://github.com/Sambigeara/cerbos/blob/97dc68926c2d564d8d3c2354d7ff760656bb26a1/internal/test/testdata/store/resource_policies/policy_07.yaml#L32-L41), but not returning the output.

I wonder if the output would be valuable here, because you'd be able to tell at a glance the reason for DENY (rather than by inspecting the policies afterwards to match).

Sambigeara (approved):
Left a comment for future thought, but all LGTM 👍

internal/test/testdata/verify/cases/case_011.yaml.input
charithe: I see your point but `output` is not a debugging aid. It's a signal to the API user that a particular rule was activated (all the conditions matched) during the evaluation so that they can take extra steps based on that information. If we produce outputs for all rules that were considered but not activated, then that signal is lost.

For debugging purposes we have the engine traces and the REPL.

PR #1637, opened 2023-06-12
fix: Render the condition correctly in an Engine trace

### LINKED ISSUE #1635: --verbose flag prints out incorrect information about the condition

### Current Behavior

I run the tests with the `--verbose` flag when I have a failure. If I have a condition match, any, of, the output shows the correct result of the evaluation of all the conditions but the wrong condition descriptions. The condition description is always the last condition in the list.

So if I have the condition like

```
 condition:
        match:
          any:
            of:
              - expr: 1 == 2
              - expr: 2 == 1
```

I get the following, with two `condition#1` entries

```
policy=cerbos.resource.location.vdefault > scope="" > rule=rule-001 > action=create_entry > condition > conditionAny > condition#1 > expr=`2 == 1`
    activated
    result → false

  policy=cerbos.resource.location.vdefault > scope="" > rule=rule-001 > action=create_entry > condition > conditionAny > condition#1 > expr=`2 == 1`
    activated
    result → false

  policy=cerbos.resource.location.vdefault > scope="" > rule=rule-001 > action=create_entry > condition > conditionAny
    activated
    result → false
```

### Expected Behavior

I would expect this output, with a `condition#0` and `condition#1`

```
policy=cerbos.resource.location.vdefault > scope="" > rule=rule-001 > action=create_entry > condition > conditionAny > condition#0 > expr=`1 == 2`
    activated
    result → false

  policy=cerbos.resource.location.vdefault > scope="" > rule=rule-001 > action=create_entry > condition > conditionAny > condition#1 > expr=`2 == 1`
    activated
    result → false

  policy=cerbos.resource.location.vdefault > scope="" > rule=rule-001 > action=create_entry > condition > conditionAny
    activated
    result → false
```

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- Cerbos version: 0.26.0 (I also checked 0.27.0 and the same thing happens)
- Docker version:
- Go version:
```

### Anything else?

_No response_

charithe:
The engine trace sometimes displays the wrong condition in its ouput.
This is due to the underlying slice being reused inadvertantly.

Fixes #1635

### DIFF
--- a/internal/engine/tracer/context.go
+++ b/internal/engine/tracer/context.go
@@ -143,9 +143,13 @@ func (c *context) StartOutput(ruleName string) Context {
 }
 
 func (c *context) start(component *enginev1.Trace_Component) Context {
+	components := make([]*enginev1.Trace_Component, len(c.components)+1)
+	copy(components, c.components)
+	components[len(c.components)] = component
+
 	return &context{
 		sink:       c.sink,
-		components: append(c.components, component),
+		components: components,
 	}
 }
 
--- a/internal/test/testdata/verify_junit/cases/case_007.yaml.golden
+++ b/internal/test/testdata/verify_junit/cases/case_007.yaml.golden
@@ -36,14 +36,8 @@
       <failure type="RESULT_FAILED" message="Output expectation unsatisfied">
         <outputs>
           <output src="resource.company.vdefault#rule-001">
-            <expected><![CDATA[{
-  "principal":  "blah",
-  "resource":  "yy1"
-}]]></expected>
-            <actual><![CDATA[{
-  "principal":  "user",
-  "resource":  "xx1"
-}]]></actual>
+            <expected><![CDATA[{"principal":"blah","resource":"yy1"}]]></expected>
+            <actual><![CDATA[{"principal":"user","resource":"xx1"}]]></actual>
           </output>
         </outputs>
         <actual>EFFECT_ALLOW</actual>
--- a/internal/verify/junit/junit.go
+++ b/internal/verify/junit/junit.go
@@ -11,6 +11,7 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	policyv1 "github.com/cerbos/cerbos/api/genpb/cerbos/policy/v1"
+	"github.com/tidwall/pretty"
 )
 
 const (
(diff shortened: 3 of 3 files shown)


internal/engine/tracer/context.go
```diff
@@ -143,9 +143,13 @@ func (c *context) StartOutput(ruleName string) Context {
 }
 
 func (c *context) start(component *enginev1.Trace_Component) Context {
+	components := make([]*enginev1.Trace_Component, len(c.components)+1)
+	copy(components, c.components)
+	components[len(c.components)] = component
```
Sambigeara, line 148: Would a `c.components = append(c.components, component)` not be (potentially) more efficient, rather than a full copy?

Sambigeara (dismissed):
👍

internal/engine/tracer/context.go
charithe: Unless I am missing something, that's what we originally had? The problem is that `append` could modify the first argument in place if it has capacity so it's not guaranteed to return a new slice pointer every time.

Sambigeara: I think it's to do with the explicit return-reassign. It'll try to append contiguously if there's available memory, otherwise will reallocate behind the scenes ([docs](https://pkg.go.dev/builtin#append)):

```go
// The append built-in function appends elements to the end of a slice. If
// it has sufficient capacity, the destination is resliced to accommodate the
// new elements. If it does not, a new underlying array will be allocated.
// Append returns the updated slice. It is therefore necessary to store the
// result of append, often in the variable holding the slice itself:
//
//	slice = append(slice, elem1, elem2)
//	slice = append(slice, anotherSlice...)
//
// As a special case, it is legal to append a string to a byte slice, like this:
//
//	slice = append([]byte("hello "), "world"...)
func append(slice []Type, elems ...Type) []Type
```

Sambigeara: Got this wrong, it totally is re-assigning in the struct. This is about re-use of existing slices causing unpredictable results. From Charith:

>Yeah, it's rewriting the parent. So, say the path to the any expression is `a.b`. It then needs to go to child expressions `p`, `q` and `r` and for each them the parent component should be `a.b`. If we directly append to it from the children, then the result is unpredictable.

Sambigeara (approved):

PR #1642, opened 2023-06-14
enhancement: Add filtering in the ListPolicies RPC

### LINKED ISSUE #1609: Add ability to filter policies in the Admin API ListPolicies RPC
Add the ability to filter the list of policies by name (regex?), version and scope.

Sambigeara:
- Adds support to filter by `name` and `scope` using regex, and `version` using full string matching.
- Adds application defined function for SQLite driver.

closes #1609


Sambigeara:
Whoops, forgot docs again, just adding

api/public/cerbos/request/v1/request.proto
```diff
@@ -523,6 +523,28 @@ message ListPoliciesRequest {
...
 
+message FilterPoliciesRequest {
+  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
+    json_schema: {description: "Filter policies request"}
+  };
+  bool include_disabled = 1 [
+    (google.api.field_behavior) = OPTIONAL,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Include disabled policies"}
+  ];
+  string name_regexp = 2 [
```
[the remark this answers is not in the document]
charithe: [I suggested introducing a new RPC](https://github.com/cerbos/cerbos/issues/1609#issuecomment-1582252664) because filtering can only be done on database-backed stores at the moment. My reasoning was that because the `ListPolicies` RPC works on all stores, it might be confusing if the filter option did not work for some stores.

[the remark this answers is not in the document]
haines: I feel like in terms of documentation, adding a caveat to the request fields about the supported stores is similar to adding one to an RPC itself, and it leaves the door open for us to implement this functionality for other stores without having to keep the two RPCs for backwards compatibility?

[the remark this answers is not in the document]
Sambigeara: Yeah that's an interesting one. We briefly discussed this [here](https://github.com/cerbos/cerbos/issues/1609#issuecomment-1582252664). We already support `ListPolicies` for non-mutable stores, but not other admin specific RPCs, so it might be cleaner to separate this mutable-only interface?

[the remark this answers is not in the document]
Sambigeara: Oh I didn't refresh. You guys type fast AF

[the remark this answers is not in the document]
charithe: I am torn on this. Both options have their merits. Haines' point about being able to implement the functionality to immutable stores at a later date without changing anything is compelling though. We probably don't want to end up with two almost identical RPCs in the end. So, I think I'll vote for adding the functionality to `ListPolicies`. Sorry about the confusion.

Sambigeara:
That's a fun test error

internal/storage/db/sqlite3/sqlite3.go
```diff
@@ -20,12 +21,13 @@ import (
 	"go.uber.org/zap"
 
 	// import sqlite3 driver.
```
charithe, line 23: I think we can remove this comment and let the import be grouped with the rest. The comment was only there because the linter requires comments for blank imports.

internal/storage/db/sqlite3/sqlite3.go
```diff
@@ -41,7 +43,37 @@ var (
...
 )
 
+const nRegexpFnArgs = 2
+
+var nameRegexpCache util.RegexpCache
+
 func init() {
+	nameRegexpCache = *util.NewRegexpCache()
+
+	gosqlite3.MustRegisterDeterministicScalarFunction("regexp", nRegexpFnArgs, func(ctx *gosqlite3.FunctionContext, args []driver.Value) (driver.Value, error) {
```
charithe, line 53: 😮 🙌🏽

internal/svc/admin_svc.go
```diff
@@ -121,7 +121,20 @@ func (cas *CerbosAdminService) ListPolicies(ctx context.Context, req *requestv1.
 		return nil, status.Error(codes.NotFound, "store is not configured")
 	}
 
-	policyIds, err := cas.store.ListPolicyIDs(context.Background(), req.IncludeDisabled)
+	// We've historically supported ListPolicies on non-mutable stores, but later introduced filters are not scalable.
+	// Therefore, if any of the filters in question are passed and the store is not mutable, we reject the request.
+	if _, ok := cas.store.(storage.MutableStore); !ok && (req.NameRegexp != "" || req.ScopeRegexp != "" || req.Version != "") {
+		return nil, status.Error(codes.Unimplemented, "Configured store is not mutable")
```
charithe, line 127:
```suggestion
		return nil, status.Error(codes.Unimplemented, "Store does not support filters")
```

internal/util/regexp.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"regexp"
+	"sync"
+)
+
+type RegexpCache struct {
+	cache map[string]*regexp.Regexp
+	mu    *sync.Mutex
```
charithe, line 14: I would use a `RWMutex` here 

```suggestion
	mu    sync.RWMutex
```

internal/util/regexp.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func NewRegexpCache() *RegexpCache {
+	return &RegexpCache{
+		cache: make(map[string]*regexp.Regexp),
+		mu:    &sync.Mutex{},
```
charithe, line 20: The zero value of mutex is usable out of the box so it doesn't need to be initialised.

```suggestion
		
```

internal/util/regexp.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+}
+
+// GetCompiledExpr lazily compiles (and stores) regexp.
+func (c *RegexpCache) GetCompiledExpr(re string) (*regexp.Regexp, error) {
+	if c.cache == nil {
+		c.cache = make(map[string]*regexp.Regexp)
+	}
```
charithe, line 28: Can the `cache` be nil? It's initialised in the `NewRegexpCache` function, isn't it?

If it does need to be re-initalised, shouldn't it happen inside the lock?

charithe (commented):
Nice! TIL about SQLite UDFs 🙌🏽 

I think the Admin client in the `sdk` should be updated to support filtering as part of this PR as well. 

We should update `cerbosctl get` to support filters too but that can be done in a separate PR.

api/public/cerbos/request/v1/request.proto
```diff
@@ -521,6 +521,18 @@ message ListPoliciesRequest {
...
   ];
+  string name_regexp = 2 [
+    (google.api.field_behavior) = OPTIONAL,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Filter policies by name with regexp"}
+  ];
+  string scope_regexp = 3 [
+    (google.api.field_behavior) = OPTIONAL,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Filter policies by scope with regexp"}
+  ];
+  string version = 4 [
```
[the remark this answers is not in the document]
Sambigeara: A performance thing primarily. It seemed reasonable to assume that `name` and `scope` would benefit from regex, but I went on the assumption that version would likely be generally short and repeatable. I was on the fence about implementing regex here too though, will happily update if we think it's worthwhile.

internal/storage/db/internal/db.go
```diff
@@ -655,14 +674,63 @@ func (s *dbStorage) ListPolicyIDs(ctx context.Context, includeDisabled bool) ([]
...
+		*whereExprs = append(*whereExprs, goqu.C(col).ILike(r))
+	} else {
+		*postFilters = append(*postFilters, postRegexpFilter{re: r, col: col})
+	}
+
+	return nil
+}
+
+func (s *dbStorage) regexpEnabled() bool {
+	// TODO(saml) link this to the `dialect` arg passed to `goqu` in the sqlserver package, or rethink how to indicate regexp support
```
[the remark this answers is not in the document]
Sambigeara: I meant that it's not in that we're passing string literals rather than references to the same constant. Although it's very unlikely to change 😅

internal/util/regexp.go
Sambigeara: Yeah it can't be. This was one of my OCD just-in-case things, which in hindsight is just misleading. I'll remove

api/public/cerbos/request/v1/request.proto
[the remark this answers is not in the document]
charithe: I think it would be more consistent with the rest of the filters if `version` was a regex too. 

A potential use of the `version` field is to denote environments. So, I can imagine  someone might want to do a query like `version =~ (staging|qa)`

Sambigeara:
>I think the Admin client in the sdk should be updated to support filtering as part of this PR as well.

I'm not sure where you're referring to here, I've added [additional options](https://github.com/Sambigeara/cerbos/blob/eb13e83f3e100f46ce1e6958fcf41c2d416135c6/client/model.go#L1263-L1279), is there anywhere else I'm missing?

docs/modules/api/pages/admin_api.adoc
```diff
@@ -131,10 +131,12 @@ GET /admin/policies
 
 NOTE: This endpoint is still under development and should be considered unstable.
 
-Issue a GET request to the endpoint to list the policies available in the store.
+Issue a GET request to the endpoint to list the policies available in the store. If the policy store supports filtering, you can optionally pass filter parameters to reduce the result set.
 
 Use `includeDisabled=true` query parameter in order to include disabled policies in the response.
 
+Use `nameRegexp` and `scopeRegexp` to filter using the policy name and scope with case insensitive regular expression. Use `version` to filter using the full policy version string.
```
charithe, line 138: Now `version` is a regex too

docs/modules/api/pages/admin_api.adoc
```diff
@@ -144,7 +146,7 @@ curl -k -u cerbos:cerbosAdmin \
 [source,shell]
 ----
 curl -k -u cerbos:cerbosAdmin \
-    'https://localhost:3592/admin/policies?pretty&includeDisabled=true'
+    'https://localhost:3592/admin/policies?pretty&includeDisabled=true&nameRegexp=%5Efoo&scopeRegexp=bar%24&version=default'
```
charithe, line 149:
```suggestion
    'https://localhost:3592/admin/policies?pretty&includeDisabled=true&nameRegexp=%5Efoo&scopeRegexp=bar%24&versionRegexp=default'
```

internal/storage/db/internal/db.go
```diff
@@ -655,14 +676,63 @@ func (s *dbStorage) ListPolicyIDs(ctx context.Context, includeDisabled bool) ([]
...
+	for _, f := range postFilters {
+		switch f.col {
+		case PolicyTblNameCol:
+			if !f.re.MatchString(pc.Name) {
+				return false
+			}
+		case PolicyTblScopeCol:
+			if !f.re.MatchString(pc.Scope) {
+				return false
+			}
```
charithe, line 730: Should this include the version columns as well?

internal/util/regexp.go
```diff
@@ -0,0 +1,41 @@
...
+
+	if !ok {
+		var err error
+		if r, err = regexp.Compile(re); err != nil {
+			return nil, fmt.Errorf("failed to compile regexp: %s", re)
+		}
+
+		c.mu.Lock()
+		defer c.mu.Unlock()
+		c.cache[re] = r
```
charithe, line 37: Nit: `defer` has a tiny bit of overhead and we don't really need the `defer` here.

```suggestion
		c.cache[re] = r
		c.mu.Unlock()
```

internal/storage/db/internal/db.go
Sambigeara: 🙈 thank you

Sambigeara: my tests are certainly not as good as they should be

charithe (approved):
LGTM

PR #1645, opened 2023-06-15
feat: Introduce environment-specific global variables

### LINKED ISSUE #1623: Provide environment-specific information to policies

### Feature description

[As requested in the community Slack](https://community.cerbos.dev/t/12314264/is-it-possible-to-use-environment-variables-in-policies-some#8e0feef9-7af6-40ef-becf-056bcbad3e47)

It is sometimes useful to have environment-specific rules in policies. It would be nice if the environment-specific values could be provided directly to the PDP server rather than passed through from the application in each request.

One option would be to provide a custom CEL function that wraps `os.Getenv`. This exposes the entire environment to the policies, which seems slightly wrong although the actual security impact is likely minimal (I guess the risk is that policy authors could gain access to secrets in the PDP environment that they are not supposed to be able to see?).

Perhaps a "safer" approach is to introduce a new PDP config option `globals`. This could be a key-value map, and thanks to the config interpolation it would be possible to use environment variables and even set defaults. This would be nice because those without defaults would be required - it wouldn't be possible to start the PDP without setting them.

```yaml
globals:
  foo: ${FOO}
  bar: ${BAR:}
```

These values would then be made available to CEL expressions as `globals.foo`, `globals.bar` etc

### Anything else?

_No response_

haines:
Closes #1623

This PR introduces a new configuration section, `engine.globals`, which takes a string-to-any map of global variables to be made available in policy conditions as `globals` (aliased to `G`).

### DIFF
--- a/cmd/cerbos/repl/internal/repl.go
+++ b/cmd/cerbos/repl/internal/repl.go
@@ -323,6 +323,18 @@ func (r *REPL) setSpecialVar(name, value string) error {
 		r.evalPolicyVariables()
 		r.output.PrintResult(name, r.vars[conditions.CELVariablesIdent])
 
+	case conditions.CELGlobalsIdent, conditions.CELGlobalsAbbrev:
+		var globals map[string]any
+		if err := json.Unmarshal([]byte(value), &globals); err != nil {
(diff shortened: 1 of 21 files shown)


charithe (approved):
🙌🏽

PR #1655, opened 2023-06-20
feat: Lenient scope search

### LINKED ISSUE #1587: Allow for evaluating a scope hierarchy with absent top level scope

### Feature description

Allow for policy evaluation to be performed as long as some level of the scope hierarchy is available for a resource scope. For example, given a resource scope of `a.b.c` and resource policies existing with each of scopes ``, `a`, `a.b`, the policy evaluation of the resource should use these 3 three scopes, ignoring the fact that there is no `a.b.c` policy scope defined. The current behavior results in a `NO_MATCH` and denied response.

Ideally, the default behavior would be to use whatever base hierarchy exists, skipping the higher level scopes that are not defined. An alternative to changing the default behavior (which obviously has backwards compatibility issues), is to add a request option that specifies to find the defined scope hierarchy and to ignore missing top level hierarchies.

### Anything else?

_No response_

charithe:
When configured with `--set=engine.lenientScopeSearch=true`, performs a
search down the scope tree until a policy is found. Originally, if the
request defined scope `a.b.c` a policy file with scope `a.b.c` had to
exist in the store. With this configuration if `a.b.c` doesn't exist,
the engine will try to load `a.b`, `a` and `` in that order.

Fixes #1587

### DIFF
--- a/docs/modules/configuration/pages/engine.adoc
+++ b/docs/modules/configuration/pages/engine.adoc
@@ -54,6 +54,8 @@ engine:
 
 When working with xref:policies:scoped_policies.adoc[scopes], the default behaviour of the Cerbos engine is to expect that a policy file exists for the requested scope. For example, if the API request defines `a.b.c` as the `scope`, a policy file _must exist_ in the policy repository with the `a.b.c` scope. This behaviour can be overridden by setting `lenientScopeSearch` configuration to `true`. When lenient scope search is enabled, if a policy with scope `a.b.c` does not exist in the store, Cerbos will attempt to find scopes `a.b`, `a` and `` in that order.
 
+NOTE: This setting only affects how Cerbos treats missing leaf scopes when searching for policies. The policies stored in your policy store _must_ have unbroken scope chains (for example, if you have a scoped policy `a.b.c` in the store, the policy files for scopes `a.b`, `a` and `` must also exist).
+
 
 [source,yaml,linenums]
 ----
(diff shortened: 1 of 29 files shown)


charithe:
Go home CodeCov. You're drunk 😡

docs/modules/configuration/pages/engine.adoc
```diff
@@ -48,3 +48,15 @@ engine:
   globals:
     environment: ${CERBOS_ENVIRONMENT:development}
 ----
+
+[#lenient_scopes]
+== Lenient scope search
+
+When working with xref:policies:scoped_policies.adoc[scopes], the default behaviour of the Cerbos engine is to expect that a policy file exists for the requested scope. For example, if the API request defines `a.b.c` as the `scope`, a policy file _must exist_ in the policy repository with the `a.b.c` scope. This behaviour can be overridden by setting `lenientScopeSearch` configuration to `true`. When lenient scope search is enabled, if a policy with scope `a.b.c` does not exist in the store, Cerbos will attempt to find scopes `a.b`, `a` and `` in that order.
```
haines, line 55: This could maybe use an extra sentence to clarify that this only works for leaves and doesn't remove the requirement for the policy chain to have no gaps, i.e. if you introduce a `a.b.c` you need to also have `a.b`, `a`, and `` defined too.

haines (dismissed):

haines (approved):

PR #1661, opened 2023-06-22
chore: Remove deprecated field storeIdentifer

oguzhand95:
This PR removes the -misspelled- `storeIdentifer` field. It was deprecated in the PR: #1458

### DIFF
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -168,8 +168,6 @@ func WithStoreIdentifier(p *policyv1.Policy, storeIdentifier string) *policyv1.P
 		p.Metadata = &policyv1.Metadata{}
 	}
 
-	//nolint:staticcheck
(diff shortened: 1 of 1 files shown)


charithe (commented):
I think that removing the field from the protobuf would break the Javascript SDK. @haines ?

I think we can just stop setting it in the Go code and hence get rid of that `nolint` annotation.

charithe:
Should we still set its value in code or is it OK to stop doing that?

oguzhand95:
If we are not getting rid of the misspelled field, I think we should still keep the assignment just in case.

haines:
The JS SDK won't break if we stop setting it ([it uses `storeIdentifier` if set, and only falls back to `storeIdentifer` if not](https://github.com/cerbos/cerbos-sdk-javascript/blob/0d56a835ac008fcc2896115a531b48f555a80338/packages/core/src/convert/fromProtobuf.ts#L213-L214)).

But I agree with @oguzhand95 that we should just keep setting it until we get rid of the field altogether 🤷‍♂️

PR #1667, opened 2023-06-26
feat: Add tls support kafka

shangardezi:
#### Description

Adds TLS support for Kafka Audit logs.

#### Checklist 


internal/util/tls.go
```diff
@@ -0,0 +1,40 @@
...
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"os"
+)
+
+const DefaultTLSVersion = tls.VersionTLS12
+
+func NewTLSConfig(caPath, certPath, keyPath string) (*tls.Config, error) {
```
charithe, line 13: This could probably go into the `kakfa` package just because we have a very similar piece of code in the `server` package. From a quick glance, they seem to have some tiny differences so it would make sense to follow the convention and move it closer to where it's used because it's specific to that package. https://github.com/cerbos/cerbos/blob/b0904d29cce5dcf68706e74274951473294ba568/internal/server/server.go#L321


FYI the default TLS config use by Cerbos is defined here: https://github.com/cerbos/cerbos/blob/b0904d29cce5dcf68706e74274951473294ba568/internal/util/network.go#L31

internal/audit/kafka/tls.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to load TLS key pair: %w", err)
+	}
+	reloader.cert = &cert
+
+	go func() {
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, syscall.SIGHUP)
+		for range c {
+			if err := reloader.reload(); err != nil {
```
rcrowe, line 40: What about instead using a timer that I can configure how often the certificate is reloaded? 
Tools like cert-manager will reload the certificates in-place on disk.

internal/audit/kafka/tls.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read CA certificate: %w", err)
+	}
+
+	caCertPool := x509.NewCertPool()
+	if !caCertPool.AppendCertsFromPEM(caCert) {
+		return nil, fmt.Errorf("failed to append CA certificate")
+	}
+
+	return &tls.Config{
```
rcrowe, line 81: `InsecureSkipVerify` might be a helpful addition that can be configured

internal/audit/kafka/tls.go
shangardezi: https://github.com/cerbos/cerbos/pull/1667/commits/c6138adac080eca256afeede1a71be20085fe739 lmk what you think 👍

internal/audit/kafka/tls.go
shangardezi: https://github.com/cerbos/cerbos/pull/1667/commits/03d4b466a36171e0a9281296ae327fc74f4b1136

charithe:
Hey, I am currently traveling. If no one else from the Cerbos team picks it up, I'll review ASAP.

internal/audit/kafka/conf.go
```diff
@@ -56,6 +75,10 @@ func (c *Conf) SetDefaults() {
 	c.ClientID = defaultClientID
 	c.MaxBufferedRecords = defaultMaxBufferedRecords
 	c.Compression = []string{CompressionSnappy, CompressionNone}
+
+	if c.Authentication.TLS != nil {
+		c.Authentication.TLS.InsecureSkipVerify = false
+	}
```
haines, line 81: I don't think this necessary? Won't the newly-created conf always have `c.Authentication.TLS == nil` (and `false` is the default value for booleans anyway)?

```suggestion
```

internal/audit/kafka/tls.go
```diff
@@ -0,0 +1,110 @@
...
+
+			r.mu.Lock()
+			defer r.mu.Unlock()
+			r.cert = &cert
+		}
+	}
+}
+
+func NewTLSConfig(ctx context.Context, reloadInterval time.Duration, insecureSkipVerify bool, caPath, certPath, keyPath string) (*tls.Config, error) {
+	if caPath == "" || certPath == "" || keyPath == "" || reloadInterval == 0 {
```
haines, line 74: I think `reloadInterval == 0` should probably be valid, and taken to mean "do not reload the TLS certs".

Also, would it be reasonable to support TLS encryption without a client cert (i.e. only requiring `caPath`, but requiring both `certPath` and `keyPath` if either one of them is set)?

internal/audit/kafka/tls.go
```diff
@@ -0,0 +1,110 @@
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+
+		case <-ticker.C:
+			cert, err := tls.LoadX509KeyPair(r.certPath, r.keyPath)
+			if err != nil {
+				return fmt.Errorf("failed to load TLS key pair: %w", err)
```
haines, line 63: I don't think failing to load the files should terminate the reload loop, in case the failure is transient. Otherwise, it'll fail and the only way to recover is to restart the PDP. I reckon we should move the logging in here and keep looping.

internal/audit/kafka/tls.go
```diff
@@ -0,0 +1,110 @@
...
+	"fmt"
+	"os"
+	"sync"
+	"time"
+
+	"github.com/cerbos/cerbos/internal/observability/logging"
+	"go.uber.org/zap"
+)
+
+const DefaultTLSVersion = tls.VersionTLS12
```
haines, line 20: Does this need to be exported? I would maybe inline it.

internal/audit/kafka/tls.go
```diff
@@ -0,0 +1,110 @@
...
+
+type TLSReloader struct {
```
haines, line 22: I think this can be unexported.

internal/audit/kafka/kafka_test.go
```diff
@@ -152,12 +202,71 @@ func TestAsyncProduce(t *testing.T) {
...
+	require.NoError(t, err, "Failed to start container")
+
+	t.Cleanup(func() {
+		_ = pool.Purge(resource)
+	})
+
+	brokerDSN := fmt.Sprintf("localhost:%d", hostPort)
+	duration := 10 * time.Second
+	skipVerify := false
+	tlsConfig, err := kafka.NewTLSConfig(context.Background(), duration, skipVerify, caPath, certPath, keyPath)
```
haines, line 255: I think this leaks a goroutine because the context is never canceled.

```suggestion
	ctx, cancel := context.WithCancel(context.Background())
	t.Cleanup(cancel)
	tlsConfig, err := kafka.NewTLSConfig(ctx, duration, skipVerify, caPath, certPath, keyPath)
```

(alternatively we could set the reload duration to 0, if we allow that as suggested)

internal/audit/kafka/conf.go
shangardezi: It's not necessary. The goal of this was to make it clear that this is the default. Will remove 👍

internal/audit/kafka/tls.go
shangardezi: That's reasonable 👍 
Let me know what you think https://github.com/cerbos/cerbos/pull/1667/commits/443c9bcc143c813cf7d52aa6cf96947e88e1afae

internal/audit/kafka/tls.go
shangardezi: Good spot - thanks!
https://github.com/cerbos/cerbos/pull/1667/commits/c5975d396505131a787bd7fd733c0b64ae702767

internal/audit/kafka/conf.go
```diff
@@ -23,10 +23,29 @@ const (
...
+	TLS *TLS `yaml:"tls"`
+}
+
+type TLS struct {
+	// CAPath is the path to the CA certificate.
+	CAPath string `yaml:"caPath" conf:"required,example=/path/to/ca.crt"`
+	// CertPath is the path to the client certificate.
+	CertPath string `yaml:"certPath" conf:"required,example=/path/to/tls.cert"`
+	// KeyPath is the path to the client key.
+	KeyPath string `yaml:"keyPath" conf:"required,example=/path/to/tls.key"`
```
haines, line 36: The cert and key paths are no longer required, so we can remove that from the struct tags. Note that here and for reload interval we need a leading comma (i.e. `,example=...`) to avoid that `<DEFAULT_VALUE_NOT_SET>` in the conf docs.

internal/audit/kafka/tls.go
```diff
@@ -0,0 +1,134 @@
...
+
+	caCertPool := x509.NewCertPool()
+	if !caCertPool.AppendCertsFromPEM(caCert) {
+		return nil, fmt.Errorf("failed to append CA certificate")
+	}
+
+	// #nosec G402
+	tlsConfig := &tls.Config{
+		RootCAs:            caCertPool,
+		ClientCAs:          caCertPool,
```
haines, line 42: I don't think we need to set `ClientCAs` - my understanding is that's just for use on the server end of the connection.

>ClientCAs defines the set of root certificate authorities that servers use if required to verify a client certificate by the policy in ClientAuth.

internal/policy/policy.go
```diff
@@ -168,7 +168,6 @@ func WithStoreIdentifier(p *policyv1.Policy, storeIdentifier string) *policyv1.P
 		p.Metadata = &policyv1.Metadata{}
 	}
 
-	//nolint:staticcheck
```
haines, line 171: Unfortunately the linter is [flaky](https://github.com/golangci/golangci-lint/issues/3228) - this nolint is actually correct and still needed.

internal/audit/kafka/tls.go
```diff
@@ -0,0 +1,134 @@
...
+	return reloader, nil
+}
+
+func (r *tlsReloader) reload(ctx context.Context) {
+	ticker := time.NewTicker(r.reloadInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
```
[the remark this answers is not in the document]
shangardezi: ha, good spot :D

internal/policy/policy.go
shangardezi: Ah my bad, that's me running `make lint`. Restored.

haines (approved):
Awesome, looks good to me! Thanks for the contribution 🙌

shangardezi:
> Awesome, looks good to me! Thanks for the contribution 🙌

Thanks for the great review 🙇

PR #1686, opened 2023-07-12
feat: Add cerbosctl command to export policies and schemas from store

oguzhand95:
# Exports the policies and schemas from the store into a directory.

## Export policies and schemas from the store into a directory
```
cerbosctl store export path/to/dir
```

## Export policies and schemas from the store into a zip archive
```
cerbosctl store export path/to/archive.zip
```

## Export policies and schemas from the store into a gzip archive
```
cerbosctl store export path/to/archive.gzip
cerbosctl store export path/to/archive.tar.gz
```


cmd/cerbosctl/store/export/export.go
```diff
@@ -0,0 +1,110 @@
...
+	defer exporter.Close()
+
+	if err := client.BatchAdminClientCall2(context.Background(), clientCtx.AdminClient.GetPolicy, func(ctx context.Context, policies []*policyv1.Policy) error {
+		if policies == nil {
+			return fmt.Errorf("failed to retrieve policy: %w", err)
+		}
+
+		for _, p := range policies {
+			name := strings.ReplaceAll(p.Metadata.StoreIdentifier, "/", "_")
+			if !strings.HasSuffix(name, ".yaml") {
```
charithe, line 63: If a user tries this on a `disk` store that has JSON policies, won't we end up creating a file like `foo.json.yaml`? Since disk-based stores are the only stores that return file names as IDs, I think we should probably honour the actual file type when exporting (JSON -> JSON, YAML -> YAML)

cmd/cerbosctl/store/export/internal/exporter.go
```diff
@@ -0,0 +1,137 @@
...
+	case util.IsTar(path):
+		e.tarWriter = tar.NewWriter(e.file)
+	case util.IsZip(path):
+		if e.file, err = os.Create(path); err != nil {
+			return nil, fmt.Errorf("failed to create archive file %s: %w", path, err)
+		}
+
+		e.zipWriter = zip.NewWriter(e.file)
+	default:
+		if err := os.MkdirAll(filepath.Join(path, util.SchemasDirectory), os.ModePerm); err != nil {
```
charithe, line 131: What if there are no schemas in the store? I think we should only create the schema directory if the store has them.

cmd/cerbosctl/store/export/internal/exporter.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	return e.write(name, data)
+}
+
+func (e *Exporter) write(name string, data []byte) error {
+	var w io.Writer
+	var err error
+	switch {
+	case util.IsTar(e.path) || util.IsGzip(e.path):
```
charithe, line 46: You know what kind of file this is at the point you create the exporter using `NewExporter`. So, you could store that information in the struct as a bool and check that instead of performing this name check over and over again for each file you are adding.

I'd probably design this in a different way though. I'd make `Exporter` an interface and then have three different implementation of it for tar, zip and dir. When `NewExporter` is called, based on the output filename, I'd return the correct implementation of `Exporter` back to the client. That way the code for each output format is nicely separated and you don't need to wade through a bunch of if statements.

cmd/cerbosctl/store/export/internal/exporter.go
```diff
@@ -0,0 +1,253 @@
...
+		return newZipExporter(file), nil
+	}
+
+	return newDirectoryExporter(path), nil
+}
+
+type Exporter interface {
+	WriteJSON(name string, jsonData []byte) error
+	WriteYAML(name string, jsonData []byte) error
+	Close() error
```
charithe, line 45:
```suggestion
	io.Closer
```

cmd/cerbosctl/store/reload/reload.go
```diff
@@ -12,24 +12,24 @@ import (
...
+const help = `# Reload the store
 cerbosctl store reload
 
 # Reload the store and wait until it finishes
 cerbosctl store reload --wait`
 
-type ReloadCmd struct {
-	Wait bool `help:""`
+type Cmd struct {
+	Wait bool `help:"Wait until the reloading process finalizes"`
```
charithe, line 22:
```suggestion
	Wait bool `help:"Wait until the reloading process finishes"`
```

cmd/cerbosctl/store/export/export.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	defer exporter.Close()
+
+	if err := client.BatchAdminClientCall2(context.Background(), clientCtx.AdminClient.GetPolicy, func(ctx context.Context, policies []*policyv1.Policy) error {
+		if policies == nil {
+			return fmt.Errorf("failed to retrieve policy: %w", err)
+		}
+
+		for _, p := range policies {
+			name := strings.ReplaceAll(p.Metadata.StoreIdentifier, "/", "_")
```
charithe, line 62: You need to detect whether the returned IDs are file paths first. By replacing the `/`, we lose directory information for disk-based stores. I think we should try to make the layout of the export as close to the original layout as possible. 

You could check whether the policy has a scope and, if so, whether the identifier has more than one `/` characters.

charithe (approved):
LGTM. 

It would look more responsive if we output the name of each policy as it's being written to its destination.

cmd/cerbosctl/store/export/export.go
```diff
@@ -79,6 +78,8 @@ func (c *Cmd) Run(k *kong.Kong, clientCtx *internalclient.Context) error {
 					return fmt.Errorf("failed to write policy %s: %w", name, err)
 				}
 			}
+
+			_, _ = fmt.Fprintf(k.Stdout, "Successfully wrote policy %s\n", name)
```
charithe, line 82:
```suggestion
			_, _ = fmt.Fprintf(k.Stdout, "🗸 %s (policy)\n", name)
```

cmd/cerbosctl/store/export/export.go
```diff
@@ -97,14 +98,21 @@ func (c *Cmd) Run(k *kong.Kong, clientCtx *internalclient.Context) error {
 			if err := exporter.WriteJSON(name, pretty.Bytes()); err != nil {
 				return fmt.Errorf("failed to write schema %s: %w", name, err)
 			}
+
+			_, _ = fmt.Fprintf(k.Stdout, "Successfully wrote schema %s\n", name)
```
charithe, line 102:
```suggestion
			_, _ = fmt.Fprintf(k.Stdout, "🗸 %s (schema)\n", name)
```

cmd/cerbosctl/store/export/export.go
```diff
@@ -97,14 +98,21 @@ func (c *Cmd) Run(k *kong.Kong, clientCtx *internalclient.Context) error {
...
 		return fmt.Errorf("error while getting schemas: %w", err)
 	}
 
-	_, _ = fmt.Fprintf(k.Stdout, "Successfully wrote %d policies and %d schemas to %s\n", len(policies), len(schemas), c.Path)
+	if len(policies) > 0 || len(schemas) > 0 {
+		_, _ = fmt.Fprintf(k.Stdout, "\nSuccessfully wrote %d policies and %d schemas to %s\n", len(policies), len(schemas), c.Path)
+	} else {
+		_, _ = fmt.Fprintf(k.Stdout, "No policies and schemas found in the store\n")
+	}
+
```
charithe, line 115:
```suggestion
	_, _ = fmt.Fprintf(k.Stdout, "\nExported %d policies and %d schemas to %s\n", len(policies), len(schemas), c.Path)
```

charithe (approved):

PR #1711, opened 2023-07-24
fix: Evict policies from cache after disable or enable

### LINKED ISSUE #1710: Disabled/enabled policies stays in the cache until the store is reloaded
If a policy is disabled/enabled using `DisablePolicy`, `EnablePolicy` it stays in the cache until the store is reloaded.

oguzhand95:
- Fixes the problem with `Disable` and `Enable`. (resolves #1710)

### DIFF
--- a/internal/compile/manager.go
+++ b/internal/compile/manager.go
@@ -93,7 +93,7 @@ func (c *Manager) processUpdateQueue(ctx context.Context) {
 			case storage.EventReload:
 				c.log.Info("Purging compile cache")
 				c.cache.Purge()
-			case storage.EventAddOrUpdatePolicy, storage.EventDeletePolicy:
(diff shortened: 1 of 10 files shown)


internal/compile/manager.go
```diff
@@ -93,7 +93,7 @@ func (c *Manager) processUpdateQueue(ctx context.Context) {
 			case storage.EventReload:
 				c.log.Info("Purging compile cache")
 				c.cache.Purge()
-			case storage.EventAddOrUpdatePolicy, storage.EventDeletePolicy:
+			case storage.EventAddOrUpdatePolicy, storage.EventDeleteDisableOrEnablePolicy:
```
charithe, line 96: It should be just `EventDeleteOrDisablePolicy`. Enabling a policy works differently. I think that's covered by `AddOrUpdatePolicy`.

PR #1716, opened 2023-07-26
docs: Stop building docs for older versions

charithe:
Exclude versions before 0.26 as they are quite old and slow down the
build process for not much benefit.

Also fixes some of the warnings from Antora on the `main` branch.

### DIFF
--- a/docs/antora-playbook.yml
+++ b/docs/antora-playbook.yml
@@ -9,7 +9,7 @@ site:
 content:
   sources:
     - url: ./../
-      branches: [HEAD, 'v*', '!v0.[0-25].*']
+      branches: [HEAD, 'v*', '!v0.{0..25}']
       edit_url: 'https://github.com/cerbos/cerbos/tree/main/{path}'
       start_path: docs
     - url: https://github.com/cerbos/cloud-docs.git
--- a/docs/antora-playbook.yml
+++ b/docs/antora-playbook.yml
@@ -9,8 +9,8 @@ site:
 content:
   sources:
     - url: ./../
-      branches: [HEAD, 'v*']
-      edit_url: 'https://github.com/cerbos/cerbos/tree/{refname}/{path}'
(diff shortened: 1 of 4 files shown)


docs/antora-playbook.yml
```diff
@@ -9,8 +9,8 @@ site:
 content:
   sources:
     - url: ./../
-      branches: [HEAD, 'v*']
-      edit_url: 'https://github.com/cerbos/cerbos/tree/{refname}/{path}'
+      branches: [HEAD, 'v*', '!v0.[0-25].*']
```
haines, line 12: Aren't the branches named e.g. `v0.25`, so the trailing `.*` should be omitted?

haines: Also, are you sure about this pattern syntax? I couldn't find it in the [docs](https://docs.antora.org/antora/latest/playbook/content-refname-matching/).

charithe: Good point about the branch name 👍🏽 
I am following this: https://docs.antora.org/antora/latest/playbook/content-branches/#default. It seems to suggest that the pattern is a regex. It does work in the local build. But it could be just a lucky coincidence.

haines: I don't think so, I think it's an extended glob. I think we need `!v0.{0..25}`.

charithe: Well, even `!v0.{0..25}` seems to work. I am not quite sure whether they are using globs or regexes 🤷🏽

haines: Their examples aren't valid regexps. The doc you're on links to the one I looked at:

>Antora also supports matching branch names using exclusions, braces, alternation, ranges, and repetition patterns. See Refname Matching in Content Sources.

charithe: Yep, shouldn't have skimmed the docs in between calls 🙈

haines (approved):

PR #1717, opened 2023-07-26
chore: Upgrade to CEL 0.17

### LINKED ISSUE #1713: Upgrade to CEL 0.17
We need to replace uses of some of the deprecated functions and members but that's pretty trivial.

All tests pass except for partial evaluation of `now() > timestamp("2021-04-20") && R.attr.geo in ["GB", "US"]`. There's a small bug in the CEL library (https://github.com/google/cel-go/issues/785) but even with that bug fixed, we might not get the expected result from the new CEL version. It would return `timestamp("CURRENT_TIMESTAMP") > timestamp("2021-04-20") && R.attr.geo in ["GB", "US"]` rather than the expected result of `R.attr.geo in ["GB", "US"]`.

charithe:
Fixes #1713


dbuduev (approved):
LGTM

PR #1723, opened 2023-08-02
chore: Mark auxData parameter as optional in OpenAPI spec

charithe:
Need to use a custom extension to signal that certain fields should be
omitted from example generation. However, this is still broken because
Rapidoc doesn't handle the case where request parameters are references
to other objects.

Also includes some theming changes to match Cerbos branding.

image


schema/assets/ui.html
```diff
@@ -7,12 +7,19 @@
...
+      default-schema-tab="schema"
+      schema-description-expanded="true"
       allow-authentication="true"
       allow-server-selection="false"
+      allow-spec-file-download="true"
       show-header="false"
-      schema-description-expanded="true"
-      theme="dark"></rapi-doc>
+      primary-color="#ffb901"
+      nav-bg-color="#ffc11e"
```
oguzhand95, line 19:
```suggestion
      nav-bg-color="#ffc11e"
      nav-text-color="#000000"
```
The navbar text color looks a bit washed out.

Before
---
image

After
---
image

oguzhand95 (approved):

PR #1724, opened 2023-08-04
feat(client): Add options for intercepting gRPC operations

haines:
I want to get OpenTelemetry client spans for Cerbos gRPC operations in our application, and propagate the span context to the Cerbos PDP. To do so, I need to configure interceptors ([`otelgrpc.StreamClientInterceptor`](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc#StreamClientInterceptor) and  [`otelgrpc.UnaryClientInterceptor`](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc#UnaryClientInterceptor)).

This PR allows gRPC interceptors to be passed to the client constructors.


charithe (dismissed):

internal/server/server.go
```diff
@@ -443,6 +444,7 @@ func (s *Server) mkGRPCServer(log *zap.Logger, auditLog audit.Log) (*grpc.Server
 		grpc.ChainStreamInterceptor(
 			grpc_recovery.StreamServerInterceptor(),
 			telemetryInt.StreamServerInterceptor(),
+			otelgrpc.StreamServerInterceptor(),
```
haines, line 447: It turns out that the approach of using the client interceptors doesn't work out of the box, because we only read the tracing headers from the HTTP request, whereas the client injects them into the gRPC request metadata.

charithe (approved):

PR #1751, opened 2023-08-15
chore(ci): Switch to Coveralls

charithe:
Coveralls has better support for recording and aggregating parallel test runs.

Also adds a filter to leave out generated code, test support code and, non-critical and untestable utility code like logging and metrics from test coverage reporting to reduce the noise.  


Sambigeara (approved):

charithe:
Yeah, I noticed that too. I am hoping that the E2E test run would fix that tonight.

charithe:
Oh, actually, E2E doesn't run all the unit tests so the reported number would be quite low 😞

PR #1755, opened 2023-08-16
enhancement!: Use `.cerbos.yaml` as conventional name for config file

haines:
This PR replaces the various config filenames with `.cerbos.yaml` across the documentation and wider codebase. Using a consistent and distinctive filename will make it easier to add IDE support for these files. Making the config file hidden means it won't be mistaken for a policy, and it also has symmetry with `.cerbos-cloud.yaml`.

I've flagged it as a breaking change because it also changes to the default config file that `cerbos run` looks for. We could alternatively look for both `.cerbos.yaml` and `cerbos.yaml` if we'd rather keep backwards compatibility.

I ran into a hiccup where Antora doesn't include hidden files so to be able to include the file in the photo-share example we have to remove the leading `.`.


charithe (dismissed):
🙌🏽

charithe (approved):

PR #1762, opened 2023-08-22
enhancement: Relax naming conventions for resource kinds, principals and roles

### LINKED ISSUE #1695: Principal ID regex is too restrictive

### Feature description

The principal policy protobuf definition has a regex for a principal (principal ID):

 ```protobuf
message PrincipalPolicy {
  string principal = 1 [(validate.rules).string = {
    pattern: "^[[:alpha:]][[:word:]\\@\\.\\-]*(\\:[[:alpha:]][[:word:]\\@\\.\\-]*)*$",
    min_len: 1
  }];
  string version = 2 [(validate.rules).string.pattern = "^[[:word:]]+$"];
  repeated PrincipalRule rules = 3;
  string scope = 4 [(validate.rules).string.pattern = "^([[:alnum:]][[:word:]\\-]*(\\.[[:word:]\\-]*)*)*$"];
  Variables variables = 5;
}
```

This regex forbids the Amazon Resource Names (ARNs) pattern, which may include consecutive colons `::` and a slash `/`. 
For example, `arn:aws:iam::acct123-44:user/567-99803848`.

Less restrictive regex.

### Anything else?

_No response_

### PR DESCRIPTION
**Pull request by charithe**
Relax the naming conventions for resource kinds, principals and roles in policy definitions.

For historical reasons, we had to make sure that policy names were valid identifiers. We no longer need to do that. However, because policy identifiers were computed using sanitised policy names, we still have to do sanitisation for "old style" names to avoid breaking existing users.

Sanitisation could result in ID collisions so the database stores have been updated to detect those cases as well.

Fixes #1695

### CODE REVIEW

File: internal/storage/db/sqlite3/sqlite3.go [hunk 1]
```diff
@@ -130,6 +132,76 @@ func runMigrations(db *sql.DB) error {
...
+			if _, err := tx.Update(internal.PolicyTbl).
+				Prepared(true).
+				Set(pr).
+				Where(goqu.And(
+					goqu.C(internal.PolicyTblIDCol).Eq(pr.ID),
+					goqu.C(internal.PolicyTblNameCol).Eq(pr.Name),
+				)).Executor().ExecContext(ctx); err != nil {
+				return fmt.Errorf("failed to update policy %s: %w", p.FQN, err)
+			}
+		*/
```
dbuduev, line 180: Do we need this commented-out code block?

internal/test/testdata/store/resource_policies/policy_04.yaml
```diff
@@ -5,7 +5,7 @@ resourcePolicy:
   version: "default"
   importDerivedRoles:
     - apatr_common_roles
-  resource: "album:object"
+  resource: "album_object"
```
dbuduev, line 8: Why do we need to change it? `album:object` is valid, isn't it?

dbuduev (dismissed):
LGTM

internal/test/testdata/store/resource_policies/policy_04.yaml
charithe: Careless usage of `sed` 😄

Sambigeara (commented):
Further to my comment, I can see that we protect against name collisions during `upsert` ops, so IIUC, my `album/object == album-object` case would ultimately fail at this point?

internal/namer/namer.go
```diff
@@ -229,9 +233,16 @@ func withScope(fqn, scope string) string {
...
+// sanitize replaces special characters in the string with underscores.
+// Before Cerbos 0.30 the names of resources or principals had to follow a certain pattern. We then replaced some of
+// the non-word characters with underscores because earlier versions of Cerbos used to generate Rego code for policies.
+// Because we used the sanitized name for computing the module ID of the policy, in order to maintain backward compatibility
+// and not break database stores we still have to do the same if the name matches the pattern.
+func sanitize(v string) string {
+	if oldNamePattern.MatchString(v) {
+		return invalidIdentifierChars.ReplaceAllLiteralString(v, "_")
+	}
+	return v
```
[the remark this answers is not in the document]
charithe: Yes, unfortunately there will be some false positives. For disk-based stores, the `compile` stage would complain about finding duplicate policies. For DB stores, the insert would fail because of the upsert logic added in this PR.

Sambigeara (approved):
👍

PR #1767, opened 2023-08-25
enhancement: Configuration to disable API explorer

charithe:
Several users have asked whether it's possible to disable the API
explorer UI. This PR adds a configuration option to disable it.

### DIFF
--- a/internal/server/conf.go
+++ b/internal/server/conf.go
@@ -64,6 +64,8 @@ type Conf struct {
 	LogRequestPayloads bool `yaml:"logRequestPayloads" conf:",example=false"`
 	// PlaygroundEnabled defines whether the playground API is enabled.
 	PlaygroundEnabled bool `yaml:"playgroundEnabled" conf:",ignore"`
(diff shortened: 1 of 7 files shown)


internal/server/conf.go
```diff
@@ -64,6 +64,8 @@ type Conf struct {
 	LogRequestPayloads bool `yaml:"logRequestPayloads" conf:",example=false"`
 	// PlaygroundEnabled defines whether the playground API is enabled.
 	PlaygroundEnabled bool `yaml:"playgroundEnabled" conf:",ignore"`
+	// DisableAPIExplorer disables the API explorer UI from being served on the HTTP port.
+	DisableAPIExplorer bool `yaml:"disableAPIExplorer" conf:",example=false"`
```
haines, line 68: We have `*Enabled` or `*Disabled` for other bools so maybe this should be `APIExplorerEnabled` with a default of `true` to match `MetricsEnabled`?

haines (approved):

Issue #1746, opened 2023-08-11
Variables referencing variables produce nondeterministic output


haines:
https://play.cerbos.dev/p/1Jx64d611fcEQnzV6OBENU77GA1vKzm3

We store variables in a map, so the order in which they are defined is nondeterministic. Unfortunately, when they are evaluated this seems to result in nondeterministic output if a variable references another variable.

In the playground above, we have the following simple policy:

```yaml
apiVersion: api.cerbos.dev/v1
resourcePolicy:
  version: default
  resource: example_resource

  variables:
    local:
      one: V.two
      two: V.three
      three: "true"

  rules:
    - actions:
        - one
      effect: EFFECT_ALLOW
      roles:
        - USER
      condition:
        match:
          expr: V.one

    - actions:
        - two
      effect: EFFECT_ALLOW
      roles:
        - USER
      condition:
        match:
          expr: V.two

    - actions:
        - three
      effect: EFFECT_ALLOW
      roles:
        - USER
      condition:
        match:
          expr: V.three
```

The `three` action is always allowed (as expected), but the `one` and `two` actions flip between allowed and denied on each evaluation.

haines:
@mdugue I'll take a look at this today.

Issue #1650, opened 2023-06-17
New Go SDK module free from indirect dependencies


rcrowe:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

When pulling in `github.com/cerbos/cerbos/client` it's currently pulling 81 indirect dependencies, a lot of which is not required for the client to operate over the gRPC API.

It would be great if there was a Go module specific to the client.

### What would the ideal solution look like to you?

**go.mod:**

```shell
module github.com/cerbos/cerbos/client
```

charithe:
This is something we've wanted to do for a while. The tricky part is extricating the calls to the internal packages in the main Cerbos module. Do you use the testing features provided by the SDK such as starting up an in-process Cerbos instance? I am thinking that the path of least resistance would be to split the API client into its own module without the testing and other utility features that it currently includes.

rcrowe:
To be honest, we do not actually use the current Client SDK because of this & a few other challenges for us, instead, we wrap the gRPC API ourselves.

Having test/debug functionality as a separate module that requires deeper integration with Cerbos seems fair.

charithe:
https://github.com/cerbos/cerbos-sdk-go

PR #1771, opened 2023-08-29
chore: Update Go to v1.21

haines:
This PR updates Cerbos to require Go v1.21, so that we can use e.g. the new [slices](https://pkg.go.dev/slices) package.

### DIFF
--- a/.github/actions/setup-go/action.yaml
+++ b/.github/actions/setup-go/action.yaml
@@ -30,7 +30,7 @@ runs:
     - name: Install Go
       uses: actions/setup-go@v4
       with:
-        go-version: 1.20.x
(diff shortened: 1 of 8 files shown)


haines:
The client SDK should probably support all supported Go versions (i.e. currently 1.20 and 1.21). Perhaps we should split that into a separate module and test it in CI against both versions?

charithe:
Splitting the client out seems like a necessity now. See also #1650.

haines:
Cool, I've closed this for now since I realised I bit off more than I could chew. I'll revisit it once I've finished with #1746.

charithe:
I'll look into splitting the client. It's been languishing in my list for a while now.

PR #1774, opened 2023-08-30
fix: Normalize Git store subdirectory config to handle leading `./` correctly

### LINKED ISSUE #1773: Handle leading `./` on Git store `subDir` config consistently
In the `git` store, on initial load it doesn't matter whether the `subDir` config is set to `.` or starts with `./`. However, on subsequent change detection, if `subDir` is set to `.` or starts with `./`, all files will be ignored because the change path does not have a `./` prefix.

https://github.com/cerbos/cerbos/blob/9652c909f1e07e0d55e851fec800071ac0ddae57/internal/storage/git/store.go#L451-L458

haines:
Fixes #1773

### DIFF
--- a/internal/storage/git/conf.go
+++ b/internal/storage/git/conf.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


charithe (approved):
Thanks for looking into it ⭐

PR #1778, opened 2023-09-01
feat!: Make `runtime.effectiveDerivedRoles` available in CEL expressions

haines:
This PR adds support for `request.principal.derivedRoles` in conditions. The value of the field is a lexicographically sorted list of the effective derived roles (always an empty list outside of resource policy conditions).

A caveat is that we prune imported but unreferenced derived roles at compile time, so this will only ever include derived roles that are referenced by at least one rule in the policy.

To make this work, I've unified the type of the `request` variable in the evaluator and planner by introducing an explicit type, `cerbos.engine.v1.Request`. This is a subset of fields available on both the `CheckInput` and `PlanResourcesInput` types.

I've also fixed a bug in the query planner where `request.auxData` didn't work. I couldn't work out why the `JSONFieldProvider` isn't working in the planner - from all my attempts to create a minimal reproduction of the issue, it _should_ work - so I've added a workaround instead of a proper fix.

### Potentially breaking changes

- I've omitted several fields (e.g. `request.actions`, `request.resource.policyVersion`) that would've been available previously. These don't seem useful in conditions as they have already been used by the engine to determine which conditions to evaluate.

- I've introduced an alias for the ternary conditional operator which previously appeared in query plans as the raw CEL function name `_?_:_` - this is now called `if`.


internal/engine/evaluator.go
```diff
@@ -158,6 +153,21 @@ func (rpe *resourcePolicyEvaluator) Evaluate(ctx context.Context, tctx tracer.Co
 			dctx.Activated()
 		}
 
+		if len(effectiveDerivedRoles) > 0 {
+			// populate `P.derivedRoles` for use in rules
+			for derivedRole := range effectiveDerivedRoles {
+				request.Principal.DerivedRoles = append(request.Principal.DerivedRoles, derivedRole)
+			}
+			sort.Strings(request.Principal.DerivedRoles)
```
charithe, line 161: Could we save a few CPU cycles by not sorting this? Most of the time it's going to be unused and when it's do used, it's most likely to be in an expression like `x in P.derivedRoles`.

internal/engine/planner/planner.go
```diff
@@ -236,20 +243,25 @@ func (rpe *ResourcePolicyEvaluator) EvaluateResourcesQueryPlan(ctx context.Conte
...
 					return node, nil
-				},
-				node: nil,
+				}),
 			})
 		}
 
+		sort.Slice(derivedRoles, func(i, j int) bool {
+			return derivedRoles[i].role < derivedRoles[j].role
+		})
```
charithe, line 261: Again, probably not necessary?

charithe (commented):
Nice one 🙌🏽 

I wonder whether users will be confused by the fact that not all derived roles they imported are available in `P.derivedRoles`. What if we rename it to `P.effectiveDerivedRoles` or something along those lines instead?

haines:
> I wonder whether users will be confused by the fact that not all derived roles they imported are available in `P.derivedRoles`. What if we rename it to `P.effectiveDerivedRoles` or something along those lines instead?

I also am concerned about this. I can't think of a simple name that clearly communicates this though - I don't think `effectiveDerivedRoles` is obvious without the caveat in the docs either.

I think it might be preferable to instead remove this caveat by ensuring that `P.derivedRoles` contains all the imported derived roles that are matched.

We could do this in a blunt force manner by disabling the compiler optimisation if we find a condition/output referencing `P.derivedRoles`, but it seems unfortunate to incur a performance penalty on all rules in the policy.

Perhaps a good approach would to rewrite the expressions at compile time (in a similar way to what I have done in the planner). We can supply a map of the effective derived roles as a variable to the CEL environment to avoid re-evaluating the conditions that we have already evaluated.

So, we would expand `P.derivedRoles` to 

```
("referenced_derived_role" in _effective_derived_roles ? ["referenced_derived_role"] : []) +
(condition_for_unreferenced_derived_role ? ["unreferenced_derived_role"] : []) +
...
```

This would also allow us to sort the list at compile time rather than runtime.

We could also optimize the `foo in P.derivedRoles` construct at compile time. If foo is not a string literal, then we expand it to

```
(foo == "referenced_derived_role" && "referenced_derived_role" in _effective_derived_roles) ||
(foo == "unreferenced_derived_role" && condition_for_unreferenced_derived_role) ||
...
```

If `foo` is a string literal, then this would simply be

```
"referenced_derived_role" in _effective_derived_roles
``` 

or 

```
condition_for_unreferenced_derived_role
```

internal/engine/evaluator.go
haines: I think it's nicer if it's stable (it could be useful in outputs too). However, I think we can do it at compile time instead: https://github.com/cerbos/cerbos/pull/1778#issuecomment-1719097314

internal/engine/planner/planner.go
haines: In this case, I actually think this is beneficial outside of the `P.derivedRoles` implementation. By having a nondeterministic ordering of derived roles, the output from the planner is not stable (given a rule that references multiple derived roles, the order that the corresponding conditions appear in the AST is similarly nondeterministic).

Having a stable AST is potentially desirable for consumers who want to be able to transform the query plan to some executable form and cache the result?

Given that there are only going to be a handful of roles in the list, I don't think the overhead is too bad.

charithe:
So, `P.derivedRoles` would effectively be a macro? I suppose that could work but I am a bit concerned about a few things.
- Invalid or really complicated expressions in unreferenced derived roles could produce an invalid expression when expanded inside the condition expression and produce cryptic errors or surprising behaviours
- Would the query plans become really complicated when `P.derivedRoles` is expanded to a complex expression?
- We won't be able to produce traces for derived roles activations to help debug things
- I am generally not fond of macro-like behaviour because they complicate things and make it really hard to reason about the system

I think it's OK to only have the effective derived roles available to use. Maybe we make it available under `P.runtime.effectiveDerivedRoles` to make it abundantly clear? There are couple of workarounds the users could do if they want to check for the existence of a particular set of derived roles in a policy rule.
- Define the derived role conditions as variables and reference those variables instead. Now that we have export variables, both the derived role definition and the policy rules can share them.  
- Create a dummy rule in the policy that references all the derived roles they are interested in.

Regarding sorting, I think it might be better to introduce a `sort` function instead. Then those who need stable output can use that if necessary.

haines:
> I think it's OK to only have the effective derived roles available to use. Maybe we make it available under `P.runtime.effectiveDerivedRoles` to make it abundantly clear? There are couple of workarounds the users could do if they want to check for the existence of a particular set of derived roles in a policy rule.

Yep, on balance I agree that turning it into a macro is overly complex.

I think we should introduce a new top-level identifier, `runtime`, so that this could just be `runtime.effectiveDerivedRoles`. I feel like adding `derivedRoles` to `request.principal` was potentially confusing anyway, since `request` is supposed to contain information that was passed in the check/plan request.

> Regarding sorting, I think it might be better to introduce a `sort` function instead. Then those who need stable output can use that if necessary.

I think introducing a `sort` function is potentially quite complicated. We'd have to have overloads for the different types one might want to sort.

How about we keep it sorted but we use [lazy binding](https://pkg.go.dev/github.com/google/cel-go@v0.18.0/interpreter#NewActivation) so that we only ever compute the `runtime.effectiveDerivedRoles` list if it is actually used in an expression? That would avoid an unnecessary map-to-slice conversion as well in the majority of cases where it is unused.

charithe:
Sounds good on both points (top level `runtime` and lazy binding) 👍🏽

charithe (approved):
LGTM 👍🏽

internal/conditions/cel.go
```diff
@@ -23,44 +22,50 @@ const (
...
-	decls.NewVar(CELPrincipalAbbrev, decls.NewObjectType("cerbos.engine.v1.Principal")),
-	decls.NewVar(CELResourceAbbrev, decls.NewObjectType("cerbos.engine.v1.Resource")),
-	decls.NewVar(CELVariablesIdent, decls.NewMapType(decls.String, decls.Dyn)),
-	decls.NewVar(CELVariablesAbbrev, decls.NewMapType(decls.String, decls.Dyn)),
-	decls.NewVar(CELGlobalsIdent, decls.NewMapType(decls.String, decls.Dyn)),
-	decls.NewVar(CELGlobalsAbbrev, decls.NewMapType(decls.String, decls.Dyn)),
-}
+	StdEnvOptions = []cel.EnvOption{
+		cel.CrossTypeNumericComparisons(true),
+		cel.Types(&enginev1.Request{}, &enginev1.Request_Principal{}, &enginev1.Request_Resource{}),
```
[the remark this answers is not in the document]
haines: Yep, I overlooked that!

charithe (approved):

PR #1800, opened 2023-09-12
enhancement: Migrate to protovalidate

### LINKED ISSUE #1785: Migrate to protovalidate
https://github.com/bufbuild/protovalidate/tree/main/tools/protovalidate-migrate

oguzhand95:
resolves #1785
depends on cerbos/cerbos-sdk-go#4


api/public/cerbos/engine/v1/engine.proto
```diff
@@ -192,22 +191,20 @@ message Resource {
...
     }
   ];
-  map<string, google.protobuf.Value> attr = 4 [
-    (validate.rules).map.no_sparse = true,
-    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
-      description: "Kay-value pairs of contextual data about this resource that should be used during policy evaluation."
-      example: "{\"owner\": \"bugs_bunny\"}"
-    }
-  ];
+  map<string, google.protobuf.Value> attr = 4 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
```
charithe, line 202: I think we can add some constraints like:

```
[(buf.validate.field).map.keys = { string: {min_len: 1} }],
[(buf.validate.field).map.values.required = true]
```

api/public/cerbos/engine/v1/engine.proto
```diff
@@ -254,15 +253,12 @@ message Principal {
...
     }
   ];
-  map<string, google.protobuf.Value> attr = 4 [
-    (validate.rules).map.no_sparse = true,
-    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
-      description: "Key-value pairs of contextual data about this principal that should be used during policy evaluation."
-      example: "{\"beta_tester\": true}"
-    }
-  ];
+  map<string, google.protobuf.Value> attr = 4 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
```
charithe, line 256: Same here as above

internal/validator/validator.go
```diff
@@ -0,0 +1,40 @@
...
+		protovalidate.WithMessages(
+			&requestv1.CheckResourcesRequest{},
+			&requestv1.PlanResourcesRequest{}),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create validator: %w", err)
+	}
+
+	return v, nil
+}
```
charithe, line 36: Why do we need a separate `Init` function? When the `validator` package is accessed, Go guarantees that the `init` function would have run.

charithe (approved):
LGTM

PR #1804, opened 2023-09-14
Use toolchain to manage Go version used

rcrowe:
By specifying `go 1.21.1` in the module any consumers of this module will be forced to use `1.21.1` as a minimum version. 

Now that the generated protobuf contracts are their own module, trying to depend on them forced us to use `1.21.1`.

This change introduces `toolchain` which means that a developer or the build of the main Cerbos project will automatically download this version to use (it's a module), while allowing anyone with `>=1.21.0` to still use the module.

#### Description

Fixes #

#### Checklist 


charithe:
Thanks! Will rebase and merge this after #1802.

haines:
I think maybe we should support all supported Go versions (i.e. 1.20+), at least for the api module and client SDK (once the server is separate then I think it's probably ok to only support 1.21 for that since it's not really intended for consumption as a library).

If so, (not for this PR) we should add a CI job to check that we can compile with 1.20 so that we don't inadvertently introduce something only available in 1.21.

charithe:
Make sense. @rcrowe do you mind changing the `go` directive to `1.20.0`?

charithe (approved):

PR #1815, opened 2023-09-22
enhancement: Deprecate client package

charithe:
Deprecate the `client` package and replace all uses of it with the new Go SDK available at github.com/cerbos/cerbos-sdk-go


Sambigeara (approved):
Very satisfying!

PR #1843, opened 2023-10-18
enhancement: Expose private Check API

Sambigeara:
Introduce an additional public API to the engine Check functionality.

Additionally, the `compile.Files` API returns an optional index which can be passed to `verify.Files` to prevent the need to build the index twice.

Lastly, an API is exposed which allows the retrieval of TestFixtures from a path within a given FS implementation.


oguzhand95 (commented):
LGTM

oguzhand95 (approved):

PR #1854, opened 2023-10-30
chore: Expose discovered test fixture paths

Sambigeara:
Test fixtures now carry info on the discovered path where the fixture file resides. The initial use case for this is to return the filenames from a new method in the private `TestFixtureGetter` API.

### DIFF
--- a/private/check/check.go
+++ b/private/check/check.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"io/fs"
+	"sync"
 
 	enginev1 "github.com/cerbos/cerbos/api/genpb/cerbos/engine/v1"
 	internalcompile "github.com/cerbos/cerbos/internal/compile"
@@ -20,16 +21,21 @@ import (
 type TestFixtureGetter struct {
 	fsys  fs.FS
 	cache map[string]*verify.TestFixture
+	mut   sync.RWMutex
 }
 
 func NewTestFixtureGetter(fsys fs.FS) *TestFixtureGetter {
 	return &TestFixtureGetter{
 		fsys:  fsys,
 		cache: make(map[string]*verify.TestFixture),
+		mut:   sync.RWMutex{},
 	}
 }
 
-func (g *TestFixtureGetter) LoadTestFixture(path string) (*verify.TestFixture, error) {
+func (g *TestFixtureGetter) Load(path string) (*verify.TestFixture, error) {
+	g.mut.Lock()
+	defer g.mut.Unlock()
+
 	fixture, ok := g.cache[path]
 	if !ok {
 		var err error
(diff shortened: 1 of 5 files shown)


private/check/check.go
```diff
@@ -20,16 +21,21 @@ import (
...
 		fsys:  fsys,
 		cache: make(map[string]*verify.TestFixture),
+		mut:   sync.RWMutex{},
 	}
 }
 
-func (g *TestFixtureGetter) LoadTestFixture(path string) (*verify.TestFixture, error) {
+func (g *TestFixtureGetter) Load(path string) (*verify.TestFixture, error) {
+	g.mut.Lock()
+	defer g.mut.Unlock()
```
charithe, line 37: Shouldn't we:
1. Acquire a read lock first to check whether the item exists in the cache
2. Acquire a write lock, check  whether the item exists again
3. Load the item and release the write lock

private/check/check.go
```diff
@@ -40,9 +46,33 @@ func (g *TestFixtureGetter) LoadTestFixture(path string) (*verify.TestFixture, e
...
+
 	return fixture, nil
 }
 
+type TestFixtureCtx struct {
+	Fixture *verify.TestFixture
+	Path    string
+}
+
+func (g *TestFixtureGetter) LoadAll() <-chan *TestFixtureCtx {
```
charithe, line 58: `LoadAll` sounds very similar to `Load` but it's doing something quite different. Shall we call it `GetAll` or something along those lines?

Also, since the hard work has already been done to populate the cache, is there any advantage to using a channel and a goroutine here? Can't we simply iterate over the cache, build an array of `TestFixtureCtx` and return it?

private/check/check.go
Sambigeara: Ha, I thought that was overkill for this scenario. But it's certainly more optimal. Will update

private/check/check.go
Sambigeara: Happy to return a slice over an array. I figured this might be slightly more favourable in that the memory requirements might be less than a full array. But maybe there are other overheads I'm not familiar with.

private/check/check.go
charithe: Is it likely to be accessed by multiple goroutines anyway? I can't imagine that being the case and we could potentially get rid of the mutex completely.

Sambigeara: No, you're right. In the `spitfire` usage the writes and reads happen sequentially. I'll do away with the mutex

private/check/check.go
charithe: You have a point there. If the other end simply loops over the channel and builds up a collection of fixtures, then that memory is going to be used anyway (along with at least 2kb for the goroutine stack + synchronization overheads). If it's going to work on each item as it arrives over the channel and moves on to the next, then that might be more memory efficient _if_ there are many fixtures. For a small number of fixtures, my hunch is that goroutine scheduling overhead dominates everything. Hard to say definitely without some benchmarks though.

Sambigeara: This prompted me to do some reading too (for my own curiosity) and I realised the same conclusion! Array makes loads more sense, about to push this and other changes for the comments.

charithe (approved):
LGTM

PR #1859, opened 2023-10-31
chore: Allow LoadTestFixture to continue on error

Sambigeara:
Passing `continueOnError == true` to `LoadTestFixture` will finish attempting to load all principals, resources and auxData. Any errors that occur will be stored on the type for processing later. Passing `false` will maintain current behaviour.

### DIFF
--- a/internal/verify/test_fixture.go
+++ b/internal/verify/test_fixture.go
@@ -26,21 +26,21 @@ import (
 )
 
 type Principals struct {
+	LoadError error
 	Fixtures  map[string]*enginev1.Principal
 	FilePath  string
-	LoadError error
 }
 
 type Resources struct {
+	LoadError error
 	Fixtures  map[string]*enginev1.Resource
 	FilePath  string
-	LoadError error
 }
 
 type AuxData struct {
+	LoadError error
 	Fixtures  map[string]*enginev1.AuxData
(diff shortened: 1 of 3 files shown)


internal/verify/test_fixture.go
```diff
@@ -82,15 +85,18 @@ func loadResources(fsys fs.FS, path string) (*Resources, error) {
...
 	}
 
+	resources := &Resources{
+		FilePath: fp,
+	}
+
 	pb := &policyv1.TestFixture_Resources{}
 	if err := loadFixtureElement(fsys, fp, pb); err != nil {
-		return nil, err
+		resources.LoadError = err
```
charithe, line 94: I'd probably accumulate the errors as follows:

```go
resources.LoadError = multierr.Append(resources.LoadError, err)
```

charithe (approved):
Ideally all the errors get accumulated and not just the last one but it can be fixed later.

internal/verify/test_fixture.go
Sambigeara: I did that and then removed it because it's not used down the chain at the mo. #IndecisiveSam 😫

charithe: No worries. It can always be added later if needed.

PR #1866, opened 2023-11-06
chore: Add pre-cache API to TestFixtureGetter

Sambigeara:
Adds an API which traverses the fs.FS and loads the test fixture for any non-nested directory called `testdata`

### DIFF
--- a/private/check/check.go
+++ b/private/check/check.go
@@ -13,6 +13,7 @@ import (
 	"github.com/cerbos/cerbos/internal/schema"
 	"github.com/cerbos/cerbos/internal/storage/disk"
 	"github.com/cerbos/cerbos/internal/storage/index"
+	"github.com/cerbos/cerbos/internal/util"
 	"github.com/cerbos/cerbos/internal/verify"
 )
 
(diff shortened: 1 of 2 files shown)


private/check/check.go
```diff
@@ -28,6 +32,27 @@ func NewTestFixtureGetter(fsys fs.FS) *TestFixtureGetter {
...
+		}
+
+		if !d.IsDir() {
+			return nil
+		}
+
+		// We need to search the filesystem for any directory `**/testdata`, omitting nested matches, e.g. `**/testdata/**/testdata`
+		if d.Name() == util.TestDataDirectory && !strings.Contains(filepath.Dir(path), fmt.Sprintf("/%s/", util.TestDataDirectory)) {
+			g.LoadTestFixture(path)
+		}
```
haines, line 48: Rather than checking for ancestor `testdata`s, we could just not enter the directory:

```suggestion
		if d.Name() == util.TestDataDirectory {
			g.LoadTestFixture(path)
			return fs.SkipDir
		}
```

Sambigeara: Much nicer, thanks

haines (approved):

PR #1868, opened 2023-11-09
enhancement: Support globals in policy tests

charithe:
- Allow users to define `globals` for each test or the whole test suite
- Add extra checks to make sure that test suites don't contain invalid expectations


internal/test/testdata/verify/test_fixture_get_tests/error_expectation_actions_not_in_input.yaml
```diff
@@ -13,11 +13,13 @@ table:
...
+    - principal: employee
       resource: employee_leave_request
       actions:
         view: EFFECT_ALLOW
         approve: EFFECT_ALLOW
+        delete: EFFECT_DENY
+        reject: EFFECT_DENY
 
 wantErr: |-
-  invalid test "Expectation not in input": found an expectation for principal "department_head" and resource "employee_leave_request", but at least one of these is not present in input
```
haines, line 23: The code path that produces this message still exists... is it dead code now? Or should this test case still be valid and the check be moved into the `buildExpectationLookup` function?

charithe: I deleted it by accident. Good catch!

haines (approved):

PR #1877, opened 2023-11-13
chore: Add tests to check fixture loading from testdata

charithe:
If the fixture files were not named
`testdata/{auxdata,principals,resources}.yaml` Cerbos versions up to
0.31 panicked. That's been inadvertently fixed in `main` but it needs
  tests to make sure there are no regressions.


dbuduev (approved):
LGTM

PR #1882, opened 2023-11-15
fix: Ignore empty files in policy repository

### LINKED ISSUE #1880: Repeated compilation errors
If you attempt to compile an empty file, this is what you get

```
file is not valid: [/: expected object, but got null, /: expected object, but got null, /: expected object, but got null, /: expected object, but got null, /: expected object, but got null]
```

The `/: expected object, but got null` is repeated 4 times for some reason 🤔

### LINKED ISSUE #1883: Inconsistent YAML parsing between schema validation and policy loading
We use github.com/ghodss/yaml (which uses gopkg.in/yaml.v2 under the hood) to convert YAML to JSON for policy parsing but gopkg.in/yaml.v3 to parse YAML for schema validation.

That means that the following policy passes schema validation but then cannot be loaded:

```yaml
apiVersion: api.cerbos.dev/v1
resourcePolicy:
  resource: foo
  version: yes
```

The error should be `file is not valid: [/resourcePolicy/version: expected string, but got boolean]`, but instead we get `failed to unmarshal JSON: proto: (line 1:80): invalid value for string type: true`.

We should use the same YAML parser in both places.

charithe:
- Ignore empty files or files with commented out contents when building the index.
- Use the same YAML library for reading policies and validating them

Fixes #1880
Fixes #1883

### DIFF
--- a/internal/jsonschema/jsonschema.go
+++ b/internal/jsonschema/jsonschema.go
@@ -13,8 +13,8 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/ghodss/yaml"
 	"github.com/santhosh-tekuri/jsonschema/v5"
-	"gopkg.in/yaml.v3"
 
 	"github.com/cerbos/cerbos/schema"
 )
--- a/internal/jsonschema/jsonschema.go
+++ b/internal/jsonschema/jsonschema.go
@@ -74,68 +74,56 @@ func validate(s *jsonschema.Schema, fsys fs.FS, path string) error {
 			return fmt.Errorf("unable to validate file %s: %w", path, err)
 		}
 
-		return newValidationErrorList(validationErr)
+		return newValidationError(validationErr)
 	}
 
 	return nil
 }
 
-func newValidationError(err *jsonschema.ValidationError) validationError {
-	path := "/"
-	if err.InstanceLocation != "" {
-		path = err.InstanceLocation
-	}
-	return validationError{
-		Path:    path,
-		Message: err.Message,
-	}
-}
+func newValidationError(err *jsonschema.ValidationError) error {
+	toVisit := err.Causes
+	var leaves []validationError
+	for ; len(toVisit) > 0; toVisit = toVisit[1:] {
+		ve := toVisit[0]
+		if ve == nil {
+			continue
+		}
 
-func newValidationErrorList(validationErr *jsonschema.ValidationError) validationErrorList {
-	if validationErr == nil {
-		return nil
-	}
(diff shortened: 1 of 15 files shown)


haines (approved):

haines:
Actually, I think ignoring empty files is insufficient; I think we need to de-dupe the validation errors. For example, this YAML file

```yaml
wat
```

produces this error

```
file is not valid: [/: expected object, but got string, /: expected object, but got string, /: expected object, but got string, /: expected object, but got string, /: expected object, but got string]
```

internal/jsonschema/jsonschema.go
```diff
@@ -74,68 +74,56 @@ func validate(s *jsonschema.Schema, fsys fs.FS, path string) error {
...
-
-type validationError struct {
-	Path    string
-	Message string
-}
-
-func (e validationError) Error() string {
-	return fmt.Sprintf("%s: %s", e.Path, e.Message)
-}
+	// sort the leaves by path and message, group by path and take the first message for each path to produce a stable set of errors
```
haines, line 104: I think we should include all unique messages for each path rather than just the first. If there are multiple properties missing, for example, it would be nicer to be able to see that instead of playing whack-a-mole by adding the missing one we have reported and then getting a new error next time you compile.

charithe: That's what I tried to do originally but, unfortunately, the error messages are not very good. You get whoppers like `additionalProperties 'key' not allowed, missing properties: 'apiVersion', missing properties: 'derivedRoles', missing properties: 'exportVariables', missing properties: 'principalPolicy', missing properties: 'resourcePolicy'`. 

I think it'd be quite rare to have multiple errors for a given path. We still report the errors for all  distinct paths so the user has a chance to fix all the errors at once in most cases anyway.

haines: It's pretty common in the playground to have multiple missing properties as you work on a policy.

Also, I think this example shows an issue: https://github.com/cerbos/cerbos/pull/1882/files#diff-1f26e131097220eb6929f987c1065dbae423d87e53ca85d4ac4e5bfb7e67bbc3

The message would be `missing properties: 'derivedRoles', missing properties: 'exportVariables', missing properties: 'principalPolicy', missing properties: 'resourcePolicy'` but we chop it down to `missing properties: 'derivedRoles'` which is sort of untrue - you could add any of the four types... and you definitely can get into this state easily in the playground because the autocompleter will happily add the `apiVersion` for you but doesn't then suggest any of the `oneof` options.

haines (approved):

PR #1902, opened 2023-12-07
chore: Update buf modules

charithe:
Update the buf module references


charithe:
Test failure is due to a bug(??? works-on-my-machine ™️ ) in the `dev` image which has been fixed in this PR but it won't take effect until this PR is merged 😵‍💫

PR #1921, opened 2023-12-25
fix(planner): Query planner must (pre)evaluate expressions with resource kind

dbuduev:
Resource kind is a known variable and MUST be considered while evaluating expressions:

```yaml
apiVersion: api.cerbos.dev/v1
derivedRoles:
  name: org_admin_role
  definitions:
    - name: org_admin
      parentRoles:
        - USER_ORG
      condition:
        match:
          any:
            of:
              - expr: R.kind == "organization" && R.attr.team == "RED"
              - expr: request.resource.kind != "organization" && R.attr.team == "BLUE"
```

The PR fixes this.

### DIFF
--- a/internal/conditions/cel.go
+++ b/internal/conditions/cel.go
@@ -17,17 +17,18 @@ import (
 )
 
 const (
-	CELRequestIdent    = "request"
-	CELResourceAbbrev  = "R"
-	CELResourceField   = "resource"
-	CELPrincipalAbbrev = "P"
-	CELPrincipalField  = "principal"
-	CELRuntimeIdent    = "runtime"
-	CELVariablesIdent  = "variables"
-	CELVariablesAbbrev = "V"
-	CELGlobalsIdent    = "globals"
-	CELGlobalsAbbrev   = "G"
-	CELAttrField       = "attr"
+	CELRequestIdent      = "request"
+	CELResourceAbbrev    = "R"
+	CELResourceKindField = "kind"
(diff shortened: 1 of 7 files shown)


internal/conditions/cel.go
```diff
@@ -17,17 +17,18 @@ import (
...
-	CELPrincipalField  = "principal"
-	CELRuntimeIdent    = "runtime"
-	CELVariablesIdent  = "variables"
-	CELVariablesAbbrev = "V"
-	CELGlobalsIdent    = "globals"
-	CELGlobalsAbbrev   = "G"
-	CELAttrField       = "attr"
+	CELRequestIdent      = "request"
+	CELResourceAbbrev    = "R"
+	CELResourceKindField = "kind"
```
charithe, line 22: I think the `id` field is accessible in conditions too. Is that handled as well? https://github.com/cerbos/cerbos/blob/002b035da064ea5b6a78d8f86552a9e71683fb62/api/public/cerbos/engine/v1/engine.proto#L357-L360

charithe (approved):

internal/conditions/cel.go
dbuduev: The planner uses another `Resource` type without ID: https://github.com/cerbos/cerbos/blob/002b035da064ea5b6a78d8f86552a9e71683fb62/api/public/cerbos/engine/v1/engine.proto#L20-L60

PR #1932, opened 2024-01-04
feat: Produce output if condition fails

### LINKED ISSUE #1920: Ability to produce output when the rule condition is not satisfied
Currently the output block is only evaluated if the rule is actually activated (action, roles and conditions are satisfied). In certain situations, it's desirable to produce output when the rule is nearly activated (action and roles match but the condition is not satisfied).

In order to maintain backward compatibility, reduce noise, and to keep policy execution as fast as possible (outputs incur a tiny overhead), the proposal is to let users add an optional `when` section to the `output` block to opt into this behaviour.

```yaml
    - actions: ['view']
      effect: EFFECT_ALLOW
      roles: ['user']
      condition:
        match:
          expr: timestamp(R.expiry_date) > now()
      output: 
        expr: >
          format("%d hours until expiry", (timestamp(R.expiry_date) - now()).getHours())
        when:
          cond_fail: >
            format("expired on %s", R.expiry_date)
``` 

When evaluating the above rule,
- if the action, roles and condition match, output will be the result of evaluating `output.expr`
- If the condition is not satisfied, output will be the result of evaluating `output. when.cond_fail` if it exists. Otherwise no output will be produced.

oguzhand95:
resolves #1920 

This PR introduces `output.when.{ruleActivated,ruleNotActivated}` for the principal and resource policies.

The `output.expr` field is deprecated in exchange for the `output.when.ruleActivated`. For backward compatibility reasons, `output.expr` will still work but overridden by the value of the `output.when.ruleActivated` field in case it is provided. 

```yaml
...
output:
  when:
    ruleActivated: |-
      "rule_activated"
    ruleNotActivated: |-
      "rule_not_activated"
```


api/private/cerbos/runtime/v1/runtime.proto
```diff
@@ -51,7 +51,8 @@ message RunnableResourcePolicySet {
       map<string, google.protobuf.Empty> roles = 4;
       Condition condition = 5;
       cerbos.effect.v1.Effect effect = 6;
-      Expr output = 7;
+      Expr output = 7 [deprecated = true];
+      Output output_info = 8;
```
charithe, line 55: `output_conf` might be better. Technically, we can rename the old field (e.g. `output` -> `output_old`) and name this one `output`. It'd still be backward compatible because the field number hasn't changed and we don't use the JSON or text representations of this proto anywhere. However that'd probably cause some confusion so probably not worth it.

docs/modules/policies/pages/outputs.adoc
```diff
@@ -2,11 +2,11 @@ include::ROOT:partial$attributes.adoc[]
 
 = Outputs
 
-You can define an optional expression to be evaluated when a policy rule is activated. The collected outputs from all the activated rules are included in the Cerbos API response.
+You can define an optional expression to be evaluated when a policy rule is activated or not activated.
```
charithe, line 5:
```suggestion
You can define an optional expression to be evaluated when a policy rule is fully activated (`action`, `roles` and `derivedRoles` match and `condition` is satisfied) or partially activated (`condition` is not satisfied).
```

api/private/cerbos/runtime/v1/runtime.proto
```diff
@@ -135,6 +137,15 @@ message Expr {
...
 
+message Output {
+  message On {
+    Expr condition_not_met = 1;
+  }
+
+  Expr expr = 1;
+  On on = 2;
+}
+
```
charithe, line 148:
```suggestion
message Output {
  message On {
    Expr condition_met = 1;
    Expr condition_not_met = 2;
  }

  On on = 1;
}

```

Ideally this is how we'd have represented it in the policy if it wasn't for the backward compatibility problem.

api/public/cerbos/policy/v1/policy.proto
```diff
@@ -230,7 +230,12 @@ message Match {
 }
 
 message Output {
+  message On {
+    string condition_not_met = 1;
+  }
+
   string expr = 1;
+  On on = 2;
```
charithe, line 238: Let's structure this like the following. To keep backward compatibility, we can copy the value of `expr` to `on.condition_met`. Everywhere else in the code we should just use `on.condition_met`. We should also update the documentation to show the new way of doing it.

```suggestion
  message On {
    string condition_met = 1;
    string condition_not_met = 2;
  }

  string expr = 1 [deprecated = true];
  On on = 2;
```

api/private/cerbos/runtime/v1/runtime.proto
oguzhand95: Since there is no problem with the backward compatibility, I think we could rename the old field to the `output_old`. It will look a bit ugly with the `output_info` or `output_conf` name 😄 (well, no one sees it but you know...). WDYT?

charithe: We'll have to test it to make sure that renaming the field doesn't break anything. I know it's technically possible but in practice there might be some corner cases 🤷🏽 

Maybe instead of `output_conf` we can call it `emit_output`. I think it works with the rest of the field names (`emitOutput.on.conditionNotMet`).

api/public/cerbos/policy/v1/policy.proto
```diff
@@ -230,7 +230,13 @@ message Match {
 }
 
 message Output {
-  string expr = 1;
+  message When {
+    string rule_activated = 1;
+    string rule_not_activated = 2;
```
haines, line 235: I think we should go for `rule_activated` and `condition_not_met`. The desired behaviour is for the `condition_not_met` output to be evaluated if everything except the condition matched, not if the rule wasn't activated because it is for a different role, but `rule_not_activated` would imply that the output would be produced even if the roles didn't match.

internal/engine/evaluator.go
```diff
@@ -235,16 +235,30 @@ func (rpe *resourcePolicyEvaluator) Evaluate(ctx context.Context, tctx tracer.Co
...
+				if ruleActivated && rule.EmitOutput.When.RuleActivated != nil {
+					result.Outputs = append(result.Outputs, &enginev1.OutputEntry{
+						Src: namer.RuleFQN(rpe.policy.Meta, p.Scope, rule.Name),
+						Val: evalCtx.evaluateProtobufValueCELExpr(rule.EmitOutput.When.RuleActivated.Checked, variables),
+					})
+				} else if !ruleActivated && rule.EmitOutput.When.ConditionNotMet != nil {
+					result.Outputs = append(result.Outputs, &enginev1.OutputEntry{
+						Src: namer.RuleFQN(rpe.policy.Meta, p.Scope, rule.Name),
+						Val: evalCtx.evaluateProtobufValueCELExpr(rule.EmitOutput.When.ConditionNotMet.Checked, variables),
+					})
```
charithe, line 260: I don't think you can rely on `ruleActivated` being `false` here. It can be `false` if there are no matching actions as well. Therefore, the `conditionNotMet` output has to be moved up to the loop where we determine that the condition is satisfied or not.

Also, these `if` statements can be simplified a bit for the `RuleActivated` output case. Something like:

```go
if ruleActivated {
  var outputExpr *exprpb.Expr
  switch {
     case rule.Output != nil:
        outputExpr = rule.Output.Checked
     case rule.EmitOutput != nil && rule.EmitOuput.When != nil && rule.EmitOuput.When.RuleActivated != nil:
        outputExpr = rule.EmitOutput.When.RuleActivated.Checked
    }
    
    if outputExpr != nil {
       var output *enginev1.OutputEntry
	   octx := rctx.StartOutput(rule.Name)
	   result.Outputs = append(result.Outputs, &enginev1.OutputEntry{
					Src: namer.RuleFQN(rpe.policy.Meta, p.Scope, rule.Name),
					Val: evalCtx.evaluateProtobufValueCELExpr(outputExpr, variables), //nolint:staticcheck
				})
				octx.ComputedOutput(output)
    }
} 
```

internal/engine/evaluator.go
```diff
@@ -314,16 +328,30 @@ func (ppe *principalPolicyEvaluator) Evaluate(ctx context.Context, tctx tracer.C
...
+					if ruleActivated && rule.EmitOutput.When.RuleActivated != nil {
+						result.Outputs = append(result.Outputs, &enginev1.OutputEntry{
+							Src: namer.RuleFQN(ppe.policy.Meta, p.Scope, rule.Name),
+							Val: evalCtx.evaluateProtobufValueCELExpr(rule.EmitOutput.When.RuleActivated.Checked, variables),
+						})
+					} else if !ruleActivated && rule.EmitOutput.When.ConditionNotMet != nil {
+						result.Outputs = append(result.Outputs, &enginev1.OutputEntry{
+							Src: namer.RuleFQN(ppe.policy.Meta, p.Scope, rule.Name),
+							Val: evalCtx.evaluateProtobufValueCELExpr(rule.EmitOutput.When.ConditionNotMet.Checked, variables),
+						})
```
charithe, line 353: Same as above

charithe (commented):
Let's add test cases for the following as well:
- Both `Output.Expr` and `Output.When` are defined
- Only `ConditionNotMet` is defined 
- Both `ConditionNotMet` and `RuleActivated` outputs defined for a rule that never activates or has condition satisfied

charithe (approved):

PR #1939, opened 2024-01-09
chore: YAML to Protobuf parser 

charithe:
Directly convert YAML To protobuf representation without converting to JSON first. This allows us to maintain source location information and handle common error cases.


internal/protoyaml/protoyaml.go
```diff
@@ -0,0 +1,907 @@
...
+		}
+
+		outMsg = append(outMsg, msg)
+		outSrc = append(outSrc, srcCtx)
+	}
+
+	return outMsg, outSrc, outErr
+}
+
+func fixStringsStartingWithQuotes(tokens token.Tokens) token.Tokens {
```
haines, line 123: I'm not so keen on this. By doing this, we're then defining our own policy language, it's no longer standard YAML. The SDKs won't be able to parse the same files that Cerbos can because they use vanilla YAML parsers.

Given it is a common error, I think it definitely makes sense to detect this case and offer a more helpful error message suggesting the use of alternative string literal syntaxes, but I don't think we should go as far as actually accepting invalid YAML.

haines (approved):
LGTM so far!

This is a YAML 1.2 parser, which I think is the right move. However, it could be a breaking change in the (hopefully unlikely) event that users are relying on YAML 1.1 behaviours (e.g. `yes` etc for booleans).

I wonder if we could easily detect if this would have parsed the document differently to the previous parser, and print a warning? Or perhaps given that it is fairly edge-casey a note in the release notes will be sufficient.

We will need to update the SDKs that parse YAML to make sure they are doing so against the 1.2 spec.

Given that YAML is a superset of JSON, are you intending to use this for JSON policy files as well? I think it would be sensible so that we can unify the error handling. Perhaps to allow this we need an additional option to the parser `WithJSON` or similar, which forces it to only accept valid JSON syntax?

charithe:
> I wonder if we could easily detect if this would have parsed the document differently to the previous parser

I hadn't actually considered that. I'll take a look to figure out where parsing behaviour might differ and add some tests. Given that policies are pretty heavy on strings and don't really have boolean fields etc. I don't think there will be a massive issue.

> Given that YAML is a superset of JSON, are you intending to use this for JSON policy files as well? 

Ideally, yes. Although it looks like it's not very good at parsing inline JSON. I haven't given it a proper try yet. Maybe we'd need a different parser for JSON files. 

I just wanted to validate my initial approach and create a more digestible PR. Things will definitely change a lot in subsequent PRs as I discover the inevitable problems and edge cases 😄

PR #1942, opened 2024-01-10
docs: Remove older versions

charithe:
Remove the older versions of documentation because they just add to
build times. We've managed to maintain backward compatibility quite well
over the releases so referencing the newer docs should still work for
those who are absolutely stuck running very old versions of Cerbos.

Also adds a banner to docs to warn users that they are looking at an old
version.

### DIFF
--- a/docs/antora-playbook.yml
+++ b/docs/antora-playbook.yml
@@ -5,12 +5,12 @@ site:
   start_page: cerbos:ROOT:index.adoc
   robots: allow
   keys:
-    google_analytics: 'G-8G3G3MS838'
+    google_analytics: "G-8G3G3MS838"
 content:
   sources:
     - url: ./../
-      branches: [HEAD, 'v{0..9}*']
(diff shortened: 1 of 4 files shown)


docs/antora-playbook.yml
```diff
@@ -5,12 +5,12 @@ site:
...
   robots: allow
   keys:
-    google_analytics: 'G-8G3G3MS838'
+    google_analytics: "G-8G3G3MS838"
 content:
   sources:
     - url: ./../
-      branches: [HEAD, 'v{0..9}*']
-      edit_url: 'https://github.com/cerbos/cerbos/tree/main/{path}'
+      branches: [HEAD, "v{0..9}*", "!v0.{0..29}"]
```
alexolivier, line 12: For anyone with a deep link to pre-0.29 can we notify/redirect them?

alexolivier (approved):

PR #1960, opened 2024-01-25
feat: Better diagnostic error messages for policy issues

charithe:
Produce better error messages for parsing and validation problems found in policy files. Includes accurate line and column numbers and surrounding context where possible.


charithe:
The protobuf check failure is not relevant because the changed protobuf is not yet used anywhere else.

cmd/cerbos/compile/internal/lint/display.go
```diff
@@ -32,42 +32,54 @@ func displayJSON(p *printer.Printer, errs *index.BuildError, colorLevel outputco
...
+	if len(errs.DisabledDefs) > 0 {
+		p.Println(colored.Header("Disabled policies"))
+		for _, d := range errs.DisabledDefs {
+			p.Printf("%s %s\n", colored.Position(d.GetFile(), d.GetPosition()), colored.PolicyKey(d.GetPolicy()))
+		}
+		p.Println()
+	}
+
+	if len(errs.MissingScopes) > 0 {
+		p.Println(colored.Header("Missing Scopes"))
```
haines, line 44:
```suggestion
		p.Println(colored.Header("Missing scopes"))
```

cmd/cerbos/compile/internal/lint/display.go
```diff
@@ -32,42 +32,54 @@ func displayJSON(p *printer.Printer, errs *index.BuildError, colorLevel outputco
...
 		p.Println(colored.Header("Duplicate definitions"))
 		for _, dd := range errs.DuplicateDefs {
-			p.Printf("%s is a duplicate of %s\n", colored.FileName(dd.File), colored.FileName(dd.OtherFile))
+			p.Printf("%s duplicate definition of %s (previous definition in %s)\n", colored.Position(dd.GetFile(), dd.GetPosition()), colored.PolicyKey(dd.GetPolicy()), colored.FileName(dd.OtherFile))
 		}
 		p.Println()
 	}
 
 	if len(errs.MissingImports) > 0 {
 		p.Println(colored.Header("Missing Imports"))
```
haines, line 60:
```suggestion
		p.Println(colored.Header("Missing imports"))
```

internal/parser/parser_test.go
```diff
@@ -24,20 +25,20 @@ import (
 	policyv1 "github.com/cerbos/cerbos/api/genpb/cerbos/policy/v1"
 	privatev1 "github.com/cerbos/cerbos/api/genpb/cerbos/private/v1"
 	sourcev1 "github.com/cerbos/cerbos/api/genpb/cerbos/source/v1"
-	"github.com/cerbos/cerbos/internal/protoyaml"
+	"github.com/cerbos/cerbos/internal/namer"
+	parser1 "github.com/cerbos/cerbos/internal/parser"
```
haines, line 29: This is a slightly confusing alias... what about importing the YAML parser as `yamlparser` and letting this one just be `parser`?

internal/util/serdeio.go
```diff
@@ -35,6 +35,11 @@ func ReadJSONOrYAML(src io.Reader, dest proto.Message) error {
 	return d.decode(dest)
 }
 
+func IsJSON(src []byte) bool {
```
haines, line 38: If I'm being pedantic,

```yaml
{ foo: [bar, baz] }
```

is valid YAML but not valid JSON.

I think the parser should be aware if it is parsing YAML or JSON based on the file extension of the input, and valid YAML in a `.json` file should be rejected (presumably someone using `.json` policies is doing so because they want them to be easily machine-parseable, so it would be better to be strict).

internal/test/testdata/protoyaml/case_019.json.input
haines, line 1: 😆

haines (commented):
This is awesome 🌟

internal/parser/parser_test.go
charithe: Indeed. I renamed the package from `protoyaml` to `parser` at the last minute and didn't check too closely what it did to the source 🤦🏽

internal/util/serdeio.go
charithe: Good point. The problem is that the parser works off of an `io.Reader` because I wanted it to be generic and work with any byte stream. The file extension information gets lost because of that. Maybe the signature should be changed to let the caller define whether the stream is YAML or JSON 🤔

haines: > Maybe the signature should be changed to let the caller define whether the stream is YAML or JSON 🤔

Sounds good to me!

charithe: So, even if the caller provides a hint about the kind of bytes we are dealing with, the underlying parser doesn't provide any indication about whether the contents are YAML or JSON. Without that information, I don't think we can enforce that `.json` files contain only JSON and so forth.

The only reason for the `IsJSON` function here is because the detection of quoted strings in YAML gives false positives for single-line JSON content. Furthermore, in the current version of Cerbos, we do a [similar check and disregard the extension](https://github.com/cerbos/cerbos/blob/8db9898212a72a52c840a753addcec0635d5cbd0/internal/util/serdeio.go#L38-L48) so if we change that behaviour it would be a breaking change as well. Therefore, I am inclined to keep it as it is.

haines: Ok, fair enough!

haines (approved):

PR #1968, opened 2024-01-30
enhancement: Better compilation errors

charithe:
Follows #1960 

This PR adds location and context information to compilation errors. It also dedupes some errors to make the output less noisy. For instance, if there was an error in an `exportVariables` policy we used to report that error repeatedly for every policy that imported it.

| Before | After |
|--------|-------|
| image       | image |


internal/policy/protopath.go
```diff
@@ -0,0 +1,73 @@
...
+func ResourcePolicyResourceSchemaProtoPath() string {
+	return "resource_policy.schemas.resource_schema.ref"
+}
+
+func PrincipalPolicyRuleProtoPath(idx int) string {
+	return fmt.Sprintf("principal_policy.rules[%d]", idx)
+}
+
+func PrincipalPolicyActionRuleProtoPath(parentIdx, idx int) string {
+	return fmt.Sprintf("principal_policy.rules[%d].actions[%d]", parentIdx, idx)
```
Sambigeara, line 37:
```suggestion
	return fmt.Sprintf("%s.actions[%d]", PrincipalPolicyRuleProtoPath(parentIdx), idx)
```

Sambigeara (approved):
Wow! 👏

Sambigeara (approved):

PR #1977, opened 2024-02-06
feat: Add tests filtering

dbuduev:
Allow selecting test particular resources and principals. Private API only.

### DIFF
--- a/private/verify/verify.go
+++ b/private/verify/verify.go
@@ -101,8 +101,32 @@ type Checker interface {
 	Check(ctx context.Context, inputs []*enginev1.CheckInput, opts CheckOptions) ([]*enginev1.CheckOutput, error)
 }
 
-func WithCustomChecker(ctx context.Context, fsys fs.FS, eng Checker) (*policyv1.TestResults, error) {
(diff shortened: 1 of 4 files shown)


private/verify/verify.go
```diff
@@ -101,8 +101,32 @@ type Checker interface {
...
+}
+
+func WithPrincipalTestFilter(name string) Opt {
+	return func(config *verify.Config) {
+		if config.RunPrincipals == nil {
+			config.RunPrincipals = make(map[string]struct{})
+		}
+		config.RunPrincipals[name] = struct{}{}
+	}
+}
```
charithe, line 122:
```suggestion
func WithResourceTestFilter(names ...string) Opt {
	return func(config *verify.Config) {
		if config.RunResources == nil {
			config.RunResources = make(map[string]struct{}, len(names))
		}
		
		for _, n := range names {
		    config.RunResources[n] = struct{}{}
		}
	}
}

func WithPrincipalTestFilter(names ...string) Opt {
	return func(config *verify.Config) {
		if config.RunPrincipals == nil {
			config.RunPrincipals = make(map[string]struct{}, len(names))
		}
		
		for _, n := range names {
		    config.RunPrincipals[n] = struct{}{}
		}
	}
}
```

PR #1978, opened 2024-02-07
chore: Reduce Docker healthcheck interval

charithe:
In Docker Compose, the helathcheck interval can add an artificial delay
because it waits for the first healthcheck to run.

### DIFF
--- a/Dockerfile.cerbos
+++ b/Dockerfile.cerbos
@@ -7,7 +7,7 @@ ENV CERBOS_CONFIG="__default__"
 VOLUME ["/policies", "/tmp", "/.cache"]
 ENTRYPOINT ["/cerbos"]
 CMD ["server"]
-HEALTHCHECK --interval=1m --timeout=3s CMD ["/cerbos", "healthcheck"]
+HEALTHCHECK --interval=10s --timeout=2s --retries=2 --start-period=5s CMD ["/cerbos", "healthcheck"]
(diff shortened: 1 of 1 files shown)


dbuduev (approved):

Dockerfile.cerbos
```diff
@@ -7,7 +7,7 @@ ENV CERBOS_CONFIG="__default__"
 VOLUME ["/policies", "/tmp", "/.cache"]
 ENTRYPOINT ["/cerbos"]
 CMD ["server"]
-HEALTHCHECK --interval=1m --timeout=3s CMD ["/cerbos", "healthcheck"]
+HEALTHCHECK --interval=10s --timeout=2s --retries=2 --start-period=5s CMD ["/cerbos", "healthcheck"]
```
haines, line 10: I don't think `start period < interval` makes sense?

>start period provides initialization time for containers that need time to bootstrap. Probe failure during that period will not be counted towards the maximum number of retries

If the interval is greater than the start period, then there won't be any probes to fail anyway.

I think we want to set something like `--start-interval=1s` as well so that we can detect when Cerbos is ready before the first regular probe.

>start interval is the time between health checks during the start period.

charithe: Good point about `--start-period`. The reason I didn't include `--start-interval` is because it was only introduced in Docker Engine 25.0 which was only just released. I think it'd take a little while for people to upgrade.

PR #1981, opened 2024-02-09
chore: Copy metadata to runtime policies

dbuduev:


charithe (commented):
Can't we use the top-level `metadata` field of `*PolicySet`? A policy set is usually just a single policy anyway (except for scoped policies)

dbuduev:
`*PolicySet` corresponds to several policies in the same scope hierarchy. These policies might have their own metadata.
Using single top-level metadata requires merging annotations from all these files. 
Is that what you mean?

PR #1988, opened 2024-02-15
chore: Move compiled policies annotations to *PolicySet

dbuduev:
There's no need for the policy set to contain annotations for the ancestor policies.


charithe (approved):
LGTM. It's not a breaking change since the last change hasn't been released yet.

PR #2035, opened 2024-03-07
chore(docs): Add redirects for image URLs

charithe:
Because Antora doesn't have support for Git LFS, we need to create
redirect rules to load the images correctly. This has to be done on a
version-by-version basis to avoid redirect loops.

### DIFF
--- a/hack/scripts/prep-release.sh
+++ b/hack/scripts/prep-release.sh
@@ -40,6 +40,29 @@ update_version() {
 	go run ./hack/tools/generate-npm-packages
 }
 
+add_redirects() {
+	local VER="$1"
+	local STANZA=$(
+		cat <<EOF
+
+[[redirects]]
+from = "/cerbos/$VER/_images/*"
+to = "/cerbos/prerelease/_images/:splat"
+status = 302
+force = true
+
+[[redirects]]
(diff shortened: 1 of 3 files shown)


hack/scripts/prep-release.sh
```diff
@@ -40,6 +40,29 @@ update_version() {
 	go run ./hack/tools/generate-npm-packages
 }
 
+add_redirects() {
+	local VER="$1"
+	local STANZA=$(
+		cat <<EOF
+
+[[redirects]]
+from = "/cerbos/$VER/_images/*"
```
haines, line 49: Won't this add a redirect for the to-be-released version number (which will actually be "latest")? We need it for the previous version.

Although, it might be fine to add the redirect preemptively - we'd just need to manually add 0.34 in now since the next one to be added will be 0.35

charithe: Yeah, the idea was to add the entry preemptively because the script doesn't have an easy way of determining what the soon-to-be-outdated version is.

I forgot to add 0.34 because in my mind we were still on 0.33 🤦🏽 Thanks for catching that.

haines (approved):

PR #2047, opened 2024-03-11
chore: Refactor Cerbos Hub configuration

charithe:
Move credentials and connection settings to their own configuration section under `hub` so that they can be shared.

### DIFF
--- a/docs/modules/configuration/partials/fullconfiguration.adoc
+++ b/docs/modules/configuration/partials/fullconfiguration.adoc
@@ -75,6 +75,16 @@ engine:
   globals: {"environment": "staging"} # Globals are environment-specific variables to be made available to policy conditions.
   lenientScopeSearch: false # LenientScopeSearch configures the engine to ignore missing scopes and search upwards through the scope tree until it finds a usable policy.
 hub:
+  connection: # Connection holds advanced connection settings for Cerbos Hub.
+    apiEndpoint: https://api.cerbos.cloud # Required. APIEndpoint is the address of the API server.
+    bootstrapEndpoint: https://cdn.cerbos.cloud # Required. BootstrapEndpoint is the addresses of the server serving the bootstrap configuration.
+    heartbeatInterval: 2m # HeartbeatInterval is the interval for sending regular heartbeats.
+    maxRetryWait: 120s # MaxRetryWait is the maximum amount of time to wait between retries.
+    minRetryWait: 1s # MinRetryWait is the minimum amount of time to wait between retries.
+    numRetries: 5 # NumRetries is the number of times to retry before giving up.
+    tls: # TLS defines settings for TLS connections.
+      authority: domain.tld # Authority overrides the Cerbos PDP server authority if it is different from what is provided in the address.
+      caCert: /path/to/CA_certificate # CACert is the path to the CA certificate chain to use for certificate verification.
   credentials: # Credentials holds Cerbos Hub client credentials.
     clientID: 92B0K05B6HOF # ClientID of the Cerbos Hub credential. Defaults to the value of the CERBOS_HUB_CLIENT_ID environment variable.
     clientSecret: ${CERBOS_HUB_CLIENT_SECRET} # ClientSecret of the Cerbos Hub credential. Defaults to the value of the CERBOS_HUB_CLIENT_SECRET environment variable.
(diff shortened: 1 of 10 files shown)


docs/modules/configuration/partials/fullconfiguration.adoc
```diff
@@ -74,6 +74,22 @@ engine:
   defaultPolicyVersion: "default" # DefaultPolicyVersion defines what version to assume if the request does not specify one.
   globals: {"environment": "staging"} # Globals are environment-specific variables to be made available to policy conditions.
   lenientScopeSearch: false # LenientScopeSearch configures the engine to ignore missing scopes and search upwards through the scope tree until it finds a usable policy.
+hub:
+  connection: # Connection holds advanced connection settings for Cerbos Hub.
```
haines, line 78: I feel like "advanced" implies optional and that the defaults should suffice for most users, but then the endpoints are required. Should we change that so that they have the suggested values as defaults?

internal/storage/bundle/conf.go
```diff
@@ -181,17 +87,28 @@ func (conf *Conf) Validate() (outErr error) {
...
+		return fmt.Errorf("failed to read Cerbos Hub configuration: %w", err)
 	}
 
-	return outErr
+	switch {
+	case hubConf.Credentials.IsUnset():
+		return conf.Credentials.Validate()
+	case conf.Credentials.IsUnset():
+		conf.Credentials = hubConf.Credentials
+		return conf.Credentials.Validate()
```
haines, line 108: Should we print a deprecation warning in this case to encourage people to migrate to the `hub` section?

haines (approved):

docs/modules/configuration/partials/fullconfiguration.adoc
charithe: Actually, this section is not meant to be configured by users at all. Should we just hide it?

haines: Yeah, I think it's reasonable for this section to be an undocumented escape hatch.

charithe:
I've refactored this again a little bit more since your last review @haines. For now, if there's an explicit configuration section under `bundle`, it'll take precedence and there won't be an error if there's a `hub` configuration section as well. This is because I ended up writing a lot of messy and complicated logic to juggle between the two and it was confusing even myself.

docs/modules/configuration/partials/fullconfiguration.adoc
haines: I think the `connection` section still needs an ignore tag on it.

internal/hub/conf.go
```diff
@@ -0,0 +1,208 @@
...
+	// Deprecated: Use PDPID
+	InstanceID string `yaml:"instanceID" conf:",ignore"`
+	// Deprecated: Use WorkspaceSecret
+	SecretKey string `yaml:"secretKey" conf:",ignore"`
+}
+
+func (cc *CredentialsConf) Validate() (outErr error) {
+	// SecretKey was renamed to WorkspaceSecret in Cerbos 0.31.0
+	if cc.WorkspaceSecret == "" && cc.SecretKey != "" {
+		cc.WorkspaceSecret = cc.SecretKey
```
haines, line 105: Could we log deprecation warnings here and below too?

internal/hub/conf.go
```diff
@@ -0,0 +1,208 @@
...
+func GetEnv(key EnvVarKey) string {
+	varNames, ok := envVars[key]
+	if !ok {
+		return ""
+	}
+
+	for _, v := range varNames {
+		val, ok := os.LookupEnv(v)
+		if ok {
+			return val
```
haines, line 43: Maybe we could also log a deprecation warning here if it is not the first key?

internal/hub/conf.go
```diff
@@ -0,0 +1,208 @@
...
+	NumRetries uint `yaml:"numRetries" conf:",example=5"`
+	// HeartbeatInterval is the interval for sending regular heartbeats.
+	HeartbeatInterval time.Duration `yaml:"heartbeatInterval" conf:",example=2m"`
+}
+
+func (cc ConnectionConf) IsUnset() bool {
+	return cc == ConnectionConf{}
+}
+
+func (cc *ConnectionConf) Validate() error {
```
Sambigeara, line 162: Should this be a `SetDefaults` call?

internal/hub/conf.go
```diff
@@ -0,0 +1,208 @@
...
+	case cc.HeartbeatInterval > 0 && cc.HeartbeatInterval < minHeartbeatInterval:
+		cc.HeartbeatInterval = minHeartbeatInterval
+	}
+
+	return nil
+}
+
+// TLSConf holds TLS configuration for the remote connection.
+type TLSConf struct {
+	// Authority overrides the Cerbos PDP server authority if it is different from what is provided in the address.
```
haines, line 197:
```suggestion
	// Authority overrides the Cerbos Hub server authority if it is different from what is provided in the API and bootstrap endpoints.
```

internal/hub/conf.go
```diff
@@ -0,0 +1,208 @@
...
+	ClientSecret string `yaml:"clientSecret" conf:",example=${CERBOS_HUB_CLIENT_SECRET}"`
+	// WorkspaceSecret used to decrypt the bundles. Defaults to the value of the CERBOS_HUB_WORKSPACE_SECRET environment variable.
+	WorkspaceSecret string `yaml:"workspaceSecret" conf:",example=${CERBOS_HUB_WORKSPACE_SECRET}"`
+	// Deprecated: Use PDPID
+	InstanceID string `yaml:"instanceID" conf:",ignore"`
+	// Deprecated: Use WorkspaceSecret
+	SecretKey string `yaml:"secretKey" conf:",ignore"`
+}
+
+func (cc *CredentialsConf) Validate() (outErr error) {
```
Sambigeara, line 102: Also here?

docs/modules/configuration/partials/fullconfiguration.adoc
charithe: Lost in the git shuffle 🤦🏽

internal/hub/conf.go
charithe: I am actually abusing the `Validate` method here. The reason it's not `SetDefaults` is because we want to prevent the user from setting illegal values and revert them to sensible values. That has to be done at the end of loading the config hence why it's in `Validate`

internal/hub/conf.go
charithe: Same as above, this swap around has to be done at the end of config loading.

haines (approved):

PR #2056, opened 2024-03-18
feat: Integrate `hub` audit backend with cloud api

Sambigeara:
This PR wires up the `IngestSyncer` with the new Cloud API client.

* I've separated the key scan into two concurrent streams (access + decision logs) to avoid fiddly logic to infer types from each key when retrieving the access log (retrieving the log requires reading sync keys, retrieving the actual log key, then retrieving the log value with that key).

* I've also reworked the `ForceWrite` logic in an attempt to fix the flakey tests 🤞.

* I've temporarily reverted the commit which renames `cerboshub` -> `hub` as it ruined the PR diff, I can add this back in afterwards.

### DIFF
--- a/internal/audit/cerboshub/cerboshub.go
+++ b/internal/audit/cerboshub/cerboshub.go
@@ -181,12 +181,7 @@ func schedule(db *badgerv4.DB, muTimer *mutexTimer, syncer IngestSyncer, minFlus
 	// (and therefore burdening the backend).
 	muTimer.set(minFlushInterval)
 
-	// TODO(saml) is the select necessary?
-	// https://github.com/cerbos/cerbos/pull/2056#discussion_r1528673965
-	select {
-	case waitCh <- struct{}{}:
-	default:
-	}
+	waitCh <- struct{}{}
 }
 
 func streamLogs(db *badgerv4.DB, syncer IngestSyncer, maxBatchSize, numGo int, flushTimeout time.Duration) error {
--- a/internal/audit/cerboshub/cerboshub.go
+++ b/internal/audit/cerboshub/cerboshub.go
@@ -1,7 +1,7 @@
 // Copyright 2021-2024 Zenauth Ltd.
 // SPDX-License-Identifier: Apache-2.0
 
-package hub
+package cerboshub
 
 import (
 	"context"
@@ -22,7 +22,7 @@ import (
 )
 
 const (
-	Backend = "hub"
+	Backend = "cerboshub"
 )
 
 var (
@@ -35,7 +35,7 @@ func init() {
 	audit.RegisterBackend(Backend, func(_ context.Context, confW *config.Wrapper, decisionFilter audit.DecisionLogEntryFilter) (audit.Log, error) {
 		conf := new(Conf)
 		if err := confW.GetSection(conf); err != nil {
-			return nil, fmt.Errorf("failed to read hub audit log configuration: %w", err)
+			return nil, fmt.Errorf("failed to read cerboshub audit log configuration: %w", err)
 		}
 
 		logger := zap.L().Named("auditlog").With(zap.String("backend", Backend))
--- a/internal/audit/cerboshub/cerboshub.go
+++ b/internal/audit/cerboshub/cerboshub.go
@@ -14,8 +14,10 @@ import (
 	"github.com/cerbos/cerbos/internal/audit"
 	"github.com/cerbos/cerbos/internal/audit/local"
 	"github.com/cerbos/cerbos/internal/config"
+	logsv1 "github.com/cerbos/cloud-api/genpb/cerbos/cloud/logs/v1"
 	badgerv4 "github.com/dgraph-io/badger/v4"
 	"github.com/dgraph-io/ristretto/z"
+	"github.com/sourcegraph/conc/pool"
 	"go.uber.org/zap"
 )
 
@@ -24,9 +26,9 @@ const (
 )
 
 var (
-	SyncStatusPrefix   = []byte("bs") // "b" for contiguity with audit log keys in LSM, "s" because "sync"
-	AccessSyncPrefix   = []byte("bsacc")
-	DecisionSyncPrefix = []byte("bsdec")
+	SyncStatusPrefix   = []byte("bs")   // "b" for contiguity with audit log keys in LSM, "s" because "sync"
+	AccessSyncPrefix   = []byte("bsac") // these need to be len(4) to correctly reuse `local.GenKey`
+	DecisionSyncPrefix = []byte("bsde")
 )
 
 func init() {
@@ -86,8 +88,10 @@ func (l *Log) WriteAccessLogEntry(ctx context.Context, record audit.AccessLogEnt
 	}
 
 	key := local.GenKey(AccessSyncPrefix, callID)
+	// TODO(saml) retrieve key generated in embedded call above to avoid recalc?
+	value := local.GenKey(local.AccessLogPrefix, callID)
 
-	return l.Write(ctx, key, nil)
+	return l.Write(ctx, key, value)
 }
 
 func (l *Log) WriteDecisionLogEntry(ctx context.Context, record audit.DecisionLogEntryMaker) error {
@@ -108,8 +112,10 @@ func (l *Log) WriteDecisionLogEntry(ctx context.Context, record audit.DecisionLo
 	}
 
 	key := local.GenKey(DecisionSyncPrefix, callID)
+	// TODO(saml) retrieve key generated in embedded call above to avoid recalc?
+	value := local.GenKey(local.DecisionLogPrefix, callID)
 
-	return l.Write(ctx, key, nil)
+	return l.Write(ctx, key, value)
 }
 
 type mutexTimer struct {
@@ -131,8 +137,6 @@ func (mt *mutexTimer) set(waitDuration time.Duration) {
 	if mt.t == nil {
 		mt.t = time.AfterFunc(waitDuration, func() {
 			mt.expireCh <- struct{}{}
-			// TODO(saml) could close/recreate a new channel each time if we want more than one listener
-			// close(mt.wait)
 
 			mt.mu.Lock()
 			mt.t = nil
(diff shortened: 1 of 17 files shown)


Sambigeara:
image

...

internal/audit/cerboshub/cerboshub.go
```diff
@@ -171,27 +181,52 @@ func schedule(db *badgerv4.DB, muTimer *mutexTimer, syncer IngestSyncer, minFlus
 	// This prevents a retry completion occurring immediately before the next sync
 	// (and therefore burdening the backend).
 	muTimer.set(minFlushInterval)
+
+	select {
+	case errCh <- err:
```
charithe, line 186: Won't this `err` be `nil` in most cases?

internal/audit/cerboshub/cerboshub.go
```diff
@@ -171,27 +181,52 @@ func schedule(db *badgerv4.DB, muTimer *mutexTimer, syncer IngestSyncer, minFlus
...
-
-	ctx, cancelFn := context.WithTimeout(context.Background(), flushTimeout)
-	defer cancelFn()
+	switch kind {
+	case logsv1.IngestBatch_ENTRY_KIND_ACCESS_LOG:
+		stream.Prefix = AccessSyncPrefix
+	case logsv1.IngestBatch_ENTRY_KIND_DECISION_LOG:
+		stream.Prefix = DecisionSyncPrefix
+	case logsv1.IngestBatch_ENTRY_KIND_UNSPECIFIED:
+		return errors.New("unspecified IngestBatch_EntryKind")
```
charithe, line 218: If we determine the `prefix` in `streamLogs` and pass it on as an argument to `streamPrefix`, this `switch` won't be needed on every batch.

internal/audit/cerboshub/cerboshub.go
```diff
@@ -171,27 +181,52 @@ func schedule(db *badgerv4.DB, muTimer *mutexTimer, syncer IngestSyncer, minFlus
...
 		if err != nil {
 			return err
 		}
 
-		keys := make([][]byte, len(kvList.Kv))
-		for i, kv := range kvList.Kv {
-			keys[i] = kv.Key
+		keys := make([][]byte, 0, len(kvList.Kv))
+		for _, kv := range kvList.Kv {
+			keys = append(keys, kv.Key)
```
charithe, line 229: It was faster to use indexing rather than `append` in a benchmark I did a few Go versions ago. Not sure whether that has changed now.

internal/audit/cerboshub/conf.go
```diff
@@ -30,54 +32,82 @@ var (
 )
 
 type Conf struct {
-	local.Conf
 	Ingest IngestConf `yaml:"ingest"`
+	// Local is configuration for the local Audit backend. See 'audit.local' section for documentation
+	Local local.Conf `yaml:"local" conf:",example=see audit.local section"`
```
charithe, line 37: Could we embed this struct or name it something other than `local`? I think it would be confusing for some users if we redirect them to the `local` config. Ideally, everything they need to configure the `hub` backend should be visible in a single documentation block. 

If `confdocs` is unable to handle embedded structs, we should fix that. @oguzhand95  can probably take a look at that in parallel.

internal/audit/cerboshub/cerboshub.go
Sambigeara: Yeah, I was on the edge of using a `chan struct{}` but figured err might be more idiomatic/future-proof. Perhaps pointless

internal/audit/cerboshub/cerboshub.go
Sambigeara: Ah. I assumed that pre-allocating a large enough capacity (can't determine exact length so used the sum of the two) would mean that `appends` would still just add to the end of the pre-existing array? To index, I'd need to set the length to (average-case) double the required number, and then the length would be all skewiff?

Sambigeara: Ohhh I see now. What I just described applied when I was still handling both prefixes in a single scan, I forgot to change in the update. Thanks for raising

internal/audit/cerboshub/conf.go
Sambigeara: Perhaps embedding would be the tidiest option. Seems like the existing config parsing logic doesn't take embedded structs into account, I can take a look at that.

Separately, when I tinkered with this a couple of weeks ago, I came across issues with `confdocs` generation as well. Seems like it gets a bit fiddly because in it's existing form, it only has access to the local struct. I think we'd need to somehow cache traversed nodes or pass state around so that `audit.(cerbos)hub` can access the struct metadata in `audit.local`. I [stubbed out the part where it hits embedded nodes](https://github.com/cerbos/cerbos/blob/f63a63d2a0194d731b29ae5dfdff416a7b5e1868/hack/tools/confdocs/confdocs.go#L243-L245) in the generation logic as a temp measure. Help appreciated @oguzhand95!

internal/audit/cerboshub/cerboshub.go
Sambigeara: It's used for both testing and waiting for `sync` completion on shutdown. The [`ForceWrite`](https://github.com/cerbos/cerbos/blob/f63a63d2a0194d731b29ae5dfdff416a7b5e1868/internal/audit/local/badgerdb.go#L189-L198) method closes the buffer and [triggers a call to the callback](https://github.com/cerbos/cerbos/blob/f63a63d2a0194d731b29ae5dfdff416a7b5e1868/internal/audit/local/badgerdb.go#L131-L137). The error channel just means we can block until we're sure the logs have been synced. Similarly [on shutdown](https://github.com/cerbos/cerbos/blob/f63a63d2a0194d731b29ae5dfdff416a7b5e1868/internal/audit/local/badgerdb.go#L126-L130). The test case in particular means that we can do away with any arbitrary and non-guaranteed `Sleep`s in the test code.

In it's existing form, yeah -- it could totally just be a `chan struct{}`.

> which means that there's a small chance that the `awaitCallbackFn` hangs forever because it's reading from an empty channel

I figured this was safe "enough" given the close locality. E.g. you can only wait for channel closure if you create the channel to pass to the callback?

charithe: I see. Makes sense. Since the channel has a size of 1, the `select` for writing to it is probably not necessary? Also might be better to change the type of the callback to `func(chan <-struct{})` to make sure that it's write-only for the callback.

Sambigeara: OK this is probably _really_ silly, but the reason for the `select` was just in case the channel creator initialised an unbuffered channel. At this point I'm not sure which component should be inferring these requirements 🤪

Sambigeara: I totally contradicted myself here:

> I figured this was safe "enough" given the close locality. E.g. you can only wait for channel closure if you create the channel to pass to the callback?

> just in case the channel creator initialised an unbuffered channel

Unbuffered channel with nicely isolated code block seems reasonable enough:

```go
ch := make(chan struct{})
go l.callbackFn(ch)
<-ch
```

internal/audit/cerboshub/conf.go
Sambigeara: I've fixed the yaml inlining so can move onto the `confdocs` generation now (in case you read this Oguzhan).

internal/audit/hub/conf.go
```diff
@@ -0,0 +1,112 @@
...
+	maxFlushTimeout     = 10 * time.Second
+)
+
+var (
+	errInvalidFlushInterval = fmt.Errorf("flushInterval must be at least %s", minMinFlushInterval)
+	errInvalidFlushTimeout  = fmt.Errorf("flushTimeout cannot be more than %s", maxFlushTimeout)
+)
+
+type Conf struct {
+	Ingest     IngestConf `yaml:"ingest"`
```
charithe, line 35: Shall we make this section hidden by adding `conf:",ignore"`? Pretty much all of the settings in it are not things that end users should adjust.

internal/audit/hub/ingest.go
```diff
@@ -0,0 +1,102 @@
...
+	Sync(context.Context, *logsv1.IngestBatch) error
+}
+
+type Impl struct {
+	client *logcap.Client
+	log    *zap.Logger
+}
+
+func NewIngestSyncer(conf *Conf, logger *zap.Logger) (*Impl, error) {
+	pdpID := util.PDPIdentifier(conf.Ingest.Credentials.PDPID)
```
charithe, line 42: I think we may need to update the `util.PDPIdentifier` function a bit. If the user has not specified a PDP ID, we generate one. I think that ID changes each time you call the function. So, we should use a `sync.OnceValue` to keep the generated ID consistent for the lifetime of the process.

internal/audit/local/badgerdb.go
```diff
@@ -122,37 +102,48 @@ func NewLog(conf *Conf, decisionFilter audit.DecisionLogEntryFilter) (*Log, erro
...
 	logger := l.logger.With(zap.String("component", "batcher"))
 
+	ticker := time.NewTicker(flushInterval)
+	defer ticker.Stop()
+
+	awaitCallbackFn := func() {
+		if l.callbackFn != nil {
+			ch := make(chan struct{})
+			go l.callbackFn(ch)
+			<-ch
```
charithe, line 120: Since we are always waiting on `waitCh`, can't we just simplify this to just calling the function without a goroutine? 🤔 

```suggestion
			l.callbackFn()
```

Sambigeara: As it stands, if the server issues a backoff, the callback [implicitly retries in a separate goroutine](https://github.com/cerbos/cerbos/blob/9445e81c786fde9f3c6220420af7ab203dfa1dbd/internal/audit/hub/hub.go#L173), so we can't guarantee that it'll happen synchronously.

The alternative was that we could block for the duration of the retry cycle, but this would have the knock-on effect that the local storage flushes would be blocked also.

internal/audit/hub/conf.go
Sambigeara: If they're going to be hidden, should we just remove them from the config entirely and hardcode them?

internal/audit/hub/ingest.go
Sambigeara: Ah good catch

internal/audit/local/badgerdb.go
charithe: Because of the `return` in line 174, nothing gets published to the `waitCh` until one of the retries succeeds, right? Isn't that the same as having a blocking a function call?

internal/audit/hub/conf.go
charithe: I think it might be useful to have the escape hatch to change them for testing and debugging purposes.

charithe (approved):

PR #2065, opened 2024-03-21
docs: Remove policy version from .NET quickstart

### LINKED ISSUE #2064: .NET Quickstart broken

### Current Behavior

Following the Quickstart documentation using the .NET SDK at `docs/modules/ROOT/pages/quickstart.adoc` fails after adding the rule to the policy to allow users to view public albums.

The .NET SDK example file is here: `docs/modules/ROOT/examples/quickstart/example.cs`

Removing the line `.WithPolicyVersion("20210210")` makes it work again.

It's possible I'm missing something, but it appears to me that the other SDK examples don't specify any policy version. I think it was just a mistake that was accidentally added in this commit: https://github.com/cerbos/cerbos/commit/c7c5f37141ca44dcb9b174ac5d6882eb4eadf949 

If this makes sense, I can raise a PR to remove this line.

### Expected Behavior

Following the Quickstart documentation with the .NET SDK should work properly.

### Steps To Reproduce

Follow the Quickstart documentation using the .NET SDK.

Observe it works properly until the final step `If you try the request again, bugs_bunny now has view:public access to the album owned by daffy_duck but not comment access.`

Remove the line `.WithPolicyVersion("20210210")`, and then the final step works as expected.

### Environment

```markdown
- OS: Arch Linux
- Cerbos version: ghcr.io/cerbos/cerbos:0.34.0
- Docker version: 25.0.4
- Go version: N/A
```

### Anything else?

_No response_

jimpudar:
Specifying a policy version for the resource "DAFFY002" in the .NET SDK version of the quickstart docs causes the final step to not work as described.

Removing this policy version specification fixes the issue, and the quickstart works as intended.

Fixes #2064

- [X] All commits are signed-off (`git commit -s ...`) to provide the [DCO](https://developercertificate.org/)

### DIFF
--- a/docs/modules/ROOT/examples/quickstart/example.cs
+++ b/docs/modules/ROOT/examples/quickstart/example.cs
@@ -27,7 +27,6 @@ internal class Program
 
                 ResourceEntry
                     .NewInstance("album:object", "DAFFY002")
-                    .WithPolicyVersion("20210210")
                     .WithAttribute("owner", AttributeValue.StringValue("daffy_duck"))
                     .WithAttribute("public", AttributeValue.BoolValue(true))
                     .WithAttribute("flagged", AttributeValue.BoolValue(false))


charithe (approved):
Thanks!

PR #2073, opened 2024-03-28
enhancement: Add audit log filtering to Hub backend

Sambigeara:
Add support to specify JSONPath filters for certain Audit Log fields. The values are removed prior to exporting to the Hub backend.

The following fields are included:
* metadata
* peer
* method: `input(s)` and `output(s)` (Decision Logs only)

I opted to use the JSONPath implementation within https://github.com/ohler55/ojg, which seems to be actively maintained and well adopted.

The approach I've taken is to serialize each `IngestBatch` to JSON and then apply the defined JSONPath expressions to it, in their entirety. This means we avoid per-log JSON serialization and deserializaton, and also allows us to take advantage of the performance of the aforementioned library. The tradeoff is that we end up holding the entire serialized batch in memory. However, we've already paginated by the time we reach the filtering step, so I think this is OK.


Sambigeara:
Been doing some rudimentary profiling. Interestingly, the number of JSONPath expressions doesn't seem to impact the filter time, irrespective of the complexity of the rule. I suspect the lions share of the latency is from the JSON (de)serialization.

For reference, I'm fairly consistently seeing the following `Filter()` times:

```
1,000 logs -> ~55ms
10,000 -> ~360ms
100,000 -> ~ 3600ms
etc
```

We should keep an eye on this, but it seems reasonable performance for now.

charithe (commented):
LGTM for the first cut. I am a little concerned about the JSON marshaling and unmarshaling that we need to do to make this work though. I was thinking more along the lines of applying the JSONPath filters directly to the protobufs by traversing them using their `MessageDescriptor`. The downside is that we have to write our own JSONPath parser but we can restrict that to a small subset of things that we need for our purposes.

charithe (dismissed):

Sambigeara:
Makes sense. Given this wouldn't be the final implementation and that we're only interested in a subset of JSONPath operators, perhaps it'll make sense to hold off merging this one and implementing the custom parser? If this version went all the way to production usage, there's a chance we'd break backwards compatibility if users ended up with some whacky, valid filters 🤔.

Could crack on with the `MessageDescriptor` this week. Thoughts?

Sambigeara:
I believe (😅) that this is a subset of JSONPath functionality suitable for our needs.

Pretty satisfying ~7-8x speedup compared to the previous JSON (de)serialization approach.

internal/audit/hub/filter.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+func (f *AuditLogFilter) Filter(ingestBatch *logsv1.IngestBatch) error {
+	if f.astRoot == nil {
+		return nil
+	}
+
+	for _, c := range f.astRoot.children {
+		visit(c, ingestBatch.ProtoReflect())
+	}
```
charithe, line 325:
```suggestion
    ib := ingestBatch.ProtoReflect()
	for _, c := range f.astRoot.children {
		visit(c, ib)
	}
```

internal/audit/hub/filter.go
```diff
@@ -0,0 +1,474 @@
...
+//	    'bosh': {
+//	        'baz',
+//	    },
+//	  }
+//	}
+func visit(t *Token, m protoreflect.Message) {
+	if value, ok := m.Interface().(*structpb.Value); ok {
+		visitStructpb(t, value)
+		return
+	}
```
charithe, line 359: Just avoiding an allocation.

```suggestion
    if m.Type().Descriptor().FullName() == "google.protobuf.Value" {
		visitStructpb(t, m.Interface().(*structpb.Value))
		return
	}
```

internal/audit/hub/filter.go
```diff
@@ -0,0 +1,474 @@
...
+				if !ok {
+					continue
+				}
+				if c.children == nil {
+					if l := len(k.ListValue.Values); idx < l {
+						if l == 1 {
+							v = nil
+						} else {
+							copy(k.ListValue.Values[idx:], k.ListValue.Values[idx+1:])
+							k.ListValue.Values = k.ListValue.Values[: len(k.ListValue.Values)-1 : len(k.ListValue.Values)-1] // also reduce capacity
```
charithe, line 465:
```suggestion
							k.ListValue.Values = slices.Delete(k.ListValue.Values, idx, idx+1)
```

charithe (approved):
🙌🏽

charithe (approved):

charithe (approved):

PR #2078, opened 2024-04-03
feat: Add cerbosctl hub epdp list-candidates command

oguzhand95:
This PR adds `cerbosctl hub epdp list-candidates <path>` command. The command scans the repo and lists the policy IDs and file paths of the policies to be included in the ePDP bundles.

Example
---
image


cmd/cerbosctl/hub/epdp/epdp.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2021-2024 Zenauth Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+package epdp
+
+import "github.com/cerbos/cerbos/cmd/cerbosctl/hub/epdp/list"
+
+type Cmd struct {
+	List list.Cmd `cmd:"" name:"list-candidates" aliases:"lc" help:"List candidates to be included in the ePDP bundle"`
```
charithe, line 9:
```suggestion
	ListCandidates list.Cmd `cmd:"" name:"list-candidates" aliases:"lc" help:"List candidates to be included in the ePDP bundle"`
```

cmd/cerbosctl/hub/epdp/list/list.go
```diff
@@ -0,0 +1,76 @@
...
+
+	policyIDs, err := idx.ListPolicyIDs(context.Background())
+	if err != nil {
+		return fmt.Errorf("failed to list policies: %w", err)
+	}
+
+	policies, err := idx.LoadPolicy(context.Background(), policyIDs...)
+	if err != nil {
+		return fmt.Errorf("failed to load policies: %w", err)
+	}
```
charithe, line 38: Prefer loading policies on demand rather than loading all of them into memory at once.

cmd/cerbosctl/hub/epdp/list/list.go
```diff
@@ -0,0 +1,76 @@
...
+	tw := printer.NewTableWriter(k.Stdout)
+	if !c.NoHeaders {
+		tw.SetHeader([]string{"POLICY ID", "PATH"})
+	}
+
+	items := make([]item, len(policies))
+	foundAnnotation := false
+	for i, p := range policies {
+		hasAnnotation := false
+		if value, ok := p.Metadata.Annotations[embeddedPDPKey]; ok && value == "true" {
```
charithe, line 49: Metadata could be `nil` so we should be more defensive here to avoid a panic.

docs/modules/cli/pages/cerbosctl.adoc
```diff
@@ -349,6 +349,26 @@ cerbosctl get derived_roles my_derived_roles --output=json
...
+
+[#epdp]
+=== `epdp`
+
+Operations related to embedded PDPs.
+
+[#list-candidates]
+==== `list-candidates`
+
+This command lists candidates to be included in the ePDP bundle.
```
charithe, line 365:
```suggestion
This command lists policies that are candidates for inclusion in the ePDP bundle. A policy is marked for inclusion if it is annotated with `hub.cerbos.cloud/embedded-pdp: "true"` in the `metadata.annotations` section of the policy. If a policy has the correct annotation, that policy and its ancestors (if it's a scoped policy) are included the Cerbos Hub embedded PDP bundle. If none of the policies in the repo are annotated, they are all included in the bundle by default. 
```

cmd/cerbosctl/hub/epdp/list/list.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	}
+
+	for _, i := range items {
+		if !foundAnnotation {
+			tw.Append([]string{i.fqn, i.storeIdentifier})
+		} else if i.hasAnnotation {
+			tw.Append([]string{i.fqn, i.storeIdentifier})
+		}
+	}
```
charithe, line 66: You don't need the too-similar-sounding boolean variables here. Instead, maintain two lists: `allPolicies` and `annotatedPolicies`. If `annotatedPolicies` is not empty at the end of the loop, use that. Otherwise use `allPolicies`.

One more thing that's missing here is that if one of the scoped policies is annotated, its ancestors are automatically included regardless of annotations as well. If scope `a.b.c` is annotated, we should include `a.b`, `a` and ` ` too.

cmd/cerbosctl/hub/epdp/list/list.go
charithe, line 1: Just name this file `list_candidates.go` and move it to `cmd/cerbosctl/hub/epdp`. This is not Java so we don't need to have very deep directory trees with trivial packages as leaves :)

cmd/cerbosctl/hub/epdp/list_candidates.go
```diff
@@ -0,0 +1,142 @@
...
+			if len(fqns) > 1 {
+				for _, fqn := range fqns[1:] {
+					if _, ok := annotatedPolicies[namer.PolicyKeyFromFQN(fqn)]; !ok {
+						if pID, ok := allPolicies[namer.PolicyKeyFromFQN(fqn)]; ok {
+							annotatedPolicies[namer.PolicyKeyFromFQN(fqn)] = pID
+						} else {
+							annotatedPolicies[namer.PolicyKeyFromFQN(fqn)] = policyNotFoundInRepo
+						}
+					}
+				}
```
charithe, line 108:
```suggestion
				for _, fqn := range fqns[1:] {
				    key := namer.PolicyKeyFromFQN(fqn)
					if _, ok := annotatedPolicies[key]; !ok {
						if pID, ok := allPolicies[key]; ok {
							annotatedPolicies[key] = pID
						} else {
							annotatedPolicies[key] = policyNotFoundInRepo
						}
					}
				}
```

cmd/cerbosctl/hub/epdp/list_candidates.go
```diff
@@ -0,0 +1,142 @@
...
+		policies = annotatedPolicies
+	} else {
+		policies = allPolicies
+	}
+
+	for policyKey, policyID := range policies {
+		if policyID == policyNotFoundInRepo {
+			delete(policies, policyKey)
+		}
+	}
```
charithe, line 124: Scoped policies must have all their ancestors present in the repo. So, if one of them is missing, the repo is not valid and we should throw an error. That being said, I don't see how it's possible for a policy to be missing at this stage because `index.Build()` should have caught that earlier.

cmd/cerbosctl/hub/epdp/list_candidates.go
```diff
@@ -0,0 +1,142 @@
...
+		tw.SetHeader([]string{"POLICY ID", "PATH"})
+	}
+
+	candidates, err := listCandidates(ctx, loadPolicy(idx), policyIDs...)
+	if err != nil {
+		return fmt.Errorf("failed to list candidates: %w", err)
+	}
+
+	for policyKey, policyID := range candidates {
+		tw.Append([]string{namer.PolicyKeyFromFQN(policyKey), policyID})
```
charithe, line 66: We're already getting the policy keys here so this is unnecessary. 
```suggestion
		tw.Append([]string{policyKey, policyID})
```

cmd/cerbosctl/hub/epdp/list_candidates_test.go
```diff
@@ -0,0 +1,146 @@
...
+			},
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.testFile, func(t *testing.T) {
+			fsys := test.ExtractTxtArchiveToFS(t, filepath.Join("testdata", testCase.testFile))
+			pl := policyLoader{fsys: fsys}
+
+			candidates, err := listCandidates(context.Background(), pl.loadPolicy, testCase.files...)
```
charithe, line 115: This test is not good because `listCandidates` is missing the crucial `index.Build()` part that would make most of your test cases fail because the repos are invalid. So, it's not behaving as it would in a real-world scenario.

cmd/cerbosctl/hub/epdp/list_candidates_test.go
```diff
@@ -0,0 +1,88 @@
...
+			},
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.testFile, func(t *testing.T) {
+			fsys := test.ExtractTxtArchiveToFS(t, filepath.Join("testdata", testCase.testFile))
+			candidates, err := listCandidates(context.Background(), fsys)
+			if testCase.expectedErr != "" {
+				require.ErrorContains(t, err, "failed to build index: failed to build index: missing imports=0, missing scopes=1, duplicate definitions=0, load failures=0")
```
charithe, line 73:
```suggestion
				require.ErrorContains(t, err, testCase.expectedErr)
```

cmd/cerbosctl/hub/epdp/testdata/descendant_observed_first.txt
charithe, line 1: Add another un-annotated policy with scope `regional.uk` or something to make sure that the list doesn't contain it.

charithe (approved):

PR #2079, opened 2024-04-04
chore: InspectPolicies returns FQN instead of policy key

oguzhand95:
`cerbos.resource.leave_request.vdefault/regional` to `resource.leave_request.vdefault/regional`


charithe (approved):
Good catch!

PR #2109, opened 2024-04-19
chore(docs): Update cloud-platforms.adoc

rohitg00:
Topic: Fly listens only on "0.0.0.0"
Changes: Tip added

> Services to be routed to using Fly Proxy need to bind to IPv4 addresses, and services to be reached over 6PN need to bind to IPv6, as described above. However, you may find that your service works even if you haven’t used the addresses we specified. IPv4-mapped IPv6 addresses, if enabled on the VM, can allow IPv4 (and thus Fly Proxy) connections to a service bound on [::]. Further, the wildcard syntax [::] or 0.0.0.0 sometimes covers both IPv4 and IPv6, depending on the language, library, or application.

#### Description

Fixes #

#### Checklist 


oguzhand95 (approved):
lgtm

PR #2114, opened 2024-04-22
chore(ci): Fix E2E tests combining the host address with extra colon

oguzhand95:
Due to a previous [commit](https://github.com/cerbos/cerbos/commit/4929745082e4e82794e87acce30d8d6de79cbb28), the E2E tests have been combining a host address with an extra colon.
This PR fixes the problem and adds an extra log for the gRPC health checks which made it easier to debug this problem.

### DIFF
--- a/internal/server/tests.go
+++ b/internal/server/tests.go
@@ -104,7 +104,7 @@ func (tr *TestRunner) RunGRPCTests(addr string, opts ...grpc.DialOption) func(*t
 	return func(t *testing.T) {
 		grpcConn := mkGRPCConn(t, addr, opts...)
 		require.Eventually(t,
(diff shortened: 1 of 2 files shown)


internal/server/tests.go
```diff
@@ -452,6 +452,7 @@ func grpcHealthCheckPasses(grpcConn *grpc.ClientConn, reqTimeout time.Duration)
 
 		resp, err := client.Check(ctx, &healthpb.HealthCheckRequest{})
 		if err != nil {
+			t.Errorf("gRPC health check failed: %s", err.Error())
```
charithe, line 455: Errorf actually fails the test.

```suggestion
			t.Logf("gRPC health check failed: %v", err)
```

charithe (approved):

PR #2122, opened 2024-04-26
fix: Stop blocking Kafka audit publishing when an outage occurs

### LINKED ISSUE #2123: Kafka audit will block even when configured to be flushing records async

### Current Behavior

Kafka audit publisher will block once the max buffered records are hit, even when `sync: false` is set in the configuration.

### Expected Behavior

If the underlying Kafka instance has an issue, it was designed that the non-sync publishing would continue but report an error through logs & metrics. However, after a period of time the buffer becomes full & the underlying Kafka client moves to blocking.

### Steps To Reproduce

We replicated the bug by :
- configure invalid TLS certificates
- start sending requests to Cerbos
- after enough requests are sent, these will block until a timeout

### Environment

```markdown
- Cerbos version: v0.35.1
```

### Anything else?

#2122

rcrowe:
#### Description

`TryProduce` is similar to the previous `Produce` call, but rather than blocking if the Kafka client currently has MaxBufferedRecords or MaxBufferedBytes buffered, this fails immediately.

https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#Client.TryProduce

Cerbos when configured with `produceSync: false` should handle an outage at the Kafka layer by continuing but reporting the problem through metrics & logs.

Fixes #2123 

#### Checklist 

### DIFF
--- a/internal/audit/kafka/publisher.go
+++ b/internal/audit/kafka/publisher.go
@@ -68,7 +68,7 @@ func init() {
 type Client interface {
 	Close()
 	Flush(context.Context) error
-	Produce(context.Context, *kgo.Record, func(*kgo.Record, error))
+	TryProduce(context.Context, *kgo.Record, func(*kgo.Record, error))
 	ProduceSync(context.Context, ...*kgo.Record) kgo.ProduceResults
 }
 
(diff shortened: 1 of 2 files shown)


charithe (commented):
Thanks. This seems sensible to me. Just to make sure, if someone preferred consistency over availability (i.e. not lose any audit records), they would set `produceSync: true` and that would block everything until Kafka recovers, right?

rcrowe:
@charithe exactly, if you must have audit records published for every request/decision you should look to configure `produceSync: true; I haven't taken a look at the change in #2104 yet to understand how some of this plays out when you want sync as I'm only interested in async.

charithe (approved):

PR #2141, opened 2024-05-07
feat: List local and imported variables in the policy with InspectPolicies

oguzhand95:
InspectPolicies on `disk` store
---
```json
{
    "results": {
        "derived_roles.common_roles": {
            "variables": [
                {
                    "name": "derivedRoleVariable",
                    "value": "request.resource.attr.isDerivedRoleVar",
                    "kind": "KIND_LOCAL",
                    "source": "derived_roles.common_roles"
                }
            ]
        },
        "export_variables.common_variables": {
            "variables": [
                {
                    "name": "commonMarkedResource",
                    "value": "request.resource.attr.markedResource",
                    "kind": "KIND_EXPORTED",
                    "source": "export_variables.common_variables"
                }
            ]
        },
        "principal.john.vdefault": {
            "actions": [
                "*"
            ],
            "variables": [
                {
                    "name": "commonMarkedResource",
                    "value": "request.resource.attr.markedResource",
                    "kind": "KIND_IMPORTED",
                    "source": "export_variables.common_variables"
                },
                {
                    "name": "markedResource",
                    "value": "request.resource.attr.markedResource",
                    "kind": "KIND_LOCAL",
                    "source": "principal.john.vdefault"
                }
            ]
        },
        "resource.leave_request.vdefault": {
            "actions": [
                "*",
                "create",
                "duplicate",
                "view"
            ],
            "variables": [
                {
                    "name": "commonMarkedResource",
                    "value": "request.resource.attr.markedResource",
                    "kind": "KIND_IMPORTED",
                    "source": "export_variables.common_variables"
                },
                {
                    "name": "markedResource",
                    "value": "request.resource.attr.markedResource",
                    "kind": "KIND_LOCAL",
                    "source": "resource.leave_request.vdefault"
                }
            ]
        }
    }
}
```

`cerbosctl inspect policies` on `disk` store
---
```bash
POLICY ID                               ACTIONS                 VARIABLES                           
derived_roles.common_roles                                      derivedRoleVariable                     
export_variables.common_variables                               commonMarkedResource                    
principal.john.vdefault                 *                       commonMarkedResource,markedResource     
resource.leave_request.vdefault         *,create,duplicate,view commonMarkedResource,markedResource 
```

InspectPolicies on `hub` store
---
```json
{
    "results": {
        "principal.john.vdefault": {
            "actions": [
                "*"
            ],
            "variables": [
                {
                    "name": "commonMarkedResource",
                    "value": "request.resource.attr.markedResource",
                    "kind": "KIND_UNKNOWN"
                },
                {
                    "name": "markedResource",
                    "value": "request.resource.attr.markedResource",
                    "kind": "KIND_UNKNOWN"
                }
            ]
        },
        "resource.leave_request.vdefault": {
            "actions": [
                "*",
                "create",
                "duplicate",
                "view"
            ],
            "variables": [
                {
                    "name": "commonMarkedResource",
                    "value": "request.resource.attr.markedResource",
                    "kind": "KIND_UNKNOWN"
                },
                {
                    "name": "markedResource",
                    "value": "request.resource.attr.markedResource",
                    "kind": "KIND_UNKNOWN"
                }
            ]
        }
    }
}
```

`cerbosctl inspect policies` on `hub` store
---
```bash
POLICY ID                       ACTIONS                 VARIABLES                           
principal.john.vdefault         *                       commonMarkedResource,markedResource     
resource.leave_request.vdefault *,create,duplicate,view commonMarkedResource,markedResource
```


api/public/cerbos/response/v1/response.proto
```diff
@@ -444,8 +444,29 @@ message InspectPoliciesResponse {
   option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
     json_schema: {description: "Inspect policies response"}
   };
+
+  message Variable {
+    message Source {
+      enum Type {
```
charithe, line 450: I think `Kind` would be a better name because we are talking about variables and "type" could be misinterpreted as the actual data type of the variable. Also, because `type` is a keyword, it'd be a bit painful to work with in code.

api/public/cerbos/response/v1/response.proto
```diff
@@ -444,8 +444,29 @@ message InspectPoliciesResponse {
...
+        TYPE_UNSPECIFIED = 0;
+        TYPE_EXPORTED = 1;
+        TYPE_IMPORTED = 2;
+        TYPE_LOCAL = 3;
+        TYPE_LOCAL_OR_IMPORTED = 4;
```
charithe, line 455: Is this for variables that are defined more than once?

api/public/cerbos/response/v1/response.proto
```diff
@@ -444,8 +444,29 @@ message InspectPoliciesResponse {
...
+      }
+
+      Type type = 1 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Type of the variable source."}];
+      string id = 2 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Source policy ID of the variable if available."}];
+    }
+
+    string name = 1 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Variable name defined in the policy."}];
+    string value = 2 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Raw value of the variable defined in the policy."}];
+    Source source = 3 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Source of the variable defined in the policy."}];
```
charithe, line 464: Instead of using a message here, I'd use two fields. `Kind` (local or imported) and `Source` if it's an imported variable. If the variable is local, the source is obviously the policy we are currently inspecting so it's a bit redundant to repeat its name.

docs/modules/api/pages/admin_api.adoc
```diff
@@ -170,7 +170,7 @@ curl -k -u cerbos:cerbosAdmin \
 GET /admin/policies/inspect
 ----
 
-Issue a GET request to the endpoint to list actions covered by the policies in the store. If the policy store supports
+Issue a GET request to the endpoint to list actions, variables covered by the policies in the store. If the policy store supports
```
charithe, line 173:
```suggestion
Issue a GET request to the endpoint to list actions and variables covered by the policies in the store. If the policy store supports
```

internal/policy/policy.go
```diff
@@ -359,6 +359,111 @@ type Wrapper struct {
...
+				Source: &responsev1.InspectPoliciesResponse_Variable_Source{
+					Type: responsev1.InspectPoliciesResponse_Variable_Source_TYPE_LOCAL,
+					Id:   namer.PolicyKey(p),
+				},
+			})
+		}
+	}
+
+	sort.Slice(variables, func(i, j int) bool {
+		return sort.StringsAreSorted([]string{variables[i].Name, variables[j].Name})
```
charithe, line 462: Isn't it more efficient to do `variables[i].Name < variables[j].Name`? Otherwise this requires allocating an array and calling a function, which may or may not get inlined.

api/public/cerbos/response/v1/response.proto
charithe: `Unknown` would be better.

oguzhand95:
```proto
enum Kind {
  KIND_UNSPECIFIED = 0;
  KIND_EXPORTED = 1;
  KIND_IMPORTED = 2;
  KIND_LOCAL = 3;
}
```

Since `KIND_UNSPECIFIED` is mandatory, I assume you mean adding `UNKNOWN` as number 4 to keep it explicit?

charithe: Yes. `UNSPECIFIED` is a protobuf convention to indicate an unset enum so we shouldn't use it for anything else.

internal/inspect/internal/policy/inspect.go
```diff
@@ -0,0 +1,324 @@
...
+		Actions:    policy.ListActions(p),
+		Conditions: conditions,
+		Imports:    i.inspectImports(p),
+		Variables:  policy.ListVariables(p),
+	}
+
+	return nil
+}
+
+func (i *Inspect) Results() (map[string]*responsev1.InspectPoliciesResponse_Result, error) {
```
charithe, line 53: It took me a long time to see what was going on here because of all the deeply nested loops and if conditions 🙂 I think we can we try to split this into smaller, more understandable bits.

The `Inspect` method above should take a policy and do the following:
- If it's an `ExportVariables` policy, all its variables are resolved already. Just add the list of sorted variable names to `variables` as `KIND_EXPORT`.
- For any other policy type:
  - Extract the list of variables referenced by its conditions
  - Add the local variables to `variables` list with kind `KIND_LOCAL` and, if any of them are referenced by the conditions, mark as "used". (I just realised we need another flag in the protobuf to mark whether a variable is used or not.)
  - For the remainder of referenced variables that are not local variables, add them to a temporary field `toResolve`

In the `Results` method: 
  - Go through the list of policies that have a non-empty `toResolve` field 
     - If they import any `ExportVariables` load them (if we haven't already loaded them during our traversal) and resolve the `KIND_IMPORT` variables.
     - Any variable that's not defined in either local or imported variables should be marked as `KIND_UNDEFINED` and have `null` as the value.

oguzhand95: Since we do not have the option to show unused variables in the compiled policy stores, I opted to only list the used variables in the `inspect.Policy` too. If we introduce a `used` field, we cannot provide this information in the compiled policy stores which could be confusing.

> Any variable that's not defined in either local or imported variables should be marked as KIND_UNDEFINED and have null as the value.

I think we should always assume the compilation step is successful, and this step is not possible in that case, no?

I have put some effort to modify the code to follow the steps you've written and added some more steps in between to cover what the current implementation does. All in all, I think it did not change much in terms of readability though I have not completed it.

The actual complexity comes from resolving the variables used in the conditions. For example the following is not as simple because I also remove the unused variables from the export variables.

> If it's an ExportVariables policy, all its variables are resolved already. Just add the list of sorted variable names to variables as KIND_EXPORT.

charithe: > Since we do not have the option to show unused variables in the compiled policy stores, I opted to only list the used variables in the `inspect.Policy` too. If we introduce a `used` field, we cannot provide this information in the compiled policy stores which could be confusing.

Everything would be "used" in compiled policies because they are in an optimized format and we don't have the raw source. I don't think that would be confusing for users. The output of `inspect` would be consistent for the store in use because Cerbos can only have one store anyway.   


> I think we should always assume the compilation step is successful, and this step is not possible in that case, no?

If you're talking about `condition.Compile`, that doesn't detect undefined variables because `V` is a dynamic container and it has no idea what it could contain at runtime.


> I also remove the unused variables from the export variables.

If I inspect an `ExportVariables` policy with 10 variables but none of those are used by any other policies, do I see 10 variables or 0? It should be 10 when inspecting that policy explicitly. For all other policy types, if they import any variables, we should list all of them along with whether they are actually used or not.

oguzhand95: > If I inspect an ExportVariables policy with 10 variables but none of those are used by any other policies, do I see 10 variables or 0?

The result was 0 in this case.

oguzhand95:
- Refactored the code following your steps and it is simpler now.
- Added `used` field to denote if a variable is used from a condition. However, the variables defined in the `ExportVariables` will not show up as `used` as you've stated. In the policy where they are used, they appear as `imported` and `used`.
- If a variable is referenced from a condition, and the variable is nowhere to be found it is now marked as `UNDEFINED`.

I was querying the unfiltered and filtered policies from the database and still do. I use the unfiltered one to resolve the state of each variable so that we do not report any undefined variables in the inspection. If we only check the filtered policies while inspecting, then some variables may appear as undefined even if they are actually defined in the policies left out. I am not sure if this is the expected resolution though, let me know if this is not what you had in mind.

internal/storage/db/internal/db.go
```diff
@@ -655,29 +656,43 @@ func (s *dbStorage) Delete(ctx context.Context, ids ...namer.ModuleID) error {
...
 
-	results := make(map[string]*responsev1.InspectPoliciesResponse_Result, len(policyIDs))
-	if err := storage.BatchLoadPolicy(ctx, storage.MaxPoliciesInBatch, s.LoadPolicy, func(p *policy.Wrapper) error {
-		actions := policy.ListActions(p.Policy)
-		if len(actions) > 0 {
-			results[namer.PolicyKeyFromFQN(p.FQN)] = &responsev1.InspectPoliciesResponse_Result{
-				Actions: actions,
-			}
-		}
+	policyIDs, err := s.ListPolicyIDs(ctx, storage.ListPolicyIDsParams{})
```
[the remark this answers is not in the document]
oguzhand95: Answered [here](https://github.com/cerbos/cerbos/pull/2141#issuecomment-2108821353).

charithe:
> I was querying the unfiltered and filtered policies from the database and still do. I use the unfiltered one to resolve the state of each variable

If there are a lot of policies, this would be quite inefficient. The list of filtered policies might not even be importing any variables -- in which case we would have done a lot of work for nothing. 

Get the list of policies the user is interested in inspecting. While going through that list one by one, load and cache any imported variables as necessary. Or, do one pass, collect the set of imported policies that are referenced but not loaded initially, load all of them as a batch, and do a second pass to handle the unresolved variables.

internal/inspect/policy.go
```diff
@@ -0,0 +1,315 @@
...
+	for name := range referencedVariables {
+		if _, ok := localVariables[name]; !ok {
+			if pol.toResolve[policyID] == nil {
+				pol.toResolve[policyID] = make(map[string]bool)
+			}
+
+			toResolve = true
+			pol.toResolve[policyID][name] = false
+		}
+	}
```
charithe, line 80: Rename `variables` to `localVariables` as suggested above and avoid a copy.

```suggestion
	for referenced := range referencedVariables {
	    if v, ok := localVariables[referenced]; ok {
	        v.Used = true
	    } else {
	        toResolve, exists := pol.toResolve[policyID]
	        if !exists {
	           pol.toResolve[policyID] = map[string]struct{}{name: {}}
	        } else {
	           toResolve[name] = struct{}{}
	        }
	    }
	}
```

internal/inspect/policy.go
```diff
@@ -0,0 +1,315 @@
...
+
+	a := policy.ListActions(p)
+	sort.Strings(a)
+	if len(a) > 0 || len(variables) > 0 {
+		pol.results[policyID] = &responsev1.InspectPoliciesResponse_Result{
+			Actions:   a,
+			Variables: variables,
+		}
+	}
```
charithe, line 96:
```suggestion
	// sort variables if there is nothing to resolve since we are not going to modify variables in the future.
	if len(po.toResolve[policyID]) > 0 {
		sort.Slice(variables, func(i, j int) bool {
			return variables[i].Name < variables[j].Name
		})
	}

	a := policy.ListActions(p)
	sort.Strings(a)

	pol.results[policyID] = &responsev1.InspectPoliciesResponse_Result{
		Actions:   a,
		Variables: variables,
	}
```

internal/inspect/policy.go
```diff
@@ -0,0 +1,315 @@
...
+
+// Results returns the final inspection results.
+func (pol *Policy) Results() (map[string]*responsev1.InspectPoliciesResponse_Result, error) {
+	for policyID, variables := range pol.toResolve {
+		importedPolicies, ok := pol.imports[policyID]
+		if ok {
+			for _, importedPolicyID := range importedPolicies {
+				importedResult, ok := pol.results[importedPolicyID]
+				if !ok {
+					return nil, fmt.Errorf("failed to find imported policy %s in the inspected policies", importedPolicyID)
```
charithe, line 109: We should try to load this policy from the store. It's quite possible that the list of policies the user asked to inspect doesn't contain all the required `ExportVariables` policies. It's not an error.

- Go through the list of policies with unresolved variables
- Collect their imports into a set
- Load that set from the store and try to resolve the unresolved variables again
- Anything remaining unresolved is an undefined variable

internal/inspect/policyset.go
```diff
@@ -0,0 +1,57 @@
...
+			return sort.StringsAreSorted([]string{variables[i].Name, variables[j].Name})
+		})
+	}
+
+	if len(actions) > 0 || len(variables) > 0 {
+		ps.results[namer.PolicyKeyFromFQN(pset.Fqn)] = &responsev1.InspectPoliciesResponse_Result{
+			Actions:   actions,
+			Variables: variables,
+		}
+	}
```
charithe, line 49: If the user asked to inspect policy `foo`, it should exist in the final result regardless of whether it has any actions or variables.

```suggestion

		ps.results[namer.PolicyKeyFromFQN(pset.Fqn)] = &responsev1.InspectPoliciesResponse_Result{
			Actions:   actions,
			Variables: variables,
		}

```

internal/storage/hub/bundle.go
```diff
@@ -236,31 +237,49 @@ func (b *Bundle) loadPolicySet(idHex, fileName string) (*runtimev1.RunnablePolic
...
+			return nil, fmt.Errorf("failed to load policy %s: %w", fqn, err)
 		}
 
-		actions := policy.ListPolicySetActions(pset)
-		if len(actions) > 0 {
-			results[namer.PolicyKeyFromFQN(pset.Fqn)] = &responsev1.InspectPoliciesResponse_Result{
-				Actions: actions,
-			}
+		if err = ins.Inspect(pset); err != nil {
+			return nil, fmt.Errorf("failed to inspect policy %s: %w", fqn, err)
```
charithe, line 266: Isn't this inspecting the whole bundle and not just the ones the user asked for?

internal/inspect/policy.go
oguzhand95: I do something like this with the dbStorage. The steps:
- Call `ListPolicyIDs` with filters
  - For each policy, call `Inspect` method which also caches the imported policy names into `pol.imports`
- After the list of policies are exhausted, call the [MissingImports](https://github.com/cerbos/cerbos/pull/2141/files#diff-9fe34d6e3fe21b240b8b0e0e84996b570289e47dd03e655d5e9f44dcd08d8fd1R100-R117) method to get policies left outside of filtered policies
- [Call LoadPolicy with the list of missing policies and call Inspect on them](https://github.com/cerbos/cerbos/pull/2141/files#diff-1823d40180fc37bc05da9d1175fee7d28d4ccad65bde1542981f12e4cfe30782R677)
- Call and return `Results`

internal/inspect/policy.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+// Inspect inspects the given policy and caches the inspection related information internally.
+func (pol *Policy) Inspect(p *policyv1.Policy) error {
+	if p == nil {
+		return fmt.Errorf("policy is nil")
+	}
+
+	policyID := namer.PolicyKey(p)
+	variables := policy.ListVariables(p)
```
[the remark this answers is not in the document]
oguzhand95: I am not sure if understood what you mean exactly. 

Though I changed the `policy.ListVariables` to return a map instead of a list prevent creating the map afterwards. Because it was only called from a single place and it'd better to return a map for that case.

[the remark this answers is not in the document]
charithe: Yeah, it makes sense to get the return value in the shape you need.

internal/inspect/policy.go
charithe: I saw that. There are a couple of issues with that.
- It implicitly adds the missing imported variables to your result. The user didn't ask for those so it's surprising behaviour.
- The logic is now in multiple places. You have to remember to call `MissingImports` and inspect those too before calling `Result`. 

The missing imports are a special case. They shouldn't be added the final result as "policies". We only need them for resolving variables we found while inspecting the original set of policies. I'd add a parameter to the `Result` method. It'd be a function that takes the list of missing imports and returns the variables defined in those imports.

internal/inspect/policy.go
```diff
@@ -0,0 +1,289 @@
...
+					}
+				} else {
+					missingPolicies = append(missingPolicies, importedPolicyID)
+				}
+			}
+		}
+
+		if loadPolicy != nil {
+			for _, importedPolicyID := range missingPolicies {
+				importedPolicy, err := loadPolicy(ctx, importedPolicyID)
```
charithe, line 131: Wouldn't it be more efficient to load all the missing policies at once? Saves some roundtrips to the backing store if it's a remote one like S3 or a database.

internal/storage/index/index.go
```diff
@@ -454,25 +455,19 @@ func (idx *index) InspectPolicies(ctx context.Context) (map[string]*responsev1.I
...
-		return nil
+	ins := inspect.Policies()
+	if err := storage.BatchLoadPolicy(ctx, 1, idx.LoadPolicy, func(wp *policy.Wrapper) error {
+		return ins.Inspect(wp.Policy)
 	}, policyIDs...); err != nil {
 		return nil, fmt.Errorf("failed to load policy: %w", err)
 	}
 
-	return results, nil
+	return ins.Results(ctx, nil)
```
charithe, line 470: Should the `loadPolicy` argument be `nil`?

charithe (approved):

PR #2156, opened 2024-05-16
chore: Update error message for invalid expression

charithe:
Makes the error message consistent with others by moving the compile errors to the description.

### DIFF
--- a/internal/test/testdata/compile/bad_cel_expr.yaml
+++ b/internal/test/testdata/compile/bad_cel_expr.yaml
@@ -3,9 +3,9 @@
 wantErrors:
   - file: principal_policies/donald_duck_20210210.yaml
     error: |-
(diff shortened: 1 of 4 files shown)


internal/test/testdata/compile/bad_cel_expr.yaml
```diff
@@ -3,9 +3,9 @@
 wantErrors:
   - file: principal_policies/donald_duck_20210210.yaml
     error: |-
-      failed to compile `resource.attr.dev_record ^^ true` [Syntax error: token recognition error at: '^', Syntax error: token recognition error at: '^', Syntax error: extraneous input 'true' expecting <EOF>]
+      invalid expression `resource.attr.dev_record ^^ true`
```
haines, line 6: If we have the expression in the description, do we need it in the error message also? It would be better to keep the error to a couple of words to match the others. There is not much room in the Hub UI where we display this, and the expression could be quite long.

haines (approved):

PR #2160, opened 2024-05-22
feat: Add filtering by policy IDs to InspectPolicies RPC

### LINKED ISSUE #2159: Allow specifying list of policies in inspect policies RPC
The `InspectPolices` RPC and `cerbosctl inspect policies` should be able to accept a list of policy IDs to inspect. This is not the same as the filters because they are not supported by all stores and regexes are too cumbersome to construct when you know the exact list of things you want to inspect.

oguzhand95:
resolves #2159


api/public/cerbos/request/v1/request.proto
```diff
@@ -651,6 +651,23 @@ message InspectPoliciesRequest {
...
+  repeated string id = 5 [
+    (google.api.field_behavior) = OPTIONAL,
+    (buf.validate.field).repeated = {
+      unique: true,
+      max_items: 25,
+      items {
+        string {
+          min_len: 1,
+          max_len: 1280
+        }
```
charithe, line 663: I don't think that uniqueness is strictly required because we can deal with that on the server side. It'd be a bad experience to reject a request because the user accidentally included the same ID twice.

Do we have a limit of `max_len = 1280` defined elsewhere? Otherwise, I don't think we need to limit the maximum length.

```suggestion
      max_items: 50,
      items {
        string {
          min_len: 1,
        }
```

api/public/cerbos/request/v1/request.proto
```diff
@@ -651,6 +651,23 @@ message InspectPoliciesRequest {
     (google.api.field_behavior) = OPTIONAL,
     (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Filter policies by version with regexp"}
   ];
+  repeated string id = 5 [
```
charithe, line 654:
```suggestion
  repeated string policy_id = 5 [
```

docs/modules/api/pages/admin_api.adoc
```diff
@@ -175,6 +175,8 @@ store supports filtering, you can optionally pass filter parameters to reduce th
 
 Use `includeDisabled=true` query parameter in order to include disabled policies in the response.
 
+Use `id` query parameter to provide list of policy IDs to inspect.
```
charithe, line 178:
```suggestion
Use `policy_id` query parameter to provide a list of policy IDs to inspect.
```

internal/storage/db/internal/db.go
```diff
@@ -655,29 +655,41 @@ func (s *dbStorage) Delete(ctx context.Context, ids ...namer.ModuleID) error {
 	return nil
 }
 
-func (s *dbStorage) InspectPolicies(ctx context.Context, listParams storage.ListPolicyIDsParams) (map[string]*responsev1.InspectPoliciesResponse_Result, error) {
-	policyIDs, err := s.ListPolicyIDs(ctx, listParams)
+func (s *dbStorage) InspectPolicies(ctx context.Context, params storage.InspectPoliciesParams) (map[string]*responsev1.InspectPoliciesResponse_Result, error) {
+	policyIDs, err := s.ListPolicyIDs(ctx, storage.ListPolicyIDsParams{
```
charithe, line 659: We shouldn't make filtering by IDs a special case. It can be just another filtering option in the same vein as `NameRegexp`, `IncludeDisabled` etc. 

The list of IDs should be part of the database query. Otherwise, if no other filters are defined, we'll be loading everything in the database to memory and filtering them out -- which is really inefficient.

api/public/cerbos/request/v1/request.proto
oguzhand95: `GetPolicy` request also has a limit of 1280.

charithe: OK. I think we can remove that restriction because it doesn't include the scope nor the periods and slashes that are part of the ID.

internal/storage/db/internal/db.go
oguzhand95: > We shouldn't make filtering by IDs a special case. It can be just another filtering option in the same vein as NameRegexp, IncludeDisabled etc.

Then maybe we could add `IDs` filtering option to the `storage.ListPolicyIDsParams` and use it in the `InspectPolicies` like used to. This way `ListPolicyIDs` could support filtering by IDs too.

charithe: Yes. Sorry, I should have been clear about it. I thought it was obvious.

internal/storage/index/index.go
```diff
@@ -450,33 +450,38 @@ func (idx *index) Inspect() map[string]meta {
...
-	policyIDs, err := idx.ListPolicyIDs(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to list policies: %w", err)
-	}
-
-	if len(policyIDs) == 0 {
-		return nil, nil
-	}
-
+func (idx *index) InspectPolicies(ctx context.Context, file ...string) (map[string]*responsev1.InspectPoliciesResponse_Result, error) {
```
charithe, line 453: There's a bug here. If the user does not explicitly provide any policy IDs to inspect, the result should contain everything in the store. So, you still need to call `ListPolicyIDs` if the `file` argument is empty.

charithe (approved):
LGTM. Let's add the RPC to the SDK and implement `cerbosctl` so that it can be tested easily.

PR #2169, opened 2024-05-29
feat(repl): Autocomplete for REPL directives, and a small fix for filenames

tcolgate:
#### Description

There are two commits here, but one is a simple one liner.
- fix `:load` with a plain filename, these were rejected by the parser as they were treated as identifiers
- adds some autocomplete support for `:` directives:
  - `:...` get completed with directives
  - `:load` completes filenames
  - `:let` completes variables
  - `:exec` completes rule numbers

### DIFF
--- a/cmd/cerbos/repl/internal/repl.go
+++ b/cmd/cerbos/repl/internal/repl.go
@@ -9,11 +9,16 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
+	"path/filepath"
 	"reflect"
+	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
 	"github.com/alecthomas/participle/v2"
 	"github.com/google/cel-go/cel"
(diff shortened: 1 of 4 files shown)


tcolgate:
Added tests

cmd/cerbos/repl/internal/repl.go
```diff
@@ -688,6 +694,137 @@ func (r *REPL) mkEnv() (*cel.Env, error) {
...
+		res[i] = strings.Join(parts, " ")
+	}
+
+	return res
+}
+
+func (r *REPL) completeCmd(prefix string) (c []string) {
+	var matches []string
+
+	ds := []string{
```
charithe, line 740: This array can be a package-level variable to avoid allocating it over and over again whenever `completeCmd` is called.

cmd/cerbos/repl/internal/repl.go
```diff
@@ -99,6 +104,7 @@ func NewREPL(reader *liner.State, output Output) (*REPL, error) {
 		parser:   parser,
 		output:   output,
 		toRefVal: conditions.StdEnv.CELTypeAdapter().NativeToValue,
+		fs:       os.DirFS("."),
```
charithe, line 107: Instead of limiting to the current working directory, I think we can create the FS on demand in the `completeFile` method using the parent directory of the `prefix`.

charithe (commented):
Very cool! Left a couple of suggestions below. LGTM otherwise.

charithe (approved):

cmd/cerbos/repl/internal/repl_test.go
```diff
@@ -544,3 +545,133 @@ func loadConditionsFromPolicy(t *testing.T, path string) []*runtimev1.Condition
...
+	rpPath := filepath.Join(test.PathToDir(t, "store"), "resource_policies", "policy_01.yaml")
+	p := test.LoadPolicy(t, rpPath)
+	rp := p.PolicyType.(*policyv1.Policy_ResourcePolicy)
+
+	testRules := rp.ResourcePolicy.Rules
+	testRules = testRules[0:3]
+	testRulesProtos := make([]proto.Message, len(testRules))
+
+	//nolint:gosimple // seems to be a false report of possible copy
+	for i, r := range testRulesProtos {
```
oguzhand95, line 560: Is looping over `testRulesProtos` and assigning to `testRulesProtos` intentional?

tcolgate: nope! It doesn't impact the test because the completion / input is only the the index. Though Interestingly, when I tried using an empty array, it was crashing, (hence loading the valid rules).

cmd/cerbos/repl/internal/repl.go
tcolgate: This turned out to require dropping fs.FS and using actual files for the test. I've removed fs.FS from code I've written more often than I've left it in, a real shame.

charithe (approved):

oguzhand95 (approved):

cmd/cerbos/repl/internal/repl.go
charithe: Yeah, I still have to fallback to the Afero library for anything that requires more than the most basic read-only operations.

PR #2174, opened 2024-06-03
feat: Add policy id parameter to inspect command

oguzhand95:


cmd/cerbosctl/get/internal/flagset/filter.go
```diff
@@ -15,6 +15,7 @@ type Filters struct {
 	NameRegexp      string   `help:"Filter policies by name, using regular expression"`
 	Version         []string `help:"Filter policies by version"`
 	VersionRegexp   string   `help:"Filter policies by version, using regular expression"`
+	PolicyID        []string `help:"Filter policies when listing by policy ID"`
```
charithe, line 18: It's more natural to include the list of IDs as trailing arguments. E.g. `cerbosctl inspect policies a.b.c d.e.f`. It's also consistent with how the `get` commands already work. So, I'd mark this as an `arg` type and remove the special `policyIds` struct field in the `get` sub commands in  favour of this new filter field.

cmd/cerbosctl/inspect/policies.go
```diff
@@ -28,13 +28,20 @@ cerbosctl inspect policies`
 	separator = ","
 )
 
+//nolint:govet
 type PoliciesCmd struct {
 	flagset.Filters
 	flagset.Format
+
+	PolicyIDs []string `arg:"" name:"id" optional:"" help:"list of policy ids"`
```
charithe, line 36: I was confused by this until I realised that `inspect` command has its own `flagset.Filters` which is nearly identical to the `get` command filter flags except for `name` and `version`. That's fair enough. However, there are two inconsistencies in here now.
- One `flagset.Filters` definition includes `PolicyIDs` while the other doesn't. That's not ideal in terms of code consistency.
- The validation logic for flags is now inconsistent: the `get` command doesn't allow using `--include-disabled` if any `policyIDs` are defined but, the `inspect` command does. That's a user-facing inconsistency. 

I think that that validation check for `--include-disabled` + `policyIDs` in `get` is not necessary because `include-disabled` is a harmless no-op in that case anyway. It should be removed so that we don't throw unnecessary errors at users. I'd also try to keep the `flagset.Filters` definitions as similar as possible in the code.

docs/modules/cli/pages/cerbosctl.adoc
```diff
@@ -255,6 +255,8 @@ Separately, you can filter the output using the `name-regexp`, `version-regexp`
 
 You can include disabled policies in the results by adding `--include-disabled` flag.
 
+`policy-id` flag allows to filter the policies by the policy id while listing.
```
charithe, line 258:
```suggestion
You can optionally restrict the list to a specific set of policies by supplying their IDs as trailing arguments. 
```

docs/modules/cli/pages/cerbosctl.adoc
```diff
@@ -288,6 +290,12 @@ cerbosctl get derived_roles --name-regexp "^(my|a)_policy\$"
 cerbosctl get dr --name-regexp "^(my|a)_policy\$"
 ----
 
+.List derived_roles but filter policy id common_roles.yaml and other_roles.yaml
+----
+cerbosctl get derived_roles --policy-id=common_roles.yaml,other_roles.yaml
+cerbosctl get dr --policy-id=common_roles.yaml,other_roles.yaml
```
charithe, line 296:
```suggestion
.Get derived roles policies having the ID `common_roles.yaml` and `other_roles.yaml`
----
cerbosctl get derived_roles common_roles.yaml other_roles.yaml
cerbosctl get dr common_roles.yaml other_roles.yaml
```

The other command examples need to be similarly updated as well.

charithe (commented):
There's a problem when using `inspect` with a disk store. The IDs displayed by `inspect policies` are the canonical names like `resource.leave_request.v20210210`. If you try to retrieve one by ID using `inspect policies resource.leave_request.v20210210` it fails because the store is expecting the file name as the ID.

cmd/cerbosctl/inspect/policies.go
```diff
@@ -55,7 +60,7 @@ func (c *PoliciesCmd) Run(k *kong.Kong, cctx *client.Context) error {
 
 	tw := printer.NewTableWriter(k.Stdout)
 	if !c.Format.NoHeaders {
-		tw.SetHeader([]string{"POLICY ID", "ACTIONS", "VARIABLES"})
+		tw.SetHeader([]string{"POLICY ID", "POLICY", "ACTIONS", "VARIABLES"})
```
charithe, line 63: The `POLICY` column is redundant and just takes up space so, let's not display it.

internal/inspect/policyset.go
```diff
@@ -42,8 +42,9 @@ func (ps *PolicySet) Inspect(pset *runtimev1.RunnablePolicySet) error {
 	}
 
 	ps.results[namer.PolicyKeyFromFQN(pset.Fqn)] = &responsev1.InspectPoliciesResponse_Result{
-		Actions:   actions,
-		Variables: variables,
+		Actions:         actions,
+		Variables:       variables,
+		StoreIdentifier: pset.Fqn,
```
charithe, line 47: I think this should be the policy key instead of the FQN. Remember to update `ListPolicyIDs` method of the `bundle` driver to also return policy keys instead of FQNs.

charithe (commented):
Just a couple of small things to address below. LGTM otherwise. This command should have a JSON output option as well for anyone who'd like to script something based on its output. I'll create a new issue for it.

internal/storage/hub/bundle.go
```diff
@@ -239,29 +239,29 @@ func (b *Bundle) loadPolicySet(idHex, fileName string) (*runtimev1.RunnablePolic
 }
 
 func (b *Bundle) InspectPolicies(ctx context.Context, params storage.ListPolicyIDsParams) (map[string]*responsev1.InspectPoliciesResponse_Result, error) {
-	var fqns []string
+	var policyIDs []string
 	if len(params.IDs) == 0 {
 		var err error
-		if fqns, err = b.ListPolicyIDs(ctx, params); err != nil {
+		if policyIDs, err = b.ListPolicyIDs(ctx, params); err != nil {
```
charithe, line 245: It should be `b.ListPolicyIDs` that should be changed to return the policy keys.  That'll make sure that the behaviour is consistent across the board for any existing or future code that uses `ListPolicyIDs` and does something with it.

cmd/cerbosctl/inspect/policies.go
```diff
@@ -60,7 +60,7 @@ func (c *PoliciesCmd) Run(k *kong.Kong, cctx *client.Context) error {
 
 	tw := printer.NewTableWriter(k.Stdout)
 	if !c.Format.NoHeaders {
-		tw.SetHeader([]string{"POLICY ID", "POLICY", "ACTIONS", "VARIABLES"})
+		tw.SetHeader([]string{"POLICY ID", "ACTIONS", "VARIABLES"})
```
charithe, line 63: The output should be sorted by store identifier now because it's the first column.

internal/storage/hub/bundle.go
charithe: Oh sorry, I was confused by the diff.

charithe (approved):

PR #2194, opened 2024-06-20
enhancement: More details for missing scope errors

oguzhand95:
This PR adds more details on missing scope errors.

When policy `resource.lr.v20210210/hr` doesn't exist, but `resource.lr.v20210210/hr.uk` and `resource.lr.v20210210/hr.us` exist.
```yaml
- missingPolicy: resource.lr.v20210210/hr
  descendants:
    - resource.lr.v20210210/hr.uk
    - resource.lr.v20210210/hr.us
```

In addition to changing the API response, this PR also changes the way `cerbos compile` reporting the missing scope errors:
```bash
$ cerbos compile .
Missing scopes
scoped policy resource.lr.v20210210/hr is not found but is required by descendant policies resource.lr.v20210210/hr.uk, resource.lr.v20210210/hr.uk.london, resource.lr.v20210210/hr.us, resource.lr.v20210210/hr.us.newyork
```

Playground response also reports the missing scope(s) in the same way with an error like this:
```json
{
   "playgroundId":"test",
   "failure":{
      "errors":[
         {
            "error": "Scoped policy resource.lr.v20210210/hr is not found but is required by descendant policies resource.lr.v20210210/hr.uk, resource.lr.v20210210/hr.us"
         }
      ]
   }
}
```


api/private/cerbos/runtime/v1/runtime.proto
```diff
@@ -218,7 +225,7 @@ message IndexBuildErrors {
   repeated DuplicateDef duplicate_defs = 2;
   repeated LoadFailure load_failures = 3;
   repeated MissingImport missing_imports = 4;
-  repeated string missing_scopes = 5;
+  repeated MissingScope missing_scopes = 5;
```
charithe, line 228: This would be a breaking change in other places. Let's mark it as deprecated and stop populating it but we can't drop the field or change its data type.

cmd/cerbos/compile/internal/lint/display.go
```diff
@@ -42,8 +45,11 @@ func displayList(p *printer.Printer, errs *index.BuildError) error {
 
 	if len(errs.MissingScopes) > 0 {
 		p.Println(colored.Header("Missing scopes"))
-		for _, mi := range errs.MissingScopes {
-			p.Printf("scoped policy %s is required but no definition found\n", colored.PolicyKey(mi))
+		for _, ms := range errs.MissingScopes {
+			p.Printf("scoped policy %s has missing scope(s): %s\n",
+				colored.PolicyKey(fmt.Sprintf("'%s.%s'", ms.Policy, ms.Version)),
```
charithe, line 50: So, `policy` is `kind.name`. Why is `version` stored separately?

internal/namer/namer.go
```diff
@@ -143,6 +143,15 @@ func buildFQNTree[T any](fqn, scope string, elementFn func(string) T) []T {
 	return fqnTree
 }
 
+func ScopeFromFQN(fqn string) string {
+	split := strings.Split(fqn, "/")
+	if len(split) == 1 {
+		return ""
+	}
+
+	return split[1]
```
charithe, line 152:
```suggestion
	_, scope, _ := strings.Cut(fqn, "/")
	return scope
```

charithe (commented):
The generated error message (`Scoped policy 'resource.leave_request.vdefault' has missing scope(s): '', 'acme', 'acme.hr'`) is not entirely correct for two reasons:

- It's not correct to say that the no-scope policy has missing scopes because that's not the actual policy missing an ancestor (the no-scope one can exist without any descendants). The error message should be about the child policy that has a missing ancestor in its scope chain.
- Keying by the topmost, no-scope policy is wrong because you could conceivably have multiple trees rooted at it. Consider the case where `acme.uk.london` and `acme.us.washington` exists but `acme.uk` and `acme.us` scopes don't. Those are two separate trees and need two separate error messages.

internal/storage/index/builder.go
```diff
@@ -242,9 +254,24 @@ func (idx *indexBuilder) addPolicy(file string, srcCtx parser.SourceCtx, p polic
 	idx.fileToModID[file] = p.ID
 	idx.modIDToFile[p.ID] = file
 	delete(idx.missing, p.ID)
-	delete(idx.missingScopes, p.ID)
-
 	idx.stats.add(p)
+	if ids, ok := idx.deleteFromMissingScopes[p.ID]; ok {
```
charithe, line 258: I am struggling to keep track of the juggling of entries between _four_ maps 😄 I think it can be simpler.


- Find the required ancestors for each policy passed to the `addPolicy` method
- If there are any ancestors that we haven't seen yet, add them to a map. The key is the module ID of the missing ancestor. The value is a set of FQNs that are descendants of it.
  Say you're processing `resource.foo.v1/a.b.c.d` and find that scopes `a.b` and `a.b.c` are missing. The map contents would be:
     ```go
      missingScopes["resource.foo.v1/a.b"] = map[string]struct{}{"resource.foo.v1/a.b.c.d": {}}
      missingScopes["resource.foo.v1/a.b.c"] = map[string]struct{}{"resource.foo.v1/a.b.c.d": {}}
     ```
- If there's an entry for the current policy in `missingScopes`, delete that because we know that the policy now exists
- At the end, create an error message for each item left in the `missingScopes` map. The message would be "scoped policy x is not found but is required by descendant policies a, b and c".

charithe (approved):

PR #2206, opened 2024-07-04
chore: Add PolicyWrapper protobuf

charithe:
Represent the internal wrapper as a protobuf so that it can be ported
across.


oguzhand95 (approved):
Wouldn’t converting the whole `policy.Wrapper` into `sourcev1.PolicyWrapper` and completely getting rid of `policy.Wrapper` simplify things?

Also I've opened [a PR](https://github.com/cerbos/cerbos/pull/2207) for failing tests.

charithe:
> Wouldn’t converting the whole `policy.Wrapper` into `sourcev1.PolicyWrapper` and completely getting rid of `policy.Wrapper` simplify things?

I did try but it's quite deeply ingrained in the code and changing it is not trivial.

PR #2221, opened 2024-07-16
chore: Add identity components to policy wrapper

charithe:
Include the kind, name, version, and scope in the wrapper protobuf so
that they don't need to be parsed out of the key every time.


dbuduev (approved):
LGTM

PR #2224, opened 2024-07-19
enhancement: List referenced attributes in InspectPolicies

oguzhand95:


api/public/cerbos/response/v1/response.proto
```diff
@@ -445,6 +445,17 @@ message InspectPoliciesResponse {
     json_schema: {description: "Inspect policies response"}
   };
 
+  message Attribute {
+    enum Type {
```
charithe, line 449: Let's call this `kind` or `source`. We may want to add the actual data type at some point so let's keep `type` free for that.

internal/inspect/policy.go
```diff
@@ -50,37 +52,76 @@ func (pol *Policy) Inspect(p *policyv1.Policy) error {
...
+
 	if _, ok := p.PolicyType.(*policyv1.Policy_ExportVariables); ok {
 		sortedLocalVariables := make([]*responsev1.InspectPoliciesResponse_Variable, 0, len(localVariables))
 		for _, lv := range localVariables {
 			sortedLocalVariables = append(sortedLocalVariables, lv)
 		}
 		sort.Slice(sortedLocalVariables, func(i, j int) bool {
-			return sortedLocalVariables[i].Name < sortedLocalVariables[j].Name
+			return sort.StringsAreSorted([]string{sortedLocalVariables[i].Name, sortedLocalVariables[j].Name})
 		})
```
charithe, line 67:
```suggestion
slices.SortFunc(sortedLocalVariables, func(a, b *responsev1.InspectPoliciesResponse_Variable) int {
    return cmp.Compare(a.GetName(), b.GetName())
}))
```

internal/inspect/policy.go
```diff
@@ -50,37 +52,76 @@ func (pol *Policy) Inspect(p *policyv1.Policy) error {
...
 
+		attributes := make([]*responsev1.InspectPoliciesResponse_Attribute, 0, len(referencedAttributes))
+		for _, attr := range referencedAttributes {
+			attributes = append(attributes, &responsev1.InspectPoliciesResponse_Attribute{
+				Name: attr.Name,
+				Type: attr.Type,
+			})
+		}
+		if len(attributes) > 0 {
+			sort.Slice(attributes, func(i, j int) bool {
```
charithe, line 77: Use same sorting method as above

internal/inspect/policy.go
```diff
@@ -296,19 +377,59 @@ func (pol *Policy) listImports(p *policyv1.Policy) (derivedRoleImports, variable
...
+				}
+			}
+		}
+	}
+
+	return referencedAttributes, nil
+}
+
+// inspectDefinitionsAndRules inspects the definitions and rules in the given policy to find references to the attributes, derived roles and variables.
+func (pol *Policy) inspectDefinitionsAndRules(p *policyv1.Policy) (referencedAttributes map[string]*responsev1.InspectPoliciesResponse_Attribute, referencedDerivedRoles map[string]struct{}, referencedVariables map[string]*responsev1.InspectPoliciesResponse_Variable, err error) {
```
charithe, line 414: The return values from this function are getting too complicated. It'd be better to create a struct to hold them and return that instead.

internal/inspect/policy.go
```diff
@@ -50,37 +52,76 @@ func (pol *Policy) Inspect(p *policyv1.Policy) error {
...
+				Variables:  sortedLocalVariables,
+				PolicyId:   storeIdentifier,
 			}
 		}
 
 		return nil
 	}
 
-	referencedDerivedRoles, referencedVariables, err := pol.inspectDefinitionsAndRules(p)
+	referencedAttributesFromVariables, err := pol.inspectVariableDefinitions(p)
```
charithe, line 93: Why can't this be done as part of the `inspectDefinitionsAndRules` function above?

internal/inspect/inspect_test.go
```diff
@@ -30,14 +30,15 @@ import (
 
 func TestInspect(t *testing.T) {
 	type policies struct {
-		expected               map[string]*responsev1.InspectPoliciesResponse_Result
-		expectedMissingImports []string
+		expected map[string]*responsev1.InspectPoliciesResponse_Result
```
charithe, line 33: I think these tests are getting quite complicated and hard to follow. For example, the helper functions like `attributes` and `result` have so many parameters and you need to keep looking at their signatures to figure out what each one is. These  tests are a good candidate for moving into file-based tests like we do with engine, compile and server tests.

internal/inspect/policy.go
charithe, line 1: The `Inspect` function is getting too big and has too many responsibilities crammed into it. It should be split into smaller functions. For example, a function for dealing with derived roles, another for variables and one more for attributes.

internal/inspect/policy.go
```diff
@@ -229,8 +238,25 @@ func (pol *Policy) Results(ctx context.Context, loadPolicy loadPolicyFn) (map[st
...
+						referencedAttributes := make(map[string]*responsev1.InspectPoliciesResponse_Attribute)
+						if err := pol.referencedAttributesInExpr(importedVariable.Value, referencedAttributes); err != nil {
+							return fmt.Errorf("failed to find referenced attributes in imported variable: %w", err)
+						}
+
+						if len(referencedAttributes) > 0 {
+							ss := make(util.StringSet)
+							for _, attr := range pol.results[policyID].Attributes {
+								ss[attr.Name] = struct{}{}
+							}
```
charithe, line 252: Isn't this set created over and over for every imported variable that references an attribute? I suppose there won't be that many in practice but there could be degenerate cases.

charithe (commented):
Nearly there. I'd also split the `Results` method into `resolveDerivedRoles`, `resolveVariables` and so on because it's a massive function doing a lot of things at once.

internal/inspect/policy.go
```diff
@@ -209,6 +216,19 @@ func (pol *Policy) Results(ctx context.Context, loadPolicy loadPolicyFn) (map[st
...
+
+							referencedAttributes := make(map[string]*responsev1.InspectPoliciesResponse_Attribute)
+							if err := pol.referencedAttributesInExpr(importedVariable.Value, referencedAttributes); err != nil {
+								return fmt.Errorf("failed to find referenced attributes in imported variable: %w", err)
+							}
+
+							if len(referencedAttributes) > 0 {
+								for _, attr := range referencedAttributes {
+									if !attrs.Contains(attr.Name) {
+										pol.results[policyID].Attributes = append(pol.results[policyID].Attributes, attr)
```
charithe, line 228: Shouldn't this new attribute be added to the `attrs` map?

internal/inspect/policy.go
```diff
@@ -229,8 +249,20 @@ func (pol *Policy) Results(ctx context.Context, loadPolicy loadPolicyFn) (map[st
...
+
+						referencedAttributes := make(map[string]*responsev1.InspectPoliciesResponse_Attribute)
+						if err := pol.referencedAttributesInExpr(importedVariable.Value, referencedAttributes); err != nil {
+							return fmt.Errorf("failed to find referenced attributes in imported variable: %w", err)
+						}
+
+						if len(referencedAttributes) > 0 {
+							for _, attr := range referencedAttributes {
+								if !attrs.Contains(attr.Name) {
+									pol.results[policyID].Attributes = append(pol.results[policyID].Attributes, attr)
```
charithe, line 262: Same here?

charithe (approved):

PR #2231, opened 2024-07-23
chore(tooling): Set image tag for redocly/cli

oguzhand95:


.justfile
```diff
@@ -12,6 +12,8 @@ tools_mod_dir := join(justfile_directory(), "tools")
 
 export TOOLS_BIN_DIR := join(env_var_or_default("XDG_CACHE_HOME", join(env_var("HOME"), ".cache")), "cerbos/bin")
 
+redocly_cli_image_tag := '1.18.1'
```
charithe, line 15: Variables are sorted so this should be on line 8. However, I don't think we need a variable for the version. We can make it a parameter of the `generate-api-docs` target.

```
generate-api-docs REDOCLY_VERSION='1.18.1':
  ...
```

charithe (approved):

PR #2233, opened 2024-07-24
feat: Role policy derived roles

Sambigeara:
TODO (this is a WIP)


Sambigeara:
Hi @charithe @haines, the more I progress through this work, the more I question whether or not the gymnastics required to get derived role policies to fit are justified, so I wondered if I could get a premature view on whether or not you think this is a viable approach.

### Reminder of the problem we're solving

It all comes down to the policy lookup/retrieval from the index at evaluation time. The way we do it at present is by constructing module IDs from certain attributes (e.g. role name and scope for role policies) and using these to retrieve the mapped file names from the index. Policies are extracted from these files (from within the index file system), compiled, and returned to the evaluator.

Derived role policies break this model because we now have a concept of a (for lack of a better term) "virtual" role policy. Consider a file system where the only role policy we have is a derived role policy:

```yaml
-- derived_roles/p.yaml --
---
apiVersion: api.cerbos.dev/v1
derivedRoles:
  name: p
  definitions:
    - name: foo
      parentRoles: [user, admin]
      condition:
        match:
            expr: R.attr.foo == 42

-- role_policies/a.yaml --
---
apiVersion: api.cerbos.dev/v1
rolePolicy:
  derivedRole: p:foo
  rules:
    - resource: widget
      permissibleActions: [create, read]
```

The user doesn't need to define role policies for the referenced `parentRoles` above: `[user, admin]`, but for the purposes of the runnable policy retrieval at evaluation time, we need two target role policies available in the index generated for each of those (principal) role policies. We discussed this and the required form of the runnable role policy sets [previously](https://github.com/cerbos/cerbos-private/issues/1164#issuecomment-2189950493).

### How I've done it

The meat of the logic is in [this method](https://github.com/Sambigeara/cerbos/blob/feat/role-policy-derived-roles/internal/storage/index/index.go#L85-L85). It's run immediately after the initial file system traversal and builds out a central map of principal roles to compilation units ("virtual" or otherwise). We need to merge the compilation units so a derived role policy which (indirectly) references a principal role doesn't override any explicitly written principal role policies.

There's then just a [simple if/else](https://github.com/Sambigeara/cerbos/blob/feat/role-policy-derived-roles/internal/storage/index/index.go#L355-L357) in the `index.GetCompilationUnits` method which looks in this map first before continuing down the existing evaluation path. The compilation units have some additional data added which is required for constructing the runnable policy sets.

### Is this a path we want to go down?

This PR is currently working, and all tests passing. There's some more TODOs and refinements, and some tests to write, but otherwise it's in a decent state.

However, I had a catch up with Charith and we touched briefly on some broader ideas around changes to our indexing mechanism which could ultimately make this change (and other future changes) more natural to implement. With this in mind, is it worth adding this feature as is or should we hold off in lieu of better fitting supporting components?

PR #2243, opened 2024-07-30
fix: Reduce memory usage in badger stream for hub audit backend

Sambigeara:
We observe high memory usage in Cerbos instances which have a `hub` audit backend configured. This is an attempt to reduce memory usage by lazily evaluating buffers produced by the DB scan, and by reusing types which we use as unmarshal targets.

### DIFF
--- a/internal/audit/hub/hub.go
+++ b/internal/audit/hub/hub.go
@@ -287,8 +287,12 @@ func (l *Log) streamPrefix(ctx context.Context, kind logsv1.IngestBatch_EntryKin
 				return err
 			}
 
-			clear(keys[i])
-			keys[i] = kv.Key
+			if keys[i] == nil {
+				keys[i] = make([]byte, len(kv.Key))
+			} else {
+				clear(keys[i])
+			}
+			copy(keys[i], kv.Key)
 
 			i++
 			if i == l.maxBatchSize {
--- a/internal/audit/hub/hub.go
+++ b/internal/audit/hub/hub.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	badgerv4 "github.com/dgraph-io/badger/v4"
+	bpb "github.com/dgraph-io/badger/v4/pb"
 	"github.com/dgraph-io/ristretto/z"
 	"github.com/sourcegraph/conc/pool"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


internal/audit/hub/hub.go
```diff
@@ -252,29 +255,45 @@ func (l *Log) streamPrefix(ctx context.Context, kind logsv1.IngestBatch_EntryKin
...
+			kv.Reset()
+
+			if err := kv.Unmarshal(s); err != nil {
+				return err
 			}
 
-			logger.Log(zapcore.Level(-3), "Syncing and deleting batch")
-			if err := l.syncThenDelete(ctx, kind, keys[i:end]); err != nil {
-				return fmt.Errorf("failed to sync and delete logs: %w", err)
+			keys[i] = append([]byte(nil), kv.Key...)
```
charithe, line 280: Aren't we losing a little bit of the pool advantage by allocating a new byte slice here for each key? I think we have two options here.
- Take a new KV from the pool for each item until we reach the batch size limit, sync those and then release all of them back to the pool
- Have a second pool holding `[l.maxBatchSize][]byte` objects. [`clear`](https://pkg.go.dev/builtin#clear) then  [`copy`](https://pkg.go.dev/builtin#copy)  the key into each entry. (We might not even need the `clear` because keys are fixed size but it's probably better to be safe.)

internal/audit/hub/hub.go
```diff
@@ -268,16 +271,24 @@ func (l *Log) streamPrefix(ctx context.Context, kind logsv1.IngestBatch_EntryKin
...
 		if err := buf.SliceIterate(func(s []byte) error {
 			kv.Reset()
 
 			if err := kv.Unmarshal(s); err != nil {
 				return err
 			}
 
-			keys[i] = append([]byte(nil), kv.Key...)
+			clear(keys[i])
+			keys[i] = kv.Key
```
charithe, line 291: Doesn't this assign (and replace) the array entry with `kv.Key`? I think it has to be `copy(keys[i], kv.Key)`.

internal/audit/hub/hub.go
```diff
@@ -268,16 +271,24 @@ func (l *Log) streamPrefix(ctx context.Context, kind logsv1.IngestBatch_EntryKin
...
-		keys := make([][]byte, l.maxBatchSize)
 		if err := buf.SliceIterate(func(s []byte) error {
 			kv.Reset()
 
 			if err := kv.Unmarshal(s); err != nil {
 				return err
 			}
 
-			keys[i] = append([]byte(nil), kv.Key...)
+			clear(keys[i])
```
charithe, line 290: If this is a newly created object from the pool `keys[i]` would be nil and would require instantiation first.

```suggestion
if keys[i] == nil {
  keys[i] = make([]byte, len(kv.Key))
} else {
  clear(keys[i])
}
```

internal/audit/hub/hub.go
Sambigeara: 🤦‍♂️

PR #2248, opened 2024-07-31
chore: Remove callback from local audit log

charithe:
The callback is in the write path and blocks until the Hub log is done
with its work using its internal timer. By removing this dependency, we
can let the two logs work independently. Even if the Hub sync falls
behind and adds delays to the internal timer, the local log saving
thread can continue to work without being blocked.

The down side of this change is that tests are a bit more flaky.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -186,11 +186,11 @@ require (
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang-sql/civil v0.0.0-20220223132316-b832511892a9 // indirect
 	github.com/golang-sql/sqlexp v0.1.0 // indirect
-	github.com/golang/glog v1.2.1 // indirect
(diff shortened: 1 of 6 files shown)


Sambigeara:
The callback only blocks on shut-down, not in normal operation (in an attempt to flush any logs that might still make it up to hub).

Triggering the hub sync immediately after local batch writes (or at least, on the ticker) means that we'd effectively continue to do house keeping on any written local logs as they came in. During normal operation, it runs the hub sync in a separate goroutine (and it's the responsibility of the callback function in this case to deal with any concurrency issues, which is _attempts_ to do via the mutexTimer).

If the hub sync cycle were to run independently, I feel like we'd have less predictable syncs with a higher number of no-ops 🤔

Sambigeara (approved):
👍

PR #2260, opened 2024-08-07
feat: Role policies (#2192)

Sambigeara:
Introduce a new Role policy type, which specifies a list of resources along with permissible actions for each. Role policies are evaluated prior to resource policies, and cannot override them.

I'll delete the feature branch separately. PR'ing from here negates the need to bypass merge rules again, and feels more "proper" 🤠


Sambigeara:
This is the same change as in the main repo feature branch, but I've squashed the two commits into one.

charithe (approved):

PR #2263, opened 2024-08-08
enhancement: Atomic upgrades for blob storage

oguzhand95:
WIP

Problems
---

- Subdirectories left at the workDir after deleting a directory in the bucket.
- Readability decreased a lot.


internal/storage/blob/conf.go
```diff
@@ -35,6 +35,8 @@ type Conf struct {
 	Bucket string `yaml:"bucket" conf:"required,example=\"s3://my-bucket-name?region=us-east-2\""`
 	// Prefix specifies a subdirectory to download.
 	Prefix string `yaml:"prefix,omitempty" conf:",example=policies"`
+	// CacheDir is the local path to keep the downloaded files.
+	CacheDir string `yaml:"cacheDir" conf:",example=${HOME}/tmp/cerbos/cache"`
```
charithe, line 39: We don't need a second directory. It would be a breaking change as well. Simply use the `WorkDir` and create two subdirectories under it. `.cache` for the files stored as etags and `store` for the symlinks.

internal/storage/blob/store.go
```diff
@@ -305,6 +386,96 @@ func (s *Store) pollForUpdates(ctx context.Context) {
...
 
+func (s *Store) createSymlink(destination, source string) error {
+	if _, err := os.Lstat(source); err == nil {
+		if err := os.Remove(source); err != nil {
+			return fmt.Errorf("failed to delete left-over symlink at %q: %w", source, err)
+		}
+	}
+
+	//nolint:mnd
+	if err := os.MkdirAll(filepath.Dir(source), 0o744); err != nil {
```
charithe, line 397: We don't necessarily need to mirror the directory layout of the remote bucket and could just create the symlinks in a flat structure. That would simplify things for us but, the error messages would be confusing for the users to interpret 🤔

charithe (commented):
I think the way to deal with deletions etc. is to use a double-buffering system. 
- Create a directory inside the work directory to hold the current set of symlinks. We can name it using the current timestamp. (Let's call it `A`.)
- On next sync, create another directory using the current timestamp (Let's call that `B`). Create the new symlinks inside that directory.
- If the index build is successful, switch to `B` and delete `A`. 
- Repeat for each sync

internal/storage/blob/store.go
oguzhand95: Wouldn't that break when there are two files with the same name? (such as `dir/some.yaml` and `some.yaml`)

charithe: I was thinking of using the full path and replacing the path separator with another character. So, you'd have `dir|some.yaml` and `some.yaml`. It's probably too much though.

internal/storage/blob/store.go
```diff
@@ -65,11 +60,16 @@ func init() {
 			return nil, err
 		}
 
-		if err := validateOrCreateWorkDir(conf.WorkDir); err != nil {
-			return nil, err
+		if err := validateOrCreateDir(filepath.Join(conf.WorkDir, dotcache)); err != nil {
+			return nil, fmt.Errorf("failed to create dotcache directory: %w", err)
```
charithe, line 64:
```suggestion
			return nil, fmt.Errorf("failed to create cache directory: %w", err)
```

internal/storage/blob/store.go
```diff
@@ -65,11 +60,16 @@ func init() {
...
 		}
 
-		c, err := NewCloner(bucket, storeFS{dir: conf.WorkDir})
+		if err := validateOrCreateDir(conf.WorkDir); err != nil {
```
charithe, line 67: Isn't this already created by the step above?

internal/storage/blob/store.go
```diff
@@ -214,70 +232,67 @@ func (s *Store) init(ctx context.Context) error {
...
 
-	if changes.isEmpty() {
+	if cr.isEmpty() {
 		s.log.Debug("No changes")
 		return nil
 	}
 
-	s.log.Infof("Detected changes: added or updated (%d), deleted (%d)", len(changes.updateOrAdd), len(changes.delete))
+	s.log.Infof("Detected changes: added or updated (%d), deleted (%d)", len(cr.addedOrUpdated), len(cr.deleted))
+	dotCacheDir := filepath.Join(s.conf.WorkDir, dotcache)
```
charithe, line 248: You can probably store this as a field in the struct because it's used in many places.

internal/storage/blob/store.go
```diff
@@ -214,70 +232,67 @@ func (s *Store) init(ctx context.Context) error {
...
-			return err
-		}
-		entry := index.Entry{File: f.file, Policy: policy.Wrap(policy.WithSourceAttributes(p, driverSourceAttr, etagSourceAttr(f.etag)))}
-		event, err = s.idx.AddOrUpdate(entry)
-		if err != nil {
-			return err
+	idx, err := index.Build(ctx, os.DirFS(workDir), index.WithRootDir("."), index.WithSourceAttributes(driverSourceAttr))
+	if err != nil {
+		if rerr := os.RemoveAll(workDir); rerr != nil && !os.IsNotExist(rerr) {
+			return fmt.Errorf("failed to remove the directory %s: %w", s.workDir, errors.Join(rerr, err))
```
charithe, line 265: This part is nearly identical to initial clone above and is kind of critical for correct functioning. I think it'd be better to define it as a reusable function if possible.

internal/storage/blob/cloner.go
```diff
@@ -20,68 +19,57 @@ import (
...
 	fs.FS
 	Remove(name string) error
 	Create(name string) (io.WriteCloser, error)
 	MkdirAll(path string, perm fs.FileMode) error
 }
 
-type infoType map[string][]byte
+type (
+	etagType      string
+	fileNameType  string
```
charithe, line 31: These types are not quite necessary because we are not passing these values around to other functions.

charithe (commented):
LGTM 👍🏽 

Don't forget to notify subscribers about the changes as well.

oguzhand95:
After wondering why the hashes change now and then, I realized it was due to `protojson.Marshal`. It bit me before but I've forgotten. (See [unstable json](https://protobuf.dev/reference/go/faq/#unstable-json) section)

I'd say the PR is ready for review. However, there is a problem with the `TestReloadable` where it fails sporadically. I tried a couple of things to understand why it failed to no avail, but will dig deeper.

internal/storage/blob/cloner.go
```diff
@@ -91,45 +76,83 @@ func (c *Cloner) Clone(ctx context.Context) (*CloneResult, error) {
...
+		all:            all,
+		addedOrUpdated: addedOrUpdated,
+		deleted:        deleted,
+		failuresCount:  failuresCount,
+	}, nil
 }
 
-func (c *Cloner) downloadToFile(ctx context.Context, key, file string) (err error) {
-	// Create the directories in the path
+func (c *Cloner) downloadToFile(ctx context.Context, key, file string) error {
```
charithe, line 150:
```suggestion
func (c *Cloner) downloadToFile(ctx context.Context, key, file string) (err error) {
```

Removing the output variable makes the `multierr.AppendInvoke` on line 160 not have an effect.

internal/storage/blob/store.go
```diff
@@ -189,95 +185,133 @@ func NewStore(ctx context.Context, conf *Conf, cloner bucketCloner) (*Store, err
...
 		if err != nil {
-			return err
+			return fmt.Errorf("faield to create delete event: %w", err)
 		}
-		s.NotifySubscribers(event)
+		evts = append(evts, e)
+	}
+
+	if err := s.cloner.Clean(); err != nil {
+		return fmt.Errorf("failed to clean up the cache: %w", err)
```
charithe, line 268: We should log this as a warning but I don't think we need to fail the whole thing by returning the error because failing to clean up the files doesn't affect the correct functioning of the store.

internal/storage/blob/store.go
```diff
@@ -189,95 +185,133 @@ func NewStore(ctx context.Context, conf *Conf, cloner bucketCloner) (*Store, err
...
 	}
 
+	s.NotifySubscribers(evts...)
 	s.log.Info("Index updated")
+
+	if err := os.RemoveAll(oldWorkDir); err != nil {
+		return fmt.Errorf("failed to remove old work directory %s: %w", s.workDir, err)
```
charithe, line 275: This too can be just a warning log message.

internal/storage/blob/store.go
```diff
@@ -295,8 +329,14 @@ func (s *Store) pollForUpdates(ctx context.Context) {
 			s.log.Info("Stopped polling for updates")
 			return
 		case <-ticker.C:
-			if err := s.updateIndex(ctx); err != nil {
+			err := s.updateIndex(ctx)
+			buildErr := &indexBuildError{}
+			if err != nil {
 				s.log.Errorw("Failed to check for updates", "error", err)
```
charithe, line 335: Move this to the `else` block where we check whether the error is an index build error.

internal/storage/blob/store.go
```diff
@@ -295,8 +329,14 @@ func (s *Store) pollForUpdates(ctx context.Context) {
...
+				if errors.As(err, &buildErr) {
+					s.log.Warnf("Failed to build index from the new set of files and will keep using the existing index at %s", s.workDir)
```
charithe, line 337:
```suggestion
                    s.log.Warnf("Remote store is in an invalid state", "error", err)
					s.log.Warnf("Remote store is in an invalid state. Using the last good state from %s", s.workDir)
```

internal/storage/blob/store.go
```diff
@@ -305,6 +345,76 @@ func (s *Store) pollForUpdates(ctx context.Context) {
...
+	return nil
+}
+
+func (s *Store) createSymlinks(all map[string][]string, destDir, sourceDir string) error {
+	for etag, files := range all {
+		for _, file := range files {
+			source := filepath.Join(sourceDir, file)
+			dest := filepath.Join(destDir, etag)
+			if err := s.createSymlink(dest, source); err != nil {
+				return fmt.Errorf("failed to create symbolic links for the cloned file: %w", err)
```
charithe, line 375:
```suggestion
				return fmt.Errorf("failed to create symbolic link for %s: %w", source, err)
```

internal/storage/blob/store.go
```diff
@@ -305,6 +345,76 @@ func (s *Store) pollForUpdates(ctx context.Context) {
...
+}
+
+type indexBuildError struct {
+	err error
+	// workDir is the temporary work directory which the store tried build an index from the new set of files
+	workDir string
+}
+
+func (e *indexBuildError) Error() string {
+	return fmt.Sprintf("failed to build index at temporary work directory at %s: %v", e.workDir, e.err)
```
charithe, line 390:
```suggestion
	return fmt.Sprintf("failed to build index at temporary work directory at %s: %w", e.workDir, e.err)
```

internal/storage/blob/store.go
```diff
@@ -305,6 +345,76 @@ func (s *Store) pollForUpdates(ctx context.Context) {
...
+	}
+
+	if err := s.createSymlinks(all, s.cacheDir, newWorkDir); err != nil {
+		return nil, "", fmt.Errorf("failed to create symbolic links for the new work directory: %w", err)
+	}
+
+	idx, err := index.Build(ctx, os.DirFS(newWorkDir), index.WithRootDir("."), index.WithSourceAttributes(driverSourceAttr))
+	if err != nil {
+		if rerr := os.RemoveAll(newWorkDir); rerr != nil && !os.IsNotExist(rerr) {
+			return nil, "", fmt.Errorf("failed to remove the directory %s while cleaning up new work directory: %w", newWorkDir, errors.Join(rerr, err))
```
charithe, line 409:
```suggestion
			return nil, "", errors.Join(err,  fmt.Errorf("failed to remove directory %s: %w", newWorkDir, rerr))
```

internal/storage/blob/store.go
```diff
@@ -350,21 +460,38 @@ func (s *Store) RepoStats(ctx context.Context) storage.RepoStats {
...
+	cr, err := s.clone(ctx)
 	if err != nil {
-		return fmt.Errorf("failed to clone: %w", err)
+		return fmt.Errorf("failed to clone blob store: %w", err)
 	}
 
-	if failures := changes.failures(); failures > 0 {
-		logging.ReqScopeLog(ctx).Warn(fmt.Sprintf("Failed to download (%d) files", failures))
+	if cr.failuresCount > 0 {
+		logging.ReqScopeLog(ctx).Warn(fmt.Sprintf("Failed to download (%d) files", cr.failuresCount))
```
charithe, line 469: I think we should return an error here and exit because the download failures are quite likely to make the index invalid anyway.

internal/storage/blob/store.go
```diff
@@ -350,21 +460,38 @@ func (s *Store) RepoStats(ctx context.Context) storage.RepoStats {
...
 	}
 
-	evts, err := s.idx.Reload(ctx)
+	idx, newWorkDir, err := s.buildIndex(ctx, cr.all)
+	buildErr := &indexBuildError{}
 	if err != nil {
-		return fmt.Errorf("failed to reload the index: %w", err)
+		if errors.As(err, &buildErr) {
+			s.log.Warnf("Failed to build index from the new set of files and will keep using the existing index at %s", s.workDir)
```
charithe, line 476:
```suggestion
            s.log.Warnf("Remote store state is invalid", "error", err)
			s.log.Warnf("Remote store state is invalid. Using the last good state from %s", s.workDir)
```

internal/storage/blob/store.go
```diff
@@ -350,21 +460,38 @@ func (s *Store) RepoStats(ctx context.Context) storage.RepoStats {
...
+		}
+
+		return fmt.Errorf("failed to build index from the new set of files: %w", err)
```
charithe, line 479:
```suggestion
		return fmt.Errorf("failed to reload state from remote store: %w", err)
```

internal/storage/blob/store.go
```diff
@@ -350,21 +460,38 @@ func (s *Store) RepoStats(ctx context.Context) storage.RepoStats {
...
 
-	s.NotifySubscribers(evts...)
+	oldWorkDir := s.workDir
+	s.workDir = newWorkDir
+	s.workFS = os.DirFS(newWorkDir)
+	s.idx = idx
+	s.NotifySubscribers(storage.NewReloadEvent())
+
+	if err := os.RemoveAll(oldWorkDir); err != nil {
+		return fmt.Errorf("failed to remove old work directory %s: %w", s.workDir, err)
```
charithe, line 489: This can be just logged. No need to return the error.

internal/storage/blob/store_test.go
```diff
@@ -84,6 +91,89 @@ func TestReloadable(t *testing.T) {
...
+	}
+	conf.SetDefaults()
+
+	must := require.New(t)
+
+	bucket, err := newBucket(ctx, conf)
+	must.NoError(err)
+	_, err = NewStore(ctx, conf, NewCloner(bucket, storeFS{filepath.Join(dir, dotcache)}))
+	must.NoError(err)
+}
```
charithe, line 175: Let's remove this because we never use it.

charithe (commented):
👍🏽  

For future reference: you can use [`pretty.UglyInPlace`](https://github.com/cerbos/cerbos/blob/f4afc44ff78ccf8b39c84240d0d4241579d3d6f0/internal/verify/junit/junit.go#L207) to make the output from `protojson.Marshal` deterministic.

charithe:
Let's also add a new gauge metric that stores the last time the store was successfully updated from the remote bucket. You can do that in a separate PR if you like.

internal/storage/blob/store.go
oguzhand95: Does `%w` work with `fmt.Sprintf`?

charithe: Ah! I misread it as `fmt.Errorf` 🤦🏽

internal/storage/blob/store.go
oguzhand95: Now that I think about it, I think I will get rid of `failuresCount` all together.
If there are any type of error during clone, we should abort to keep things atomic hence we do not require any state for failures IMO.

internal/storage/blob/store.go
```diff
@@ -264,15 +251,15 @@ func (s *Store) updateIndex(ctx context.Context) error {
...
-		return fmt.Errorf("failed to clean up the cache: %w", err)
-	}
-
 	s.NotifySubscribers(evts...)
 	s.log.Info("Index updated")
 
-	if err := os.RemoveAll(oldWorkDir); err != nil {
-		return fmt.Errorf("failed to remove old work directory %s: %w", s.workDir, err)
+	if err := s.cloner.Clean(); err != nil {
+		s.log.Warnf("Failed to clean up the cache: %v", err)
```
charithe, line 258:
```suggestion
        s.log.Warnw("Failed to clean up the cache", "error", err)
```

internal/storage/blob/store.go
```diff
@@ -264,15 +251,15 @@ func (s *Store) updateIndex(ctx context.Context) error {
...
+	}
+
+	if err := os.RemoveAll(oldWorkDir); err != nil && !os.IsNotExist(err) {
+		s.log.Warnf("failed to remove old work directory %s: %v", oldWorkDir, err)
```
charithe, line 262:
```suggestion
        s.log.Warnw(fmt.Sprintf("failed to remove old work directory %s", oldWorkDir), "error", err)
```

internal/storage/blob/store.go
```diff
@@ -485,8 +469,12 @@ func (s *Store) Reload(ctx context.Context) error {
 	s.idx = idx
 	s.NotifySubscribers(storage.NewReloadEvent())
 
-	if err := os.RemoveAll(oldWorkDir); err != nil {
-		return fmt.Errorf("failed to remove old work directory %s: %w", s.workDir, err)
+	if err := s.cloner.Clean(); err != nil {
+		s.log.Warnf("Failed to clean up the cache: %v", err)
```
charithe, line 473:
```suggestion
        s.log.Warnw("Failed to clean up the cache", "error", err)
```

internal/storage/blob/store.go
```diff
@@ -485,8 +469,12 @@ func (s *Store) Reload(ctx context.Context) error {
...
+	}
+
+	if err := os.RemoveAll(oldWorkDir); err != nil && !os.IsNotExist(err) {
+		s.log.Warnf("failed to remove old work directory %s: %v", oldWorkDir, err)
```
charithe, line 477:
```suggestion
        s.log.Warnw(fmt.Sprintf("Failed to remove work directory %s", oldWorkDir), "error", err)
```

charithe:
Because the metric is called last index update timestamp, we should update it in all the places where the index is updated -- not just in the blob driver. Let's stash that commit and create a separate PR for it.

charithe (dismissed):

charithe (approved):

Issue #2280, opened 2024-08-13
Gauge metric for reporting the last store refresh timestamp


charithe:
Report the last time the store was successfully refreshed either manually (store reload), push (Cerbos Hub) or polling (disk, git, blob).

PR #2284, opened 2024-08-16
enhancement: Change how blob storage creates work directories and add metric for the last store refresh

oguzhand95:
Resolves cerbos/cerbos#2280

### DIFF
--- a/internal/storage/blob/store.go
+++ b/internal/storage/blob/store.go
@@ -323,19 +323,14 @@ func (s *Store) pollForUpdates(ctx context.Context) {
 			s.log.Info("Stopped polling for updates")
 			return
 		case <-ticker.C:
-			err := s.updateIndex(ctx)
-			if err != nil {
+			if err := s.updateIndex(ctx); err != nil {
 				if errors.Is(err, &indexBuildError{}) {
 					s.log.Warnw("Remote store is in an invalid state", "error", err)
 					s.log.Warnf("Remote store is in an invalid state. Using the last good state from %s", s.workDir)
 				} else {
 					s.log.Warnw("Failed to check for updates", "error", err)
 				}
-
-				metrics.Inc(ctx, metrics.StoreSyncErrorCount(), metrics.DriverKey(DriverName))
 			}
-
-			metrics.Inc(ctx, metrics.StorePollCount(), metrics.DriverKey(DriverName))
 		}
 	}
 }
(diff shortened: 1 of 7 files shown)


internal/storage/blob/store.go
```diff
@@ -404,7 +399,16 @@ func (e *indexBuildError) Error() string {
 // s.cacheDir according to the given map 'all' and tries to build a temporary index to see if there are any errors
 // with the incoming policies/schemas. If there are no errors returns the index built and the path to the new work directory.
 func (s *Store) buildIndex(ctx context.Context, all map[string][]string) (idx index.Index, ts string, err error) {
-	ts = strconv.FormatInt(time.Now().UnixMicro(), 10)
+	tsMicro := time.Now().UnixMicro()
```
charithe, line 402: I think that microsecond granularity is a bit too much 😄 `UnixMillis` should be more than enough.

charithe (commented):
Even the DB stores have a `Reload` method so we should update the metric from there as well.

internal/storage/blob/store.go
charithe: Does that test compare the timestamp used for this metric? I assume the problem is that the store directory is based on Unix timestamp? We should maybe add a random value as a suffix to the directory name to avoid those types of problems.

oguzhand95: Yes, the problem occurs when insufficient time has passed and the same (timestamp) directory is used for the next update.

I will implement your suggestion.

internal/storage/disk/dirwatch.go
```diff
@@ -110,18 +110,19 @@ func (dw *dirWatch) processEvent(evtInfo notify.EventInfo) {
 
 func (dw *dirWatch) triggerUpdate() {
 	dw.mu.RLock()
-	shouldUpdate := len(dw.eventBatch) > 0 && (time.Since(dw.lastEventTime) > dw.cooldownPeriod)
+	shouldUpdate := len(dw.eventBatch) > 0 && (time.Since(dw.lastEventTime).Microseconds() > dw.cooldownPeriod.Microseconds())
```
charithe, line 113: Do we need to get microseconds here?

internal/storage/disk/dirwatch.go
```diff
@@ -110,18 +110,19 @@ func (dw *dirWatch) processEvent(evtInfo notify.EventInfo) {
...
 	if shouldUpdate {
 		dw.mu.Lock()
-		proceed := len(dw.eventBatch) > 0 && (time.Since(dw.lastEventTime) > dw.cooldownPeriod)
+		proceed := len(dw.eventBatch) > 0 && (time.Since(dw.lastEventTime).Microseconds() > dw.cooldownPeriod.Microseconds())
 		if !proceed {
 			dw.mu.Unlock()
 			return
 		}
 
+		ts := dw.lastEventTime
```
charithe, line 125: I think the `ts` should be `time.Now()` because the actual store refresh happened just now.

internal/storage/blob/store.go
```diff
@@ -403,28 +378,40 @@ func (e *indexBuildError) Error() string {
...
+		if err != nil {
+			metrics.Inc(ctx, metrics.StoreSyncErrorCount(), metrics.DriverKey(DriverName))
+		} else {
+			metrics.Record(ctx, metrics.StoreLastSuccessfulRefresh(), ts, metrics.DriverKey(DriverName))
+		}
+	}()
+
+	dir, err := os.MkdirTemp(s.workDir, fmt.Sprintf("cerbos-%s-%d-*", DriverName, ts))
+	if err != nil {
+		return nil, "", 0, fmt.Errorf("failed to create new work directory with timestamp %d: %w", ts, err)
```
charithe, line 394:
```suggestion
		return nil, "", 0, fmt.Errorf("failed to create new temporary storage directory: %w", err)
```

charithe (approved):

Issue #2281, opened 2024-08-14
Reuse existing cache for blob store


charithe:
After #2263, if the work directory is on persistent storage, the cache directory would already contain the previously downloaded files. However, because the state of the cache is maintained in memory, the PDP would re-download the files every time it starts up. We can avoid this during the download step by first checking whether the file to be downloaded already exists.

There is an edge cases in this approach. If the user changes the configuration (e.g. point to a different bucket) between restarts, the existing cache might be invalid. In most cases this shouldn't be an issue because if the etag matches, it's quite likely to be the same file. However, since etags use MD5 and it is a weak algorithm, there could be cases where two different files have the same etag. We can't make this 100% foolproof because etags are all we have. One way to make it a bit more robust is to store the files in the cache under a directory named after the full bucket URL. For example, if the bucket URL is `gcs://my-policy-bucket`, the cache directory would be `<workDir>/.cache/base64encode("gcs://my-policy-bucket")`.

PR #2292, opened 2024-08-23
enhancement: Keep cached files under base64 encoded directory for blob storage

oguzhand95:
resolves cerbos/cerbos#2281

### DIFF
--- a/internal/storage/blob/cloner_test.go
+++ b/internal/storage/blob/cloner_test.go
@@ -6,7 +6,6 @@ package blob
 import (
 	"bytes"
 	"context"
-	"encoding/base64"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


internal/storage/blob/cloner_test.go
```diff
@@ -34,8 +35,7 @@ func TestCloneResult(t *testing.T) {
 
 		bucketDir := filepath.Join(dir, "bucket")
 		require.NoError(t, os.MkdirAll(bucketDir, perm775))
-
-		cacheDir := filepath.Join(dir, dotcache)
+		cacheDir := filepath.Join(dir, dotcache, base64.URLEncoding.EncodeToString([]byte(bucketDir)))
```
charithe, line 38: It's worth creating a function to generate the cache directory name given the work directory and the bucket URL. We are currently repeating the logic in multiple places and they can slowly drift apart with time.

charithe (dismissed):

charithe (approved):

PR #2301, opened 2024-08-29
enhancement: Display attributes in the cerbosctl inspect policies command

oguzhand95:
Displaying attributes in the existing command `cerbosctl inspect policies` seems to work visually.

```
$ cerbosctl inspect policies
POLICY ID : derived_roles/common_roles.yaml
ACTIONS   : 
ATTRIBUTES: R.attr.flagged,R.attr.owner
VARIABLES : 
-------------------
POLICY ID : resource_policies/policy_05_acme.hr.uk.yaml
ACTIONS   : defer,delete
ATTRIBUTES: R.attr.geography
VARIABLES : principal_location
-------------------
```

```
$ cerbosctl inspect policies --no-headers
derived_roles/common_roles.yaml

R.attr.flagged,R.attr.owner

-------------------
resource_policies/policy_05_acme.hr.uk.yaml
defer,delete                                                
R.attr.geography                                                      
principal_location  
-------------------
```

### DIFF
--- a/cmd/cerbosctl/inspect/internal/print.go
+++ b/cmd/cerbosctl/inspect/internal/print.go
@@ -14,6 +14,7 @@ import (
 	responsev1 "github.com/cerbos/cerbos/api/genpb/cerbos/response/v1"
 	"github.com/cerbos/cerbos/cmd/cerbosctl/inspect/internal/flagset"
 	"github.com/cerbos/cerbos/cmd/cerbosctl/internal/printer"
+	"github.com/cerbos/cerbos/internal/conditions"
 	"github.com/cerbos/cerbos/internal/util"
 )
 
@@ -88,10 +89,22 @@ func printJSON(w io.Writer, results []*responsev1.InspectPoliciesResponse_Result
 func printTable(w io.Writer, noHeaders bool, results []*responsev1.InspectPoliciesResponse_Result) {
 	tw := printer.NewTableWriter(w)
 	if !noHeaders {
-		tw.SetHeader([]string{"POLICY ID", "ACTIONS", "VARIABLES"})
+		tw.SetHeader([]string{"POLICY ID", "ACTIONS", "ATTRIBUTES", "VARIABLES"})
 	}
 
 	for _, result := range results {
+		attributes := make([]string, len(result.Attributes))
+		for idx, attribute := range result.Attributes {
+			switch attribute.Kind {
+			case responsev1.InspectPoliciesResponse_Attribute_KIND_PRINCIPAL_ATTRIBUTE:
+				attributes[idx] = fmt.Sprintf("%s.attr.%s", conditions.CELPrincipalAbbrev, attribute.Name)
+			case responsev1.InspectPoliciesResponse_Attribute_KIND_RESOURCE_ATTRIBUTE:
+				attributes[idx] = fmt.Sprintf("%s.attr.%s", conditions.CELResourceAbbrev, attribute.Name)
+			default:
+				attributes[idx] = attribute.Name
+			}
+		}
+
 		variables := make([]string, len(result.Variables))
 		for idx, variable := range result.Variables {
 			variables[idx] = variable.Name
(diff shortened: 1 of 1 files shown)


charithe (commented):
This works for policies with just a few variables, attributes etc. However, that won't be the case always. Even with my terminal at full screen, there are places where it's kind of breaking down already. That means it will be pretty unreadable for people running tmux or other tiling terminal multiplexers.

image


A good example of CLI tools dealing with this kind of data display are the database clients like `psql`. If the data doesn't fit on the viewport, they switch from column-oriented display to row-oriented display automatically. For now, I don't think we have to be that clever. Let's just always display the output of `cerbosctl inspect policies` as rows.

```
POLICY ID : resource_policies/policy_01.yaml
ACTIONS   : approve,assign,cancel,deprecated,resend,view:public
ATTRIBUTES: P.attr.email,P.attr.is_engineer,P.attr.org
VARIABLES : a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p,q,r,s,t,u,v,w,x,y,z
-------------------
POLICY ID : resource_policies/policy_02.yaml
ACTIONS   : approve
ATTRIBUTES: P.attr.email
VARIABLES : a
-------------------
```

cmd/cerbosctl/inspect/internal/print.go
```diff
@@ -86,23 +86,43 @@ func printJSON(w io.Writer, results []*responsev1.InspectPoliciesResponse_Result
...
+				"%s\n%s\n%s\n%s",
+				fmt.Sprintf("POLICY ID : %s", result.PolicyId),
+				fmt.Sprintf("ACTIONS   : %s", strings.Join(result.Actions, separator)),
+				fmt.Sprintf("ATTRIBUTES: %s", strings.Join(attributes, separator)),
+				fmt.Sprintf("VARIABLES : %s", strings.Join(variables, separator)),
+			)
+		}
 
-	tw.Render()
+		fmt.Fprintf(w, "%s\n-------------------\n", row)
```
charithe, line 126:
```suggestion
        const rowSeparator = strings.Repeat("-", 80)
		fmt.Fprintf(w, "%s\n%s\n", row, rowSeparator)
```

cmd/cerbosctl/inspect/internal/print.go
```diff
@@ -86,23 +86,43 @@ func printJSON(w io.Writer, results []*responsev1.InspectPoliciesResponse_Result
...
+				strings.Join(attributes, separator),
+				strings.Join(variables, separator),
+			)
+		} else {
+			row = fmt.Sprintf(
+				"%s\n%s\n%s\n%s",
+				fmt.Sprintf("POLICY ID : %s", result.PolicyId),
+				fmt.Sprintf("ACTIONS   : %s", strings.Join(result.Actions, separator)),
+				fmt.Sprintf("ATTRIBUTES: %s", strings.Join(attributes, separator)),
+				fmt.Sprintf("VARIABLES : %s", strings.Join(variables, separator)),
```
charithe, line 122:
```suggestion
				"POLICY ID : " + result.PolicyId,
				"ACTIONS   : " + strings.Join(result.Actions, separator),
				"ATTRIBUTES: " + strings.Join(attributes, separator),
				"VARIABLES : " + strings.Join(variables, separator),
```

charithe: Some policies don't have attributes, variables or actions. It'd look nicer if we don't print headers for values that don't exist.

cmd/cerbosctl/inspect/internal/print.go
```diff
@@ -86,23 +87,46 @@ func printJSON(w io.Writer, results []*responsev1.InspectPoliciesResponse_Result
...
-			strings.Join(result.Actions, separator),
-			strings.Join(variables, separator),
-		})
+		if noHeaders {
+			fmt.Fprintf(w,
+				"%s\n%s\n%s\n%s\n%s\n",
+				result.PolicyId,
+				strings.Join(result.Actions, separator),
+				strings.Join(attributes, separator),
+				strings.Join(variables, separator),
```
charithe, line 115: We should handle the empty actions, attributes and variables the same way as below. Otherwise, there are lots of extra blank lines in the output. 🙂

charithe (approved):

PR #2308, opened 2024-09-05
chore(tooling): Fix confdocs not being able to parse examples consisting of an array

### LINKED ISSUE #1638: Confdocs can't parse array examples correctly
Given this configuration: 

```go
Compression []string `yaml:"compression" conf:",example=['zstd', 'snappy', 'none']"`
```

The parser stops at the first comma and produces an incomplete example.

```asciidoc
compression: ['zstd' # Compression sets the compression algorithm to use in order of priority. Valid values are "none", "gzip", "snappy","lz4", "zstd". Default is ["snappy", "none"].
```

oguzhand95:
Parsing logic from `structtag` doesn't have a way to configure whether we want to ignore a `,` character inside of square brackets while splitting the tag value such as;
```
Compression []string `conf:",example=['zstd', 'snappy', 'none']"`
```
This makes us unable to parse array examples such as: `['zstd', 'snappy', 'none']`

One possible solution to this was copying the implementation (or maybe we should fork it?) from `structtag` and modifying it to use [go-andiamo/splitter](github.com/go-andiamo/splitter), which I did for this PR.

fixes #1638

### DIFF
--- a/hack/tools/confdocs/confdocs.go
+++ b/hack/tools/confdocs/confdocs.go
@@ -20,16 +20,16 @@ import (
 	"path/filepath"
 	"regexp"
 	"sort"
-	"strconv"
 	"strings"
 	"text/template"
 
-	"github.com/go-andiamo/splitter"
 	"github.com/mattn/go-isatty"
 	"go.elastic.co/ecszap"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 	"golang.org/x/tools/go/packages"
+
+	"github.com/cerbos/cerbos/hack/tools/confdocs/structtag"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


internal/audit/kafka/conf.go
```diff
@@ -55,7 +55,7 @@ type Conf struct {
 	// Brokers list to seed the Kafka client.
 	Brokers []string `yaml:"brokers" conf:"required,example=['localhost:9092']"`
 	// Compression sets the compression algorithm to use in order of priority. Valid values are "none", "gzip", "snappy","lz4", "zstd". Default is ["snappy", "none"].
-	Compression []string `yaml:"compression" conf:",example=['snappy']"`
+	Compression []string `yaml:"compression" conf:",example=['zstd', 'snappy', 'none']"`
```
charithe, line 58: If the default is `['snappy', 'none']`, let's set the example value to that as well.
```suggestion
	Compression []string `yaml:"compression" conf:",example=['snappy', 'none']"`
```

hack/tools/confdocs/confdocs.go
```diff
@@ -542,3 +543,122 @@ func doInitLogging(level string) {
 
 	logger = l.Sugar()
 }
+
+var (
```
charithe, line 547: Let's put the code extracted from `structtag` into its own file and include the copyright notice from https://github.com/fatih/structtag/blob/master/LICENSE.

charithe (approved):

PR #2325, opened 2024-09-26
enhancement: Introduce policy scope FALL_THROUGH_ON_ALLOW strategy

Sambigeara:
This PR introduces a policy-agnostic attribute `ScopePermissions` with two options:

- `SCOPE_PERMISSIONS_OVERRIDE_PARENT`: this is pre-existing behaviour for resource and principal policies. If the scoped policy results in allow or explicit deny, it returns that; otherwise (if the scoped policy results in no match) it falls through to the parent.
- `SCOPE_PERMISSIONS_REQUIRE_PARENTAL_CONSENT_FOR_ALLOWS`: this is pre-existing behaviour for newly introduced role policies. If the scoped policy results in no match or explicit deny, it returns deny; otherwise (if the scoped policy results in allow) it falls through to the parent.

Any policy type is now free to implement either strategy. If unimplemented, it will default to existing behaviour.


internal/engine/evaluator.go
```diff
@@ -662,14 +681,18 @@ func (er *PolicyEvalResult) unresolvedActions() []string {
...
+func (er *PolicyEvalResult) setEffect(action string, scopeFallThrough policyv1.ScopeFallThrough, effect EffectInfo) {
 	delete(er.toResolve, action)
 
 	if effect.Effect == effectv1.Effect_EFFECT_DENY {
 		er.Effects[action] = effect
 		return
 	}
 
+	if effect.Effect == effectv1.Effect_EFFECT_ALLOW && scopeFallThrough == policyv1.ScopeFallThrough_SCOPE_FALL_THROUGH_ON_ALLOW {
+		effect.Effect = effectv1.Effect_EFFECT_NO_MATCH
```
charithe, line 693: Wouldn't this ultimately lead to a DENY if none of the parents have a rule matching the action?

Sambigeara: Failed to write [this comment](https://github.com/cerbos/cerbos-private/issues/1244#issuecomment-2377302415) quick enough.

There's a whole subset of cases not covered at the moment which I'm tackling (sorry for premature draft PR)

Sambigeara: Just returning to this comment. Although I've since refactored this section out, wouldn't what you described have been the desired behaviour?

FALL_THROUGH_ON_ALLOW requires an explicit ALLOW from the ancestor policies in order to pass 🤔

Sambigeara: Actually, not in the case of the resource and principal policies in lieu of the removal of the key from `toResolve`, I think 🤔

Sambigeara:
Two things still open for discussion: [naming](https://github.com/cerbos/cerbos-private/issues/1244#issuecomment-2377227959), and [role policy evaluation](https://github.com/cerbos/cerbos-private/issues/1244#issuecomment-2379353927).

Sambigeara (commented):
Two things still open for discussion: [naming](https://github.com/cerbos/cerbos-private/issues/1244#issuecomment-2377227959), and [role policy evaluation](https://github.com/cerbos/cerbos-private/issues/1244#issuecomment-2379353927).

Sambigeara:
OK the more I work on the documentation changes, the more I realise that it's a project in itself. I want to get this PR out in front of reviewing eyes so I can make any necessary adjustments sooner rather than later, and will follow up with the docs changes afterwards.

internal/engine/engine.go
```diff
@@ -620,6 +620,23 @@ func (engine *Engine) getRolePolicySets(ctx context.Context, scope string, roles
...
+		for _, r := range sets {
+			switch r.GetRolePolicy().ScopePermissions { //nolint:exhaustive
+			case policyv1.ScopePermissions_SCOPE_PERMISSIONS_REQUIRE_PARENTAL_CONSENT_FOR_ALLOWS:
+				requireParentalConsent++
+			case policyv1.ScopePermissions_SCOPE_PERMISSIONS_OVERRIDE_PARENT:
+				overrideParent++
+			}
+
+			if requireParentalConsent > 0 && overrideParent > 0 {
+				err = errors.New("invalid scope permissions: role policies cannot combine different scope permissions within the same scope")
```
charithe, line 635: Couldn't we actually do this at compile time? If we don't allow role policies to have different permission scopes, does it make sense to allow them to exist in the repo, hoping that they don't ever get used at the same time?

internal/engine/evaluator.go
```diff
@@ -344,6 +365,14 @@ func (rpe *resourcePolicyEvaluator) Evaluate(ctx context.Context, tctx tracer.Co
 			if err != nil {
 				return nil, err
 			}
+
+			if p.ScopePermissions == policyv1.ScopePermissions_SCOPE_PERMISSIONS_REQUIRE_PARENTAL_CONSENT_FOR_ALLOWS {
+				for _, a := range result.unresolvedActions() {
+					if _, ok := requireParentalConsent[a]; !ok {
+						result.setEffect(a, EffectInfo{Effect: effectv1.Effect_EFFECT_DENY, Policy: noMatchScopePermissions, Scope: input.Resource.Scope})
```
charithe, line 372: This is tricky 🤔 
If we want to defer to the parents for the final decision, should we deny the unresolved actions here? Shouldn't all those actions that didn't get an explicit DENY be passed on up the chain? One of the parent policies could have a rule that matches the unresolved action.

internal/engine/engine.go
charithe: I just realised that we do need to keep this check here for dynamic stores like databases where policies could be added at any time. However, we should still do the same check at index build time as well so that `cerbos compile` command can warn users when they have an invalid repo.

Sambigeara: I actually tried a couple of things here:

Compile time didn't seem to work because `GetAll` is (understandably) a policy-agnostic API. It returns a list of individual policy sets, rather than a single set with all role policies. The most immediate logical boundary with access to the exclusive, related set of role policies was here. It's also interface implementation agnostic.

I also considered implementing at index build time, but I was nervous to implement there because I'm not sure we can guarantee that that code path might always be used?

internal/engine/evaluator.go
Sambigeara: Hmm. My interpretation of the desired behaviour was that this `scopePermission` mode requires an explicit match/`ALLOW` in order to pass up for further evaluation 🤔 

I think this differs from the weird role policy case because resource policies are mapped 1:1 with the request entities (unlike roles, which can return more than one role policy and should be evaluated independently).

Is there a particular use case that I'm missing here, or would this come down to a "we think it should behave like A it B"?

internal/engine/engine.go
charithe: If we have a disk-based store that contains role policies, I think we should check that they all share the same `scopePermissions` value. Otherwise, it's a bit of a gotcha because a chance combination of roles can result in an error at runtime.

That could be done in a different PR though. Doesn't have to be in this one.

Sambigeara: Oh, I hadn't refreshed so hadn't seen your follow up.

I can add back in the index build time check though (stupidly _just_ deleted the git stash I had with that change 🤠. Need a longer TTL)

internal/engine/evaluator.go
charithe: I think that one of the common use cases for `REQUIRE_PARENTAL_CONSENT` mode would be to define all your access rules at the top scope and then using the child scopes to explicitly deny certain actions. In that case, the child scopes won't have rules defined for most of the available actions and they would fall into the `unresolved` group here. Otherwise, the users would have to copy all the rules from the parent down to every child -- which kind of defeats the purpose.

You could argue that the above use case should be done with `OVERRIDE_PARENT`. The issue is that then it would allow a child to (intentionally or unintentionally) allow something prohibited by the parent as well.

Sambigeara: > I think that one of the common use cases for `REQUIRE_PARENTAL_CONSENT` mode would be to define all your access rules at the top scope and then using the child scopes to explicitly deny certain actions.

I feel like this opens up the opportunity to write riskier permissions as it would require a policy author to remember to add every explicit `DENY`. The `DENY_ALL` with explicit `ALLOW`s approach, although potentially more laborious, removes that risk. I also like that it would be consistent with role policy behaviour.

That said, like with your example above, perhaps our users would prefer less handholding and a more efficient authoring flow. I'm very much torn.

charithe: I see your point about being consistent with role policies. But, at the same time, resource policies are more complicated because the rules can have conditions and outputs and so on. Requiring users to duplicate those rules at every scope level is not great DX. It's quite laborious and leaves room for errors and mistakes. Therefore, I think that we shouldn't expect the child resource policies to cover all possible actions. 

In terms of implementation, I think the tricky part is keeping track of the approval chain. Perhaps we could change `unresolved` to `map[string]int`? 

For each policy in policy set:
- Go through every action in `unresolved` and set its counter to 0
- If there's a DENY rule for the action, it's denied. Full stop.
- If there's an ALLOW rule for the action, put it back into the `unresolved` pile with a counter value of 1.
- If there's no rule matching the action, put it back into the `unresolved` pile with a counter value of 0. 

At the end, go through `unresolved`:
- Any action with a counter of 1 is ALLOWED
- Everything else is DENIED

Sambigeara: So to clarify, the behaviour you're proposing is for resource policies with `REQUIRE_PARENTAL_CONSENT` to do the following:

- Explicit `EFFECT_DENY` -> `DENY` immediately
- Explicit `EFFECT_ALLOW` -> pass up to parent scopes for further evaluation
- No match -> pass up to parent scopes for further evaluation

So for resource policies, the only behavioural difference between `OVERRIDE_PARENT` and `REQUIRE_PARENTAL_CONSENT` is for matching rules with `EFFECT_ALLOW`:

- `OVERRIDE_PARENT`: immediate `EFFECT_ALLOW`
- `REQUIRE_PARENTAL_CONSENT`: acts as though it's an `EFFECT_NO_MATCH`

Does this sound right?

charithe: Yes, that sounds right.

charithe (approved):

PR #2333, opened 2024-10-11
enhancement: Make scopePermissions a required field on role policies

Sambigeara:


Sambigeara (commented):
Change `permissibleActions` to `allowActions`?

api/public/cerbos/policy/v1/policy.proto
```diff
@@ -130,7 +130,7 @@ message RolePolicy {
   }
   string scope = 2 [(buf.validate.field).string.pattern = "^([[:alnum:]][[:word:]\\-]*(\\.[[:word:]\\-]*)*)*$"];
   repeated RoleRule rules = 3;
-  ScopePermissions scope_permissions = 4;
+  ScopePermissions scope_permissions = 4 [(buf.validate.field).required = true];
```
charithe, line 133: I think this needs an `enum in` constraint as well. Otherwise, `SCOPE_PERMISSIONS_UNSPECIFIED` would be accepted as a valid value as well.

charithe (approved):

charithe (approved):

PR #2340, opened 2024-10-15
feat: Use groups of principals and/or resources in test cases

haines:
This PR introduces `principalGroups` and `resourceGroups`, which may be defined in test fixture files or in test suites and then referenced in the inputs and expectations of test cases.

I've relaxed the rule about specifying multiple expectations for the same principal+resource pair because in the case of groups with overlapping members you might want to specify expectations for these separately. Now, we merge the expectations and only throw an error if they are inconsistent.

A group defined in a test suite overrides one with the same key in the fixtures file, the same as fixtures themselves. A group in a fixtures file can only refer to fixtures defined within that file. A group defined in a test suite may refer to shared fixtures and/or fixtures defined in the suite.


charithe (dismissed):

internal/verify/test_matrix.go
```diff
@@ -95,31 +105,128 @@ func buildExpectationLookup(table *policyv1.TestTable) (map[testMatrixKey]testMa
...
+		for _, entry := range oe.Expected {
+			entries[entry.Src] = entry.Val
+		}
+		outputs[oe.Action] = &policyv1.Test_OutputEntries{Entries: entries}
+	}
+	return outputs
+}
+
+func mergeExpectations(key testMatrixKey, target *testMatrixExpectations, actions map[string]effectv1.Effect, outputs map[string]*policyv1.Test_OutputEntries) (*testMatrixExpectations, error) {
+	if target == nil {
```
haines, line 132: @charithe I realised there was a bug in the first iteration of this, where I had

```go
if a == nil {
	return b, nil
}
```

It seems like a nice optimization to avoid allocations but in doing so we end up sharing the same `actions` map between multiple principals; if we have expectations for groups X and Y, where there is some degree of overlap, then the expectations end up being the same for all members of both groups, not just the ones who are actually in both groups.

I've fixed that by making sure we copy the actions into a new map. The same goes for outputs (previously we were copying the `[]*TestTable_OutputExpectations` into a `map[string]*policyv1.Test_OutputEntries` inside the inner loop in `buildExpectationLookup`, but I've moved it to the outer loop now which means we have to avoid the same issue).

PR #2341, opened 2024-10-16
feat(plan): Add support for role policies

dbuduev:


internal/engine/engine.go
```diff
@@ -262,24 +262,50 @@ func (engine *Engine) doPlanResources(ctx context.Context, input *enginev1.PlanR
...
 		if err != nil {
 			return nil, nil, err
 		}
+		effInfo, ok := evalResult.Effects[input.Action]
+		if !ok {
+			return nil, nil, errors.New("role policy evaluator unexpected result")
+		}
+		effect := effInfo.Effect
+		if effect != effectv1.Effect_EFFECT_ALLOW {
+			skipResourcePolicies = true
```
Sambigeara, line 282: The role policy evaluation logic is rather confusing and [continuing to evolve](https://github.com/cerbos/cerbos-private/issues/1380) (so I'm struggling to reason through it myself), but I don't think this is quite right.

In normal `Check` calls, even `Effect_EFFECT_DENY` effects are evaluated in resource policies now, and can be overridden in the case where a resource policy rule issues an allow for an action _if there is another, independent role for that rule_.

[Haines' comment](https://github.com/cerbos/cerbos-private/issues/1334#issuecomment-2406975989) and the linked issues therein do a good job of providing the context as to why.

internal/engine/engine.go
```diff
@@ -277,23 +278,49 @@ func (engine *Engine) doPlanResources(ctx context.Context, input *enginev1.PlanR
...
+		if !ok {
+			return nil, nil, errors.New("role policy evaluator unexpected result")
+		}
+		effect := effInfo.Effect
+		if effect != effectv1.Effect_EFFECT_ALLOW {
+			engineinternal.SubstractSets(effectiveRoles, roles)
+		}
+		if result.Empty() {
+			result = mkUnconditionalPolicyPlanResult(effInfo.Scope, effect)
+		}
```
Sambigeara, line 302: I think this needs to be:

```suggestion
		// `effect` can only be ALLOW if the policy has the resource:action pair and has scope permissions set to "override parent".
		// It can only be DENY if the resource:action pair does not exist.
		// In both these cases, the evaluation (for the given role) can terminate within the role policies.
		if effect != effectv1.Effect_EFFECT_NO_MATCH {
			engineinternal.SubstractSets(unresolvedRoles, roles)
			if result.Empty() {
				result = mkUnconditionalPolicyPlanResult(effInfo.Scope, effect)
			}
		}
```

internal/engine/engine.go
```diff
@@ -277,23 +278,49 @@ func (engine *Engine) doPlanResources(ctx context.Context, input *enginev1.PlanR
...
+	rpEvaluator, err := engine.getRolePolicyEvaluator(ctx, opts.evalParams, ppScope, input.Principal.Roles)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to get check for [%s.%s]: %w", rpName, rpVersion, err)
+		return nil, nil, fmt.Errorf("failed to get role policy evaluator: %w", err)
 	}
-
-	if policy := policySet.GetResourcePolicy(); policy != nil {
-		policyEvaluator := planner.ResourcePolicyEvaluator{Policy: policy, Globals: opts.Globals(), SchemaMgr: engine.schemaMgr, NowFn: opts.NowFunc()}
-		plan, err := policyEvaluator.EvaluateResourcesQueryPlan(ctx, input)
+	effectiveRoles := engineinternal.ToSet(input.Principal.Roles)
```
Sambigeara, line 285: Perhaps this could be named `unresolvedRoles`, as we're determining whether or not they need to be evaluated in the resource policies?

Sambigeara (approved):
Nice 👍 , thanks for fighting through this complexity!

I think we can implement parent roles separately. If this is merged imminently, I can probably do it in my current, open PR.

PR #2349, opened 2024-10-17
enhancement: Role policies parent roles field

Sambigeara:
Introduce a set of parent roles which a role policy assumes.

* Parent roles can originate from role policies defined from the policy set. Therefore, at role policy set retrieval time, we recursively retrieve and aggregate all possible role policies.
* Both `scopePermissions` settings are honoured. "Override parent" inherits the permissions of parent roles and has the authority to return `ALLOW` early if a `resource:action` pair exists in any of the policies. "Required parental consent for allows" narrows the scope for the full set of (recursive) parent roles.
* Role policies with "override parent" now return `NO_MATCH` (rather than `DENY`) for `resource:action` pairs that don't exist. This covers the case where we need to inherit a base (non role policy) role, and evaluate it in the resource policies in the case where a `resource:action` pair isn't defined in the role policy. It also brings the behaviour in line with resource policy types.
* This change also introduces parent role support into the query planner.

There is scope for some caching improvements to the recursive lookup, but I don't think this is super straight forward as cache invalidation requires context from a separate area of concern within the PDP (we'd need to be aware of index updates from within the engine). This is doable, but we don't know if it's necessary yet, and we're going to revamp the index at some point anyway so that might come first.


Sambigeara (commented):
There's _one_ thing I'm not yet 100% confident is correct.

At evaluation time, I'm union'ing all parent roles (and named roles) from all matched role policies and adding them to the `AssumedRoles` slice.

Then I'm using that single slice to check against resource policy evaluations by effectively asking:

> Were all roles defined in the resource policy rule explicitly referenced in the role policy evaluation that led to the implicit deny?

Am I able to use the single, union'ed set, or should I instead evaluate each role policy set against the resources policies individually? 🤔 

**_EDIT:_** I think this works because it only applies to implicit denies (e.g. no role policy matches the `resource:action` pair). There are two cases to cover:
- resource policy rule IdP roles are a subset of role policy activated roles: in this case, we'd expect the implicit deny to be honoured, which it will, because NONE of the role policies had the `resource:action` pair.
- resource policy rule IdP roles are a superset of role policy activated roles: in this case, we'd want the non role policy activated roles to be evaluated and honoured, which they will again because the principal had a role which wasn't referenced by ANY of the role policies...

🤔 🤔

internal/engine/planner.go
```diff
@@ -30,18 +29,14 @@ func PlannerEvaluateRolePolicy(ctx context.Context, tctx tracer.Context, evaluat
...
-		return nil, nil, errors.New("role policies produced validation errors") // this shouldn't happen as role policies doesn't evaluate result.
+		return nil, errors.New("role policies produced validation errors") // this shouldn't happen as role policies doesn't evaluate result.
 	}
-	if rpe, ok := evaluator.(*rolePolicyEvaluator); ok {
-		roles := make([]string, 0, len(rpe.policies))
-		for _, policy := range rpe.policies {
-			roles = append(roles, policy.Role)
-		}
-		return result, internal.ToSet(roles), nil
+	if _, ok := evaluator.(*rolePolicyEvaluator); ok {
```
charithe, line 37: Can this be anything else? If so, shouldn't we do this check at the beginning of the function? 

Why not just change the `evaluator` type to `*rolePolicyEvaluator`?

Issue #2158, opened 2024-05-17
Add `defaultPolicyVersion` to test options


haines:
Currently it is possible to set ⅔ of the [engine configuration settings](https://docs.cerbos.dev/cerbos/latest/configuration/engine) (`globals` and `lenientScopeSearch`) in test options.

It should be possible to set `defaultPolicyVersion`, as well.

PR #2352, opened 2024-10-18
feat: Allow default policy version to be configured in tests

haines:
Closes https://github.com/cerbos/cerbos/issues/2158


docs/modules/policies/pages/compile.adoc
```diff
@@ -127,52 +129,55 @@ tests: <11>
...
+<5> Optionally set xref:configuration:engine.adoc#lenient_scopes[lenient scope search] for this test suite
+<6> Optionally set xref:configuration:engine.adoc#globals[globals] for this test suite
+<7> Map of principal fixtures. The key is a string that can be used to refer to the associated principal.
+<8> Map of principal groups. The key is a string that can be used to refer to the associated group of principal fixtures.
+<9> Map of resource fixtures. The key is a string that can be used to refer to the associated resource.
+<10> Map of resource groups. The key is a string that can be used to refer to the associated group of resource fixtures.
+<11> Map of (optional) auxiliary data fixtures required to evaluate some requests. The key is a string that can be used to refer to the associated auxData.
+<12> List of tests in this suite
+<13> Name of the test
+<14> Optionally set options that apply to just this test. Test-specific options are not merged with suite-wide options, so any unspecified values revert to the default.
```
haines, line 160: I found [this behaviour](https://github.com/cerbos/cerbos/blob/018340b74e9d00eda655386c1f5ae54c205b265d/internal/verify/run_test_suite.go#L166-L169) slightly surprising, but given we can't distinguish between unset and zero values, I think we have to live with it - but it's worth documenting.

charithe (approved):

Issue #1486, opened 2023-03-08
Error test suites that exercise time-based conditions without defined time


haines:
`options.now` is not required in test suites, even if the tests exercise time-based conditions.

Perhaps we should be a bit stricter and error out if a test executes a time-based function and `options.now` is not provided?

Alternatively we could make this a warning, but I think an error is preferable to make flaky tests impossible.

PR #2354, opened 2024-10-18
enhancement!: Report an error if a test exercised time-based policy conditions without specifying `now`

haines:
Closes https://github.com/cerbos/cerbos/issues/1486

### DIFF
--- a/internal/verify/run_test_suite.go
+++ b/internal/verify/run_test_suite.go
@@ -5,7 +5,9 @@ package verify
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"time"
 
 	effectv1 "github.com/cerbos/cerbos/api/genpb/cerbos/effect/v1"
 	enginev1 "github.com/cerbos/cerbos/api/genpb/cerbos/engine/v1"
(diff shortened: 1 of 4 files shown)


internal/verify/run_test_suite.go
```diff
@@ -347,11 +351,17 @@ func runTest(ctx context.Context, eng Checker, test *policyv1.Test, action strin
...
+func performCheck(ctx context.Context, eng Checker, inputs []*enginev1.CheckInput, options *policyv1.TestOptions, trace bool) (_ []*enginev1.CheckOutput, traces []*enginev1.Trace, _ error) {
 	var checkOpts []engine.CheckOpt
 
+	usedDefaultNow := false
 	if now := options.GetNow(); now != nil {
 		checkOpts = append(checkOpts, engine.WithNowFunc(now.AsTime))
+	} else {
+		checkOpts = append(checkOpts, engine.WithNowFunc(func() time.Time {
+			usedDefaultNow = true
+			return time.Time{}
```
dbuduev, line 363: AFAIU, this code snippet helps to fail the misconfigured test, but the test must first be error-free. Can the call to `Check` result in an error because time isn't deterministic? I think it can.

haines: I'm not sure how the return value of `now()` being the zero time could cause a `Check` call to error out. Even if it caused a CEL expression to error, at worst that error would cause a rule not to be activated, wouldn't it? It wouldn't be propagated back to the caller.

Is there a scenario you're thinking of where this wouldn't need the case?

dbuduev: No, there shouldn't be a problem with zero value. My bad.

dbuduev (approved):
LGTM

PR #2361, opened 2024-10-29
enhancement(plan): Query plan scope permissions support for resource and principal policies

dbuduev:


charithe (commented):
LGTM. Would just like to double check with @Sambigeara in case I'd forgotten something about the original implementation.

Sambigeara:
Apologies, I'll take a look at this properly tomorrow morning when my brain's less fried!

internal/engine/planner/planner.go
```diff
@@ -79,6 +81,43 @@ func CombinePlans(principalPolicyPlan, resourcePolicyPlan *PolicyPlanResult) *Po
...
+	}
+	return &PolicyPlanResult{
+		Scope:            current.Scope,
+		ScopePermissions: scopePermissions,
+		AllowFilter:      allowFilter,
+		DenyFilter:       append(acc.DenyFilter, current.DenyFilter...),
+	}
+}
+
+func NewPolicyPlanResult(scope string, scopePermissions policyv1.ScopePermissions) *PolicyPlanResult {
```
Sambigeara, line 110: Could we pass the default scope permissions setting as an argument here? I'm a little worried that if this function was ever called from the context of a role policy evaluation, it would be set to the wrong default (the default for role policies is `SCOPE_PERMISSIONS_REQUIRE_PARENTAL_CONSENT_FOR_ALLOWS`).

internal/engine/planner/planner.go
```diff
@@ -101,8 +140,16 @@ func (p *PolicyPlanResult) Add(filter *qpN, effect effectv1.Effect) {
 	}
 }
 
+func (p *PolicyPlanResult) DenyEmpty() bool {
+	return len(p.DenyFilter) == 0
+}
+
+func (p *PolicyPlanResult) AllowEmpty() bool {
```
Sambigeara, line 147: When happening across this function above, my immediate assumption was that it would be allowing something rather than just determining if the allow filters are empty. Perhaps `AllowIsEmpty` or something along those terms would be more clear? Same for `DenyIsEmpty`?

internal/engine/planner/planner.go
```diff
@@ -79,6 +81,43 @@ func CombinePlans(principalPolicyPlan, resourcePolicyPlan *PolicyPlanResult) *Po
...
+		scopePermissions = acc.ScopePermissions
+		allowFilter = acc.AllowFilter
+	} else if !acc.AllowEmpty() {
+		n := len(acc.AllowFilter) * len(current.AllowFilter)
+		allowFilter = make([]*qpN, 0, n)
+		for _, a := range acc.AllowFilter {
+			for _, c := range current.AllowFilter {
+				allowFilter = append(allowFilter, mkNodeFromLO(mkAndLogicalOperation([]*qpN{a, c})))
+			}
+		}
```
Sambigeara, line 100: For my own understanding, am I right in thinking that:

- if the previously evaluated scoped policy had "override parent" scoped permissions
- and therefore reaching this `mergePlans` function again meant that it didn't have a matching rule

That this cartesian product would yield no-op AND filters? E.g. predicates that are guaranteed to return false?

internal/engine/planner/planner.go
dbuduev: Good point. I think we can set the correct default permission in the caller function.

internal/engine/planner/planner.go
dbuduev: No, this is a regular check with the parent policy.
Let's consider a few scenarios. In all of them, rules have the effect `ALLOW`.

Scenario 1: 
A child policy has a rule with the expression: `P.attr.x == 1` and the parent policy, whatever scopePermissions it has, has a rule with the expression: `P.attr.p == 'foo'`. Then, the result is a logical `AND` of these expressions.

Scenario 2:
A resource policy with default scopePermissions has two rules for this action with expressions: `P.attr.x == 1` and `P.attr.y == 3`. A resource with ether x equal one or y equal three will satisfy this policy, so the result is a logical `OR` of these expressions. This is why [here](https://github.com/cerbos/cerbos/blob/2fb58538db3e090847854975f5b90f0502844314/internal/engine/planner/planner.go#L154) we join them with `OR`.

Scenario 3 is a combination of these scenarios:
A child policy requiring parental consent has expressions A and B.
A parent policy with any scopePermission has expressions C and D. 
The result is:  `A && C || A && D || B && C || B && D`. Hence the cartesian product in `mergePlan`.

Whether the scopePermission is `OVERRIDE_PARENT` matters when I call `Complete` to check that we can exit the loop in the :
```go
	if !p.Empty() && p.ScopePermissions == policyv1.ScopePermissions_SCOPE_PERMISSIONS_OVERRIDE_PARENT {
		return true
	}
```
If the policy has a predicate and the `scopePermissions` equals to `OVERRIDE_PARENT`, then we can exit the loop.

I realised there's a bug. I call `acc.Complete()`, but I should call `currentResult.Complete()` instead. The `acc` won't be empty as it takes predicated from the child scopes.

Sambigeara: If the child policy had "override parent" and no matches, with expressions A and B, and
A parent policy has expressions C and D

Can't we say that A and B are guaranteed to be `false`, so all cartesian products `A && C || A && D || B && C || B && D` would always be `false` too?

dbuduev: Sorry for the confusion. I meant that all expressions belong to the matched rules. Also, the expressions were intended to refer to unknown resource attributes, so I updated my comment to replace all `P.attr` occurrences with `R.attr`. 

> Can't we say that A and B are guaranteed to be false, so the cartesian product A && C || A && D || B && C || B && D would always be false too?

Taking my corrections into account, the answer is no. These expressions must appear in the Plan API response.

Sambigeara (dismissed):

dbuduev:
@Sambigeara, please give it another look, as your review was removed after I updated a test case.

Sambigeara (approved):

PR #2393, opened 2024-11-26
chore: Remove SQL Server driver

### LINKED ISSUE #2389: Drop support for SQL Server
Support for using SQL Server as a policy store will be dropped in the next release of Cerbos. This is due to several reasons.
- We don't have any developers using Windows that could properly develop and troubleshoot issues with SQL Server. Nor do we have in-house knowledge to help users troubleshoot issues in their environments.
- Testing with the SQL Server containers have been quite flaky and slow.
- Currently it requires us to maintain a different code path that uses a special SQL syntax that's not compatible with other database drivers we support. This makes testing and adding new features quite cumbersome.

charithe:
Fixes #2389


Sambigeara (approved):
Hurrah!

PR #2406, opened 2024-12-06
docs: Update what-is-cerbos.adoc tenant ->tenet

Johnpc123:
tenet : a principle, belief, or doctrine generally held to be true
especially : one held in common by members of an organization, movement, or profession

tenant : one who has the occupation or temporary possession of lands or tenements of another
specifically : one who rents or leases a dwelling (such as a house) from a landlord

#### Description

Fixes #

#### Checklist 

- [ ] The PR title has the correct prefix 
- [ ] PR is linked to the corresponding issue
- [ ] All commits are signed-off (`git commit -s ...`) to provide the [DCO](https://developercertificate.org/)

PR #2409, opened 2024-12-09
chore(docs): Update what-is-cerbos.adoc

oguzhand95:
Quoting @Johnpc123 from cerbos/cerbos#2406:

> tenet : a principle, belief, or doctrine generally held to be true
> especially : one held in common by members of an organization, movement, or profession
>
> tenant : one who has the occupation or temporary possession of lands or tenements of another
> specifically : one who rents or leases a dwelling (such as a house) from a landlord

Closes #2406


charithe:
Let's fix the title and merge the other PR.

PR #2410, opened 2024-12-09
enhancement: Rule table engine

Sambigeara:


.justfile
```diff
@@ -270,3 +270,9 @@ prometheus:
         bitnami/prometheus:latest \
         --config.file=/opt/bitnami/prometheus/conf/prometheus.yml \
         --enable-feature=otlp-write-receiver
+
+# TODO(saml) remove
```
Sambigeara, line 274: This'll go

Dockerfile.newengine
```diff
@@ -0,0 +1,20 @@
+FROM golang:1.23-alpine AS builder
```
Sambigeara, line 1: This'll also go

charithe:
Nice work 🙌🏽 There's lots to digest here and I'll try to find a distraction-free time slot to go through it all and add more useful comments.

Boring PR admin stuff: this should probably go into a feature branch for a short while until we iron everything out. The benchmarks are not particularly happy at the moment but I don't want to prematurely optimise things and make it unreadable as well.

internal/engine/rule_table.go
```diff
@@ -0,0 +1,557 @@
...
+func (rt *RuleTable) addPolicy(rps *runtimev1.RunnablePolicySet) {
+	switch rps.PolicySet.(type) {
+	case *runtimev1.RunnablePolicySet_ResourcePolicy:
+		rt.addResourcePolicy(rps.GetResourcePolicy())
+	case *runtimev1.RunnablePolicySet_RolePolicy:
+		rt.addRolePolicy(rps.GetRolePolicy())
+	}
+}
+
+func (rt *RuleTable) addResourcePolicy(rrps *runtimev1.RunnableResourcePolicySet) {
```
Sambigeara, line 93: There's a discrepancy between full and partial rule tables. Either this function needs to act on individual `RunnableResourcePolicySet_Policy` instances, or it needs to handle duplicate rows that occur when `GetAll` returns the same policy higher up the scope chain.

internal/engine/engine_test.go
```diff
@@ -43,39 +44,51 @@ func TestCheck(t *testing.T) {
...
+				if tc.WantError {
+					require.Error(t, err)
+				} else {
+					require.NoError(t, err)
+				}
 
-			for i, have := range haveOutputs {
-				require.Empty(t, cmp.Diff(tc.WantOutputs[i],
-					have,
+				for i, have := range haveOutputs {
```
Sambigeara, line 72: This is still sporadically failing 😭

Sambigeara: By "this" I mean case_21 due to being unable to order outputs by `Src`

internal/engine/evaluator.go
```diff
@@ -114,278 +118,264 @@ func (noopEvaluator) Evaluate(_ context.Context, _ tracer.Context, _ *enginev1.C
...
-		}
+func (rte *ruleTableEvaluator) Evaluate(ctx context.Context, tctx tracer.Context, input *enginev1.CheckInput) (*PolicyEvalResult, error) {
+	version := input.Resource.PolicyVersion
+	if version == "" {
+		version = defaultVersion
 	}
 
-	return &rolePolicyEvaluator{policies: policies}
-}
+	request := checkInputToRequest(input)
```
charithe, line 133: This can be moved to just before line 170 to avoid some allocations if validation fails.

internal/engine/evaluator.go
```diff
@@ -114,278 +118,264 @@ func (noopEvaluator) Evaluate(_ context.Context, _ tracer.Context, _ *enginev1.C
...
 							continue
 						}
+						conditionCache[row.EvaluationKey] = ok
+						satisfiesCondition = ok
+					}
 
-						if !ok {
-							dctx.Skipped(nil, "Condition not satisfied")
+					if satisfiesCondition { //nolint:nestif
+						// If we're just overriding an existing ALLOW, we can skip.
```
charithe, line 296: Shouldn't we still evaluate the outputs for this rule because it was activated?

charithe (commented):
I am only halfway through. The comments below might not make sense because I've missed the big picture 🙈 I'll be back to do the rest of the review soon.

internal/engine/evaluator.go
```diff
@@ -114,278 +118,264 @@ func (noopEvaluator) Evaluate(_ context.Context, _ tracer.Context, _ *enginev1.C
...
-}
+	effectiveDerivedRoles := make(internal.StringSet)
+	// This is a bit of a bolt-on evaluation in order to support the existing requirement of evaluating all derived roles
+	// defined within a resource policy (regardless of if they're used or not). This is primarily for the
+	// runtime.EffectiveDerivedRoles usage
+	includingParentRoles := make(map[string]struct{})
+	for _, r := range rte.GetParentRoles(input.Principal.Roles) {
+		includingParentRoles[r] = struct{}{}
+	}
+	for _, scope := range scopes {
```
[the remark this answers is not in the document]
Sambigeara: This puzzled me as well! I originally didn't have this loop in place because it felt _wrong_ to me, but we have a bunch of expectations in tests that require it to be here 🤔

I would expect `effectiveDerivedRoles` to be those which are defined within a rule that's matched and evaluated to produce an EFFECT. However, that doesn't seem to be existing behaviour--it seems that it includes any derived roles that satisfy the runtime constraints irrespective of whether or not the derived role is present in an effective rule (there's an [initial loop run before evaluation](https://github.com/cerbos/cerbos/blob/main/internal/engine/evaluator.go#L252-L286)).

On further reading, I assumed that this was the case because we require all _potentially_ effective derived roles to be made available for the `runtime.EffectiveDerivedRoles` feature?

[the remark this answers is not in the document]
charithe: There's a subtle distinction between `runtime.EffectiveDerivedRoles` and `meta.EffectiveDerivedRoles`.

In the current engine, each policy in the policy set is evaluated one at a time. It grabs the policy to evaluate, gets all its derived roles, finds out which ones are activated, and sets that list as `runtime.EffectiveDerivedRoles`. That gets reset for the next policy in the set. Whereas, `meta.EffectiveDerivedRoles` is the set of all effective derived roles we've encountered during the evaluation of the entire request. Again, it's not necessarily _every_ derived role from all the policies because evaluation could stop early (e.g. the leaf policy in a scoped set is enough to make all the decisions and we never needed to go up the hierarchy and evaluate the parents). 

If we are really unlucky, Hyrum's law would dictate that someone is relying on this property of `runtime.EffectiveDerivedRoles` resetting each time. But, the chances are minuscule and I think that it's an acceptable breaking change for us to make. However, I do know that some users rely on `meta.EffectiveDerivedRoles` to make some decisions on the client side. So, changing what that means is probably not quite so acceptable.

internal/engine/ruletable/rule_table.go
```diff
@@ -0,0 +1,569 @@
...
+	}
+}
+
+func (rt *RuleTable) WithPolicyLoader(policyLoader policyloader.PolicyLoader) *RuleTable {
+	rt.policyLoader = policyLoader
+	return rt
+}
+
+func (rt *RuleTable) LoadPolicies(rps []*runtimev1.RunnablePolicySet) {
+	rt.mu.Lock()
```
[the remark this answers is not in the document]
Sambigeara: It's [called each time an evaluation is carried out on a mutable store](https://github.com/Sambigeara/cerbos/blob/enhancement/generalise-engine/internal/engine/engine.go#L633), as it builds an ephemeral rule table for the purposes of that sole evaluation (not optimal, but avoids multiple sources of truth and seemed OK to start).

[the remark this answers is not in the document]
Sambigeara: > for each request and that's done by a single goroutine

Sorry I didn't engage with this. Yes, I think this lock is unnecessary

internal/storage/hub/bundle.go
```diff
@@ -214,9 +214,27 @@ func (b *Bundle) GetFirstMatch(_ context.Context, candidates []namer.ModuleID) (
 	return nil, nil
 }
 
-// GetAll attempts to retrieve all policies from the passed modIDs, unlike `GetFirstMatch` which returns the first
+func (b *Bundle) GetAll(ctx context.Context) ([]*runtimev1.RunnablePolicySet, error) {
+	res := []*runtimev1.RunnablePolicySet{}
+	for fqn := range b.manifest.PolicyIndex {
+		modID := namer.GenModuleIDFromFQN(fqn)
+
+		policySet, err := b.GetFirstMatch(ctx, []namer.ModuleID{modID})
```
charithe, line 222: We could extract the loop body of `GetFirstMatch` (minus the `stat`'ing) into its own function and call that from the `GetAll*` methods. (It feels a bit weird to call `GetFirstMatch` from here.  Furthermore, since we are looping through the manifest, the files are guaranteed to exist as well.)

internal/storage/index/builder.go
```diff
@@ -311,9 +379,52 @@ func (idx *indexBuilder) addPolicy(file string, srcCtx parser.SourceCtx, p polic
...
 
+ancestors:
 	for moduleID, fqn := range policy.RequiredAncestors(p.Policy) {
 		ancestorPolicyKey := namer.PolicyKeyFromFQN(fqn)
-		if _, ok := idx.modIDToFile[moduleID]; !ok {
+		_, ok := idx.modIDToFile[moduleID]
+
+		// check to see if matching role policies (with a rule for the given resource) reside in any of the missing scopes
+		if !ok && resourceKind != "" { //nolint:nestif
+			segments := strings.Split(ancestorPolicyKey, "/")
```
charithe, line 389:
```suggestion
			fqn, scope, _ := strings.Cut(ancestorPolicyKey, "/")
```

internal/storage/index/builder.go
```diff
@@ -311,9 +379,52 @@ func (idx *indexBuilder) addPolicy(file string, srcCtx parser.SourceCtx, p polic
...
+			baseFqn := segments[0]
+
+			var scope, version string
+			if len(segments) == 2 { //nolint:mnd
+				scope = segments[1]
+			}
+
+			subSegments := strings.SplitN(baseFqn, ".", 3) //nolint:mnd
```
charithe, line 397: This can't be guaranteed. Policy names could include `.` characters as well.

internal/engine/ruletable/rule_table.go
```diff
@@ -0,0 +1,569 @@
...
+					evaluationKey := fmt.Sprintf("%s#%s", mergedParameters.Key, ruleFqn)
+					for pr := range rdr.ParentRoles {
+						rt.rules = append(rt.rules, &Row{
+							RuleTable_RuleRow: &runtimev1.RuleTable_RuleRow{
+								OriginFqn:         rrps.Meta.Fqn,
+								Resource:          sanitizedResource,
+								Role:              pr,
+								Action:            a,
+								Condition:         cond,
+								Effect:            rule.Effect,
```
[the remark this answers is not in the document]
Sambigeara: Silly me

internal/engine/engine.go
```diff
@@ -587,22 +587,33 @@ func (engine *Engine) getPartialRuleTable(ctx context.Context, resource, policyV
...
 	addScopedPolicyRules := func(scope string) error {
 		rps, err := engine.getResourcePolicySet(ctx, resource, policyVer, scope, true)
 		if err != nil {
 			return err
 		}
 
 		if rps != nil {
-			toLoad = append(toLoad, rps)
+			// check the first policy scope
+			for _, p := range rps.GetResourcePolicy().GetPolicies() {
```
charithe, line 602: Why not just take the first policy using `rps.GetResourcePolicy().GetPolicies()[0]`?

Does this also mean that it's no longer necessary for us to include all the ancestors of the policy in the policy set? If so, we should think about deprecating that format and making it simpler (in a future release).

internal/engine/ruletable/rule_table.go
```diff
@@ -466,6 +462,17 @@ func (rt *RuleTable) GetSchema(fqn string) *policyv1.Schemas {
 	return nil
 }
 
+func (rt *RuleTable) GetMeta(fqn string) *runtimev1.RuleTableMetadata {
+	rt.mu.RLock()
+	defer rt.mu.RUnlock()
+
+	if s, ok := rt.meta[fqn]; ok {
```
charithe, line 469: It looks like `rt.meta` is not updated in the `deletePolicy` method?

internal/storage/hub/bundle.go
```diff
@@ -250,6 +239,31 @@ func (b *Bundle) GetAllMatching(ctx context.Context, modIDs []namer.ModuleID) ([
 	return res, nil
 }
 
+func (b *Bundle) getMatch(id namer.ModuleID, withFileStat bool) (*runtimev1.RunnablePolicySet, error) {
```
charithe, line 242: Sorry for causing confusion. It's probably not worth having the `withFileStat` boolean parameter here. They always make things tricky to read and I think the `stat` call has negligible overhead in this context because Go reads the zip file manifest into memory anyway.

internal/test/testdata/engine/case_23.yaml
```diff
@@ -51,7 +51,7 @@ wantOutputs:
               "scope": "acme.hr",
             },
         },
-      "effectiveDerivedRoles": ["any_employee", "employee_that_owns_the_record"],
+      "effectiveDerivedRoles": ["employee_that_owns_the_record"],
```
charithe, line 54: Why is `any_employee` not in the list any more? Looks like it should?

internal/engine/engine.go
Sambigeara: For the "just in case" implicit nil checking the `Get*` methods provide, really. But yeah, given the known type I could just do this!

internal/test/testdata/engine/case_23.yaml
charithe: Oh! I think GitHub's showing me the diff against your previous commits in this PR 🫤 
[This file on `main`](https://github.com/cerbos/cerbos/blob/41787ba1b9e47fdabc9ab15a5b03b3de4807b5b2/internal/test/testdata/engine/case_23.yaml#L54) doesn't have `any_employee` either.

Sambigeara: Lol. I too have been fooled because I remember investigating one at length, and assumed it was this one because the file has a diff, but the diff is a comma 🤦‍♂️.

Nice to know it shouldn't be there 😆

charithe (commented):
I haven't fully grokked some parts yet but, since the existing tests pass, that's good enough for me 😄 How are the benchmarks looking at the moment?

Sambigeara:
For the first 23 test cases (which are relevant for regressions), we seem to average about 1.4-1.5 times slower. Outputs as follows:

| Case    | Regression |
|---------|------------|
| case_00 | 1.79       |
| case_01 | 1.77       |
| case_02 | 1.26       |
| case_03 | 1.43       |
| case_04 | 1.33       |
| case_05 | 0.97       |
| case_06 | 0.91       |
| case_07 | 1.00       |
| case_08 | 1.97       |
| case_09 | 2.04       |
| case_10 | 1.54       |
| case_11 | 1.64       |
| case_12 | 0.61       |
| case_13 | 0.98       |
| case_14 | 1.61       |
| case_15 | 1.62       |
| case_16 | 2.97       |
| case_17 | 2.01       |
| case_18 | 1.20       |
| case_19 | 1.02       |
| case_20 | 1.08       |
| case_21 | 1.10       |
| case_22 | 1.36       |
| case_23 | 1.02       |

Issue #2379, opened 2024-11-11
Upgrade to CEL 0.22.0


charithe:
- Some query planner tests are failing with `missing attribute(s)` error for comprehensions
- CEL now supports two variable comprehensions: https://github.com/google/cel-go/pull/1034. Test that they work as expected with policy evaluation and query planning.
- Test new functions (`flatten`, `sort` etc.) and document them: https://github.com/google/cel-go/releases/tag/v0.22.0
- Switch to `ContextEval` to propagate the context.

charithe:
https://github.com/google/cel-go/issues/1079

dbuduev:
Should we consider support for optional types?

charithe:
Yes

charithe:
https://github.com/google/cel-go/issues/1080#issuecomment-2529406579 

This new limitation doesn't seem to affect us.

PR #2412, opened 2024-12-10
chore(deps): Upgrade CEL

dbuduev:
Implements https://github.com/cerbos/cerbos/issues/2379, except optional types, which will be in a separate PR. ePDP changes are coming, too.


internal/engine/planner/lambda.go
```diff
@@ -4,54 +4,156 @@
...
-			obj.lambdaExpr = elements[0]
+		if obj.iterVar2 != "" {
+			obj.operator = TransformList
 		}
+		obj.expr = loopStep.getArg(1).getListElement(0).e()
 	case operators.Conditional:
 		switch e.AccuInit.ExprKind.(type) {
 		case *exprpb.Expr_ListExpr:
-			obj.operator = Filter
+			if e2 := loopStep.getArg(1).getArg(1).getListElement(0).e(); e2.GetCallExpr() != nil {
```
charithe, line 96: Is it guaranteed that `getArg` never returns `nil` in this case? According to the implementation of `getArg`, it _could_  return `nil`...

There are a bunch of places in this file where method calls are chained without nil checks. Can we be absolutely sure that they won't panic at runtime?

internal/engine/planner/lambda.go
```diff
@@ -4,54 +4,156 @@
...
+func (w *wrapper) e() *exprpb.Expr {
+	return (*exprpb.Expr)(w)
+}
+
+func (w *wrapper) getArg(i int) *wrapper {
+	if x := w.e().GetCallExpr(); x != nil && i < len(x.Args) {
+		return (*wrapper)(x.Args[i])
+	}
+	return nil
+}
```
charithe, line 138: One way to handle the nil issues pointed out above would be to do a nil check at the beginning of each method call provided by the `wrapper` object.
```suggestion
func (w *wrapper) getArg(i int) *wrapper {
    if w == nil {
        return nil
    }
	if x := w.e().GetCallExpr(); x != nil && i < len(x.Args) {
		return (*wrapper)(x.Args[i])
	}
	return nil
}
```

docs/modules/policies/pages/conditions.adoc
```diff
@@ -373,17 +377,28 @@ NOTE: The IP address functions are Cerbos-specific extensions to CEL.
...
-| exists   | Check whether at least one element matching the predicate exists | P.attr.teams.exists(t, t.startsWith("comm"))
-| exists_one | Check that only one element matching the predicate exists | P.attr.teams.exists_one(t, t.startsWith("comm")) == false
-| filter   | Filter a list using the predicate | size(P.attr.teams.filter(t, t.matches("^comm"))) == 2
+| all      | Check whether all elements in a list match the predicate. | P.attr.teams.all(t, size(t) > 3) && [1, 2, 3].all(i, j, i < j)
+| distinct | Returns the distinct elements of a list | [1, 2, 2, 3, 3, 3].distinct() == [1, 2, 3]
+| except   | Produces the set difference of two lists | P.attr.teams.except(["design", "engineering"]) == ["communications", "product", "commercial"]
+| exists   | Check whether at least one element matching the predicate exists in a list or map. | P.attr.teams.exists(t, t.startsWith("comm")) && P.attr.limits.exists(k, v, size(k) > 3 && v > 0)
+| exists_one | Check that only one element matching the predicate exists. | P.attr.teams.exists_one(t, t.startsWith("comm")) == false && P.attr.limits.exists_one(k, v, size(k) > 3 && v > 0) == false
```
charithe, line 384: Some of these examples are a bit too complicated because users have to "run" the program inside their head 🙂 They could be simpler and much more obvious.

```suggestion
| exists_one | Check that only one element matching the predicate exists. | P.attr.teams.exists_one(t, t.startsWith("comm")) == false && P.attr.limits.exists_one(k, v, k == "design" && v > 0) == false
```

docs/modules/policies/pages/conditions.adoc
```diff
@@ -373,17 +377,28 @@ NOTE: The IP address functions are Cerbos-specific extensions to CEL.
...
+| filter   | Filter a list using the predicate. | size(P.attr.teams.filter(t, t.matches("^comm"))) == 2
+| flatten   | Flattens a list. If an optional depth is provided, the list is flattened to the specified level | 1,2,[],[],[3,4]].flatten() == [1, 2, 3, 4] && [1,[2,[3,[4]]]].flatten(2) == [1, 2, 3, [4]]
 | hasIntersection| Checks whether the lists have at least one common element | hasIntersection(["design", "engineering"], P.attr.teams)
 | in       | Check whether the given element is contained in the list or map | ("design" in P.attr.teams) && ("acme" in P.attr.clients)
 | intersect| Produces the set intersection of two lists | intersect(["design", "engineering"], P.attr.teams) == ["design"]
-| isSubset| Checks whether the list is a subset of another list | ["design", "engineering"].isSubset(P.attr.teams) == false
+| isSubset | Checks whether the list is a subset of another list | ["design", "engineering"].isSubset(P.attr.teams) == false
 | map      | Transform each element in a list | "DESIGN" in P.attr.teams.map(t, t.upperAscii())
+| range    | Returns a list of integers from 0 to n-1 | lists.range(5) == [0, 1, 2, 3, 4]
```
charithe, line 392: Isn't this a namespaced function that can only be invoked as `lists.range`? I think we should use the full name here.

docs/modules/policies/pages/conditions.adoc
```diff
@@ -373,17 +377,28 @@ NOTE: The IP address functions are Cerbos-specific extensions to CEL.
...
 | size     | Number of elements in a list or map | size(P.attr.teams) == 4 && size(P.attr.clients) == 2
+| slice    | Returns a new sub-list using the indexes provided | [1,2,3,4].slice(1, 3) == [2, 3]
+| sort     | Sorts a list with comparable elements | [3, 2, 1].sort() == [1, 2, 3]
+| sortBy   | Sorts a list by a key value, i.e., the order is determined by the result of an expression applied to each element of the list | [{ "name": "foo", "score": 0 },{ "name": "bar", "score": -10 },{ "name": "baz", "score": 1000 }].sortBy(e, e.score).map(e, e.name) == ["bar", "foo", "baz"]
+| transformList | Converts a map or a list into a list value. The output expression determines the contents of the output list. Elements in the list may optionally be filtered | [1, 2, 3].transformList(i, v, i % 2 == 0, (i * v) + v) == [1, 9] && +
+[1, 2, 3].transformList(i, v, (i * v) + v) == [1, 4, 9]
+| transformMap | Converts a map or a list into a map value. The output expression determines the value of the output map entry; however, the key remains fixed. Elements in the map may optionally be filtered| [1, 2, 3].transformMap(i, v, i % 2 == 0, (i * v) + v) == {0: 1, 1: 9}
```
charithe, line 400:
```suggestion
| transformList | Converts a map or a list into a list value. The output expression determines the contents of the output list. Elements in the list may optionally be filtered | [1, 2, 3].transformList(i, v, i > 0, 2 * v) == [4, 6] && +
[1, 2, 3].transformList(i, v, 2 * v) == [2, 4, 6]
| transformMap | Converts a map or a list into a map value. The key remains unchanged and only the value is changed. | [1, 2, 3].transformMap(i, v, i > 0, 2 * v) == {1: 4, 2: 6}
```

charithe (commented):
I think there are a few more `math` functions available now as well.

internal/engine/planner/lambda.go
dbuduev: No, it will not panic in the presence of a `nil` pointer. I've added the test `TestWrapperDoNotPanic` to demonstrate this.

internal/engine/planner/lambda.go
dbuduev: I had these checks when a wrapper was a struct, but currently, it is a simple type alias of the protobuf `Expr` type, so the `nil` checks aren't needed. The 'wrapper' name may not fit here.
`w.e()` casts `*wrapper` to `*Expr`, then calls protobuf functions, which correctly handle `nil` receivers.

The `TestWrapperDoNotPanic` test, hopefully, proves it.

docs/modules/policies/pages/conditions.adoc
```diff
@@ -380,24 +380,24 @@ NOTE: The IP address functions are Cerbos-specific extensions to CEL.
...
+| flatten   | Flattens a list. If an optional depth is provided, the list is flattened to the specified level | [1,2,[],[],[3,4]].flatten() == [1, 2, 3, 4] && [1,[2,[3,[4]]]].flatten(2) == [1, 2, 3, [4]]
 | hasIntersection| Checks whether the lists have at least one common element | hasIntersection(["design", "engineering"], P.attr.teams)
 | in       | Check whether the given element is contained in the list or map | ("design" in P.attr.teams) && ("acme" in P.attr.clients)
 | intersect| Produces the set intersection of two lists | intersect(["design", "engineering"], P.attr.teams) == ["design"]
 | isSubset | Checks whether the list is a subset of another list | ["design", "engineering"].isSubset(P.attr.teams) == false
 | map      | Transform each element in a list | "DESIGN" in P.attr.teams.map(t, t.upperAscii())
-| range    | Returns a list of integers from 0 to n-1 | lists.range(5) == [0, 1, 2, 3, 4]
+| lists.range    | Returns a list of integers from 0 to n-1 | lists.range(5) == [0, 1, 2, 3, 4]
```
charithe, line 392: This should move one line up before `map` to maintain alphabetical order.

docs/modules/policies/pages/conditions.adoc
```diff
@@ -408,8 +408,23 @@ NOTE: The IP address functions are Cerbos-specific extensions to CEL.
...
+| math.ceil | Compute the ceiling of a double value | math.ceil(1.2) == 2.0 && math.ceil(-1.2) == -1.0
+| math.floor | Compute the floor of a double value | math.floor(1.2) == 1.0 && math.floor(-1.2) == -2.0
+| math.round | Rounds the double value to the nearest whole number with ties rounding away from zero, e.g. 1.5 -> 2.0, -1.5 -> -2.0 | math.round(1.2) == 1.0 && math.round(1.5) == 2.0 && math.round(-1.5) == -2.0
+| math.trunc | Truncates the fractional portion of the double value | math.trunc(1.2) == 1.0 && math.trunc(-1.2) == -1.0
+| math.abs | Returns the absolute value of the numeric type provided as input | math.abs(1.2) == 1.2 && math.abs(-2) == 2
+| math.sign | Returns the sign of the numeric type, either -1, 0, 1 | math.sign(1.2) == 1 && math.sign(-2) == -1 && math.sign(0) == 0
+| math.isInf | Returns true if the input double value is -Inf or +Inf | math.isInf(1.0/0.0) && !math.isInf(1.2)
+| math.isNaN | Returns true if the input double value is NaN, false otherwise | math.isNaN(0.0/0.0) && !math.isNaN(1.2)
+| math.isFinite | Returns true if the value is a finite number | !math.isFinite(0.0/0.0) && math.isFinite(1.2)
```
charithe, line 427: Let's keep this list sorted by function name.

PR #2440, opened 2025-01-09
enhancement: Pre-compile rule table CEL programs

Sambigeara:
Expressions are now checked and compiled to CEL programs at build time rather than at evaluation time, thus reducing hot-path CPU pressure.

This is now more performant than the original, pre ruletable engine (for relevant engine test cases 0-23):

- `sec/op`: -15%
- `allocs/op`: -16%
- `b/ops`: -18%

And the current rule table implementation:

- `sec/op`: -24%
- `allocs/op`: -20%
- `b/ops`: -22%


charithe (approved):
🙌🏽

PR #2441, opened 2025-01-09
fix: Correctly handle GlobMap creation

Sambigeara:
We don't want to match glob patterns when building the rule table index as this leads to non-deterministic outputs and confused developers. Glob matching is only required at eval time.

The feature branch will be good to go after this!

### DIFF
--- a/internal/util/globs.go
+++ b/internal/util/globs.go
@@ -126,7 +126,7 @@ func (gm *GlobMap[T]) Get(k string) (T, bool) {
 	return zero, false
 }
 
-func (gm *GlobMap[T]) GetLiteral(k string) (T, bool) {
+func (gm *GlobMap[T]) GetWithLiteral(k string) (T, bool) {
 	if v, ok := gm.literals[k]; ok {
 		return v, true
 	}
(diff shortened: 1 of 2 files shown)


internal/util/globs.go
```diff
@@ -126,6 +126,19 @@ func (gm *GlobMap[T]) Get(k string) (T, bool) {
...
 }
 
+func (gm *GlobMap[T]) GetLiteral(k string) (T, bool) {
+	if v, ok := gm.literals[k]; ok {
+		return v, true
+	}
+
+	if v, ok := gm.globs[k]; ok {
+		return v, true
+	}
```
charithe, line 136: Should we be looking in globs? It would be a bit confusing because the method name is `GetLiteral`, won't it?

Sambigeara: Ah thanks, I added in the `globs` lookup afterwards and didn't rename.

This method is used during the build step when we want to add to the literally-keyed nested map, rather than at eval time when we're finding any viable glob matches. If we use the `Get` method (with glob matches) we end up matching to any arbitrary nested maps and incorrectly assigning the values.

In short, we need to assign to literal strings (regardless of glob patterns) and query using glob matches. I'll update the name!

PR #2448, opened 2025-01-14
fix: Use correct filterDebug type in e2e query planner test

Sambigeara:
Should solve [this e2e test failure](https://github.com/cerbos/cerbos/actions/runs/12759334544/job/35562917228).

Looks like there's a minor discrepancy between precompiled rule tables and dynamically generated ones--I'll investigate in another issue.

### DIFF
--- a/internal/server/server_test.go
+++ b/internal/server/server_test.go
@@ -45,6 +45,7 @@ type testParam struct {
 	store        storage.Store
 	policyLoader policyloader.PolicyLoader
 	schemaMgr    schema.Manager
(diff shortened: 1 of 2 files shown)


charithe (commented):
LGTM but the change seems to have broken some other tests 🤔

Sambigeara:
Yeah, sorry, I thought I'd converted it back to a draft yesterday but alas I did not

Sambigeara:
Ready now

charithe (approved):

PR #2452, opened 2025-01-16
fix: Reload rule table when store contents change

### LINKED ISSUE #2451: Rule table doesn't get updated when the store contents change
Changing the policies on disk (or on DB) doesn't refresh the rule table.

charithe:
This needs a proper unit test that I'll add in a different PR. In the
mean time, this should fix the E2E test failures.

Fixes #2451

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -160,8 +160,7 @@ func Start(ctx context.Context) error {
 
 	rt := ruletable.NewRuleTable().WithPolicyLoader(policyLoader)
 
-	// For now, we're only enabling the ruletable engine for non-mutable stores populate rule table
-	// for non-mutable stores
+	// Populate rule table for non-mutable stores.
 	if _, ok := store.(storage.MutableStore); !ok {
 		rps, err := policyLoader.GetAll(ctx)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


internal/server/server.go
```diff
@@ -171,10 +170,10 @@ func Start(ctx context.Context) error {
 		if err := rt.LoadPolicies(rps); err != nil {
 			return fmt.Errorf("failed to load policies into rule table: %w", err)
 		}
+	}
 
-		if ss, ok := store.(storage.Subscribable); ok {
-			ss.Subscribe(rt)
-		}
+	if ss, ok := policyLoader.(storage.Subscribable); ok {
+		ss.Subscribe(rt)
```
Sambigeara, line 176: https://github.com/cerbos/cerbos/issues/2451#issuecomment-2601702657

> I'm not sure I understand why this is the case. For mutable stores, the original stores remain as the source of truth, and a "partial rule table" is generated at query time (and discarded afterwards).

In the mutable-store case, partial rule tables are (currently) generated from the mutable store per-query and discarded afterwards. I chose to do this as an initial step as it seemed odd to maintain two sources of mutable truth.

This may or may not be related to the recent [E2E test failure](https://github.com/cerbos/cerbos/actions/runs/12859261034/job/35849465024). I'm looking in to that.

PR #2455, opened 2025-01-20
fix: Correctly handle partial rule table and event subscription

Sambigeara:
Previously, the code determining whether or not to generate a partial rule table was incorrect, as the rule table could never be nil (so it would always look for the precompiled table).

We don't subscribe the rule table to change events unless we're precompiling the full rule table at start time, as the evaluator builds a partial rule table from underlying store state at evaluation time.

This seems to fix the broken tests for the JS and Go SDKs when run locally. I'm having some trouble with the E2E test suite so have been unable to confirm those.

This'll hopefully fix the broken Admin API tests for now. In general, I think maintaining partial rule table generation is confusing--I'll look at unifying (all stores using a full rule table) in a follow up item.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -158,7 +158,7 @@ func Start(ctx context.Context) error {
 		return ErrInvalidStore
 	}
 
-	rt := ruletable.NewRuleTable().WithPolicyLoader(policyLoader)
+	var rt *ruletable.RuleTable
 
 	// Populate rule table for non-mutable stores.
 	if _, ok := store.(storage.MutableStore); !ok {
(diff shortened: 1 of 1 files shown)


charithe (approved):
Looks OK to me. One weird thing I noticed is that [this plan resources](https://github.com/cerbos/cerbos/blob/3da515625e783f36311f01814d3bcc99c5a092f1/hack/dev/plan.hurl#L46-L58) call returns `(false)` as `filterDebug` when tested against the disk store. However, the same set of policies in SQLite returns `<NO_MATCH>`. I think there might be some logic error somewhere in the handling of mutable stores? Could be looked at in a different PR, of course.

PR #2460, opened 2025-01-28
enhancement: Lazy rule table

Sambigeara:
Introduce a single rule table implementation which is empty on start and lazily loads required policies as they're queried. Enables this rule table for all store types.

Also solves some other issues:

* Makes query planner empty-case output consistent
* Fixes E2E tests
* Fixes principal policy `scopePermissions: *REQUIRES_PARENTAL_CONSENT` fall through behaviour

### DIFF
--- a/internal/engine/engine_test.go
+++ b/internal/engine/engine_test.go
@@ -45,62 +45,52 @@ func TestCheck(t *testing.T) {
 	eng, cancelFunc := mkEngine(t, param{auditLog: mockAuditLog, schemaEnforcement: schema.EnforcementNone})
 	defer cancelFunc()
 
-	ruleTableEngine, ruleTableCancelFunc := mkEngine(t, param{auditLog: mockAuditLog, schemaEnforcement: schema.EnforcementNone, enableRuleTable: true})
-	defer ruleTableCancelFunc()
-
-	engMap := map[string]*Engine{
-		"WithoutRuleEngine": eng,
-		"WithRuleEngine":    ruleTableEngine,
-	}
-
 	testCases := test.LoadTestCases(t, "engine")
 
-	for engPrefix, eng := range engMap {
-		for _, tcase := range testCases {
-			t.Run(fmt.Sprintf("%s/%s", engPrefix, tcase.Name), func(t *testing.T) {
-				tc := readTestCase(t, tcase.Input)
-				mockAuditLog.clear()
+	for _, tcase := range testCases {
+		t.Run(tcase.Name, func(t *testing.T) {
+			tc := readTestCase(t, tcase.Input)
+			mockAuditLog.clear()
 
-				traceCollector := tracer.NewCollector()
-				haveOutputs, err := eng.Check(context.Background(), tc.Inputs, WithTraceSink(traceCollector))
+			traceCollector := tracer.NewCollector()
+			haveOutputs, err := eng.Check(context.Background(), tc.Inputs, WithTraceSink(traceCollector))
 
-				if tc.WantError {
-					require.Error(t, err)
-				} else {
-					require.NoError(t, err)
-				}
+			if tc.WantError {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
 
-				for i, have := range haveOutputs {
-					// TODO(saml) I can't, for the life of me, figure out out to order this via a transformation
-					// function in `cmp.Diff` below, so this'll have to do for now
-					slices.SortStableFunc(have.Outputs, func(a, b *enginev1.OutputEntry) int {
-						if a.Src < b.Src {
-							return -1
-						} else if a.Src > b.Src {
-							return 1
-						}
-						return 0
-					})
+			for i, have := range haveOutputs {
+				// TODO(saml) I can't, for the life of me, figure out out to order this via a transformation
+				// function in `cmp.Diff` below, so this'll have to do for now
+				slices.SortStableFunc(have.Outputs, func(a, b *enginev1.OutputEntry) int {
+					if a.Src < b.Src {
+						return -1
+					} else if a.Src > b.Src {
+						return 1
+					}
+					return 0
+				})
 
-					require.Empty(t, cmp.Diff(tc.WantOutputs[i],
-						have,
-						protocmp.Transform(),
-						protocmp.SortRepeatedFields(&enginev1.CheckOutput{}, "effective_derived_roles"),
-					))
-				}
-
-				haveDecisionLogs := mockAuditLog.getDecisionLogs()
-
-				require.Empty(t, cmp.Diff(tc.WantDecisionLogs,
-					haveDecisionLogs,
+				require.Empty(t, cmp.Diff(tc.WantOutputs[i],
+					have,
 					protocmp.Transform(),
-					protocmp.IgnoreEmptyMessages(),
 					protocmp.SortRepeatedFields(&enginev1.CheckOutput{}, "effective_derived_roles"),
-					protocmp.SortRepeatedFields(&enginev1.Principal{}, "roles"),
-					protocmp.IgnoreFields(&auditv1.DecisionLogEntry{}, "call_id", "timestamp", "peer"),
 				))
-			})
-		}
+			}
+
+			haveDecisionLogs := mockAuditLog.getDecisionLogs()
+
+			require.Empty(t, cmp.Diff(tc.WantDecisionLogs,
+				haveDecisionLogs,
+				protocmp.Transform(),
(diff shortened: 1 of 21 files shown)


internal/engine/engine_test.go
```diff
@@ -45,62 +45,52 @@ func TestCheck(t *testing.T) {
...
+					} else if a.Src > b.Src {
+						return 1
+					}
+					return 0
+				})
 
-				require.Empty(t, cmp.Diff(tc.WantDecisionLogs,
-					haveDecisionLogs,
+				require.Empty(t, cmp.Diff(tc.WantOutputs[i],
+					have,
```
charithe, line 77: Imports stdlib cmp: `import stdcmp "cmp"`

```suggestion
				require.Empty(t, cmp.Diff(tc.WantOutputs[i],
					have,
					cmpopts.SortSlices(func(a, b *enginev1.OutputEntry) bool { return stdcmp.Less(a.Src, b.Src) }),

```

internal/engine/engine_test.go
```diff
@@ -45,62 +45,52 @@ func TestCheck(t *testing.T) {
...
+				// TODO(saml) I can't, for the life of me, figure out out to order this via a transformation
+				// function in `cmp.Diff` below, so this'll have to do for now
+				slices.SortStableFunc(have.Outputs, func(a, b *enginev1.OutputEntry) int {
+					if a.Src < b.Src {
+						return -1
+					} else if a.Src > b.Src {
+						return 1
+					}
+					return 0
+				})
```
charithe, line 74: See suggestion in the comment below.
```suggestion

```

internal/engine/engine_test.go
```diff
@@ -150,65 +140,55 @@ func TestCheckWithLenientScopeSearch(t *testing.T) {
...
```
charithe, line 170: See suggestion above.

internal/engine/evaluator.go
```diff
@@ -526,6 +524,19 @@ func (ppe *principalPolicyEvaluator) Evaluate(ctx context.Context, tctx tracer.C
 						}
 					}
 				})
+
+				if p.ScopePermissions == policyv1.ScopePermissions_SCOPE_PERMISSIONS_REQUIRE_PARENTAL_CONSENT_FOR_ALLOWS {
+					for _, a := range result.unresolvedActions() {
+						if _, ok := implicitlyAllowedActions[a]; !ok {
```
charithe, line 530: IIUC, any remaining actions in the `implicitlyAllowedActions` fall through to resource/role policy evaluation? 
Since one of the principal policies in the scope chain explicitly had a rule for those actions with `REQUIRE_PARENTAL_CONSENT`, shouldn't their decisions be resolved at this point?

Sambigeara:
I think it's prudent to add some tests cases for store updates (particularly WRT the various rule table caches), so I'll put this PR into `draft` until that's done.

internal/engine/ruletable/rule_table.go
```diff
@@ -94,27 +102,240 @@ type CelProgram struct {
...
+
+			if rps == nil {
+				registryBuffer[resourceModID] = false
+				// we used lenientScopeSearch, so we can assert that no policies exist for all child scopes
+				// TODO(saml) refactor
+				for i := len(scope) - 1; i >= 0; i-- {
+					if scope[i] == '.' || i == 0 {
+						registryBuffer[namer.ResourcePolicyModuleID(resource, policyVer, scope[:i])] = false
+					}
+				}
```
charithe, line 164: This loop gets used a few times in the code so it's probably worth making it an iterator function in the `namer` package and you can use it like `for parent := range namer.ScopeParents(scope) { ... }`.


```go
func ScopeParents(scope string) iter.Seq[string] {
	return func(yield func(string) bool) {
		for i := len(scope) - 1; i >= 0; i-- {
			if scope[i] == '.' || i == 0 {
				if !yield(scope[:i]) {
					return
				}
			}
		}
	}
}
```

internal/engine/evaluator.go
Sambigeara: You're right, I definitely got ahead of myself here.

> IIUC, any remaining actions in the `implicitlyAllowedActions` fall through to resource/role policy evaluation?

I think the expected behaviour of principal policies and `scopePermissions: REQUIRES_CONSENT` is not currently clear. Do we implicitly deny any missing actions or let them fall through to the normal (resource/role etc) evaluation? 

To try to use a practical use-case: a principal policy with `REQUIRES_CONSENT` suggests a vendor is defining a base principal override and then allowing their tenants to add narrowed customisations over that. With that in mind, it feels like we should require a matching rule to also exist within the tenant principal policies (and thus, not automatically fall through to the resource/role policies).

Frankly, I'm not really sure if tenant per-principal overrides are even a realistic use case, so should principal policies even support `REQUIRES_CONSENT`?

@alexolivier (sorry for the cold call) do you have any thoughts on this?

charithe: Scopes are not always about multi-tenancy. Since principal policies have scopes (for better or worse), they should have the same scope permission behaviour as the other policy types.

Sambigeara: > they should have the same scope permission behaviour as the other policy types

👍 

I think I misunderstood your original point: my current code is allowing actions defined in the principal policy `REQUIRES_CONSENT` scope to fall through to `resource/role` policies even if there's no matching rule in the principal policy `OVERRIDE` scopes, whereas we should `DENY` those rules at the end of the principal policy evaluation if there's no parent match?

charithe: Yes, they should be denied because parent scopes didn't have any explicit allows for them. 

I don't see anything getting removed from the `implicitlyAllowedActions` map when a rule matches. That's probably a logic bug as well.

internal/engine/engine_test.go
Sambigeara: I've tried this (and many variations of this) in the past. I reimplemented this again now and ran the tests a few time and the same un-deterministic ordering causes test failures:

```
--- FAIL: TestCheckWithLenientScopeSearch (0.08s)
    --- FAIL: TestCheckWithLenientScopeSearch/case_21 (0.00s)
        engine_test.go:152:
                Error Trace:    /Users/samuellock/cerbos/cerbos/internal/engine/engine_test.go:152
                Error:          Should be empty, but was   (*enginev1.CheckOutput)(Inverse(protocmp.Transform, protocmp.Message{
                                        "@type":                   s"cerbos.engine.v1.CheckOutput",
                                        "actions":                 map[string]protocmp.Message{"camelCase": {"@type": s"cerbos.engine.v1.CheckOutput.ActionEffect", "effect": s"EFFECT_ALLOW", "policy": string("resource.runtime_effective_derived_roles.vdefault")}, "snake_case": {"@type": s"cerbos.engine.v1.CheckOutput.ActionEffect", "effect": s"EFFECT_ALLOW", "policy": string("resource.runtime_effective_derived_roles.vdefault")}},
                                        "effective_derived_roles": Inverse(cmpopts.SortSlices, []string{"one", "two"}),
                                        "outputs": []protocmp.Message{
                                                {
                                                        "@type": s"cerbos.engine.v1.OutputEntry",
                                                        "src": strings.Join({
                                                                "resource.runtime_effective_derived_roles.vdefault#",
                                -                               "camelC",
                                +                               "snake_c",
                                                                "ase",
                                                        }, ""),
                                                        "val": protocmp.Message{"@type": s"google.protobuf.Value", "list_value": protocmp.Message{"@type": s"google.protobuf.ListValue", "values": []protocmp.Message{{"@type": s"google.protobuf.Value", "string_value": string("one")}, {"@type": s"google.protobuf.Value", "string_value": string("two")}}}},
                                                },
                                                {
                                                        "@type": s"cerbos.engine.v1.OutputEntry",
                                                        "src": strings.Join({
                                                                "resource.runtime_effective_derived_roles.vdefault#",
                                -                               "snake_c",
                                +                               "camelC",
                                                                "ase",
                                                        }, ""),
                                                        "val": protocmp.Message{"@type": s"google.protobuf.Value", "list_value": protocmp.Message{"@type": s"google.protobuf.ListValue", "values": []protocmp.Message{{"@type": s"google.protobuf.Value", "string_value": string("one")}, {"@type": s"google.protobuf.Value", "string_value": string("two")}}}},
                                                },
                                        },
                                        "request_id":  string("test"),
                                        "resource_id": string("test"),
                                  }))
```

internal/engine/ruletable/rule_table.go
Sambigeara: Cool! I remember looking for a way to do this in the past, didn't realise it had been added!

PR #2475, opened 2025-02-13
chore!: Role policy deny rows

Sambigeara:
Resource policies with `scopePermissions: REQUIRE_PARENTAL_CONSENT_FOR_ALLOWS` now inherit permissions of parent scopes rather than requiring explicit duplication of rules. This means that policy authors need only override any rules that deviate from the parent. They still can't add permissions which are not within parent scope policies--they can only narrow the set of permissions that are available within the parent scope hierarchy.

Role policies will behave as is: any resource/action pairs that are not referenced in a role policy will lead to implicit DENYs for the given role.

P.s. I'm happy to jump on a call to pair-program-review this change if it'll make things easier.


Sambigeara:
I've temporarily disabled `fail-on-error` for the Coveralls upload step in the GitHub workflows in lieu of [this ongoing outage](https://status.coveralls.io/incidents/0nz347snpb16). I can remove those bypasses in a later PR.

charithe (commented):
Could you please commit the variable name change you did during our huddle as well.

charithe (approved):

PR #2477, opened 2025-02-14
chore: Revert Coveralls GitHub workflow bypass

Sambigeara:
This reverts commit 5926145fda94a97fdeb4684af4a983207147a946.

The issue was [supposedly resolved 15 hours ago](https://status.coveralls.io/incidents/0nz347snpb16), so we should be safe to re-enable Coveralls.


charithe:
I think we can just leave them as they are. There could be future outages as well and Coveralls reporting is not important enough to break the build.

PR #2484, opened 2025-02-18
chore!: REQUIRE_PARENTAL_CONSENT refinements for resource and principal policies

Sambigeara:
This PR encompasses two notable changes.

#### Principal policies are now consistent with resource policies:

Principal policies with `scopePermissions: REQUIRE_PARENTAL_CONSENT` are now consistent with resource policy behaviour:
- If a rule is not matched for a given resource and action, it now falls through to parent scopes instead of instantly (implicitly) issuing a `DENY`.
- If a rule matches but the condition is not satisfied, the action is issued a `DENY`.

Unresolved resource/action pairs from a `REQUIRE_PARENTAL_CONSENT` principal policy scope can also fall through to the resource policy evaluator.

#### Conditional resource policy rules with `REQUIRE_PARENTAL_CONSENT` now implicitly narrow permissions:

Rules defined within resource policies with `scopePermissions: REQUIRE_PARENTAL_CONSENT` now implicitly DENY actions for matching roles if a condition is defined and not satisfied. This differs from default `scopePermissions: OVERRIDE_PARENT` behaviour, which will not evaluate the rule if the condition is not satisfied.

For the query planner, it was necessary to UN-inverse `REQUIRE_PARENTAL_CONSENT` conditional resource policy rules (back to the original conditional `ALLOW` rules) so that they could be logically `AND`ed with the parent rules.

### DIFF
--- a/internal/engine/evaluator.go
+++ b/internal/engine/evaluator.go
@@ -327,9 +327,9 @@ func (rte *ruleTableEvaluator) Evaluate(ctx context.Context, tctx tracer.Context
 
 					if satisfiesCondition { //nolint:nestif
 						roleEffectSet[row.Effect] = struct{}{}
-						roleEffectInfo.Scope = scope
 						if row.NoMatchForScopePermissions {
 							roleEffectInfo.Policy = noMatchScopePermissions
+							roleEffectInfo.Scope = scope
 						}
 
 						var outputExpr *exprpb.CheckedExpr
@@ -369,6 +369,7 @@ func (rte *ruleTableEvaluator) Evaluate(ctx context.Context, tctx tracer.Context
 						delete(roleEffectSet, effectv1.Effect_EFFECT_ALLOW)
 					case policyv1.ScopePermissions_SCOPE_PERMISSIONS_OVERRIDE_PARENT:
 						roleEffectInfo.Effect = effectv1.Effect_EFFECT_ALLOW
+						roleEffectInfo.Scope = scope
 						break scopesLoop
 					}
 				}
@@ -379,10 +380,15 @@ func (rte *ruleTableEvaluator) Evaluate(ctx context.Context, tctx tracer.Context
 				actionEffectInfo = roleEffectInfo
 			}
 
-			// Finalise and return the first independent ALLOW, if present
 			if roleEffectInfo.Effect == effectv1.Effect_EFFECT_ALLOW {
+				// Finalise and return the first independent ALLOW
 				actionEffectInfo = roleEffectInfo
 				break
+			} else if roleEffectInfo.Effect == effectv1.Effect_EFFECT_DENY &&
+				actionEffectInfo.Policy == noMatchScopePermissions &&
+				roleEffectInfo.Policy != noMatchScopePermissions {
+				// Override `noMatchScopePermissions` DENYs with explicit ones for clarity
+				actionEffectInfo = roleEffectInfo
 			}
 		}
 
--- a/internal/engine/evaluator.go
+++ b/internal/engine/evaluator.go
@@ -472,6 +472,11 @@ func (ppe *principalPolicyEvaluator) Evaluate(ctx context.Context, tctx tracer.C
 										result.Outputs = append(result.Outputs, output)
 										octx.ComputedOutput(output)
 									}
+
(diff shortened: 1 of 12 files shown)


Sambigeara:
@dbuduev wondered if you could sanity check the query planner logic (specifically the need to undo the rule inversion that occurs at build time for **conditional** resource policy ALLOW rules). I'm a little puzzled why I had to do this--perhaps the query planner was working as expected already prior to the change 🫠

dbuduev:
@Sambigeara, I think that the change to the planner is unnecessary. The condition will get into the DENY filter and be (re)inverted at that stage.

```go
func (p *PolicyPlanResult) add(filter *qpN, effect effectv1.Effect) {
	if effect == effectv1.Effect_EFFECT_ALLOW {
		p.allowFilter = append(p.allowFilter, filter)
	} else {
		p.denyFilter = append(p.denyFilter, invertNodeBooleanValue(filter))
	}
}
```

The parent scope condition for the ALLOW rule will get into the `allow` filter. If there is no ALLOW rule in the parent scope, the final plan will be an unconditional DENY, which we need.

The only question I have is whether, in the case of several ALLOW rules in the parent scope, any rule can give parental consent. If yes, I see a cartesian product of their conditions and the child scope condition. What do you think?

Sambigeara:
Thanks. I've tidied the logic somewhat.

I've updated a test suite because I realise the evaluation was exiting early if a scope returned a node for `ALLOW` or `DENY` (`OVERRIDE_PARENT`). This wouldn't cover the case where conditional rules might not be matched, so now the evaluator collects all potential nodes from all scopes and aggregates the result. I think this is the expected behaviour.

I've added some special-case handling in the inversion logic to cater for the inverted `REQUIRE_PARENTAL_CONSENT` conditional resource policy nodes.

Would appreciate another look when you had a change @dbuduev !

Sambigeara:
Thanks. I reckon I'm there now. Would appreciate another look when you get the change @dbuduev !

dbuduev:
> I'm not sure I understand this 🤔 . Would an `AND` between the `REQUIRE_PARENTAL_CONSENT` rule and the (`OR`d) parent `OVERRIDE` rules not be sufficient?

You're right. In this case, it is sufficient. What if there are several `REQUIRE_PARENTAL_CONSENT` rules?

dbuduev (dismissed):

Sambigeara:
> What if there are several REQUIRE_PARENTAL_CONSENT rules?

They're [aggregated for the given scope as normal](https://github.com/cerbos/cerbos/blob/3474ef7794d048a2d26f52b6160423bd7651b078/internal/engine/planner/planner.go#L410-L423)

dbuduev:
I see. `OR`d `REQUIRE_PARENTAL_CONSENT` rules `AND` `OR`d `OVERRIDE` rules should be enough.

I've just noticed https://github.com/cerbos/cerbos/blob/3474ef7794d048a2d26f52b6160423bd7651b078/internal/engine/planner/planner.go#L421

Shouldn't the DENY rules be aggregated using `AND`?

Sambigeara:
> Shouldn't the DENY rules be aggregated using AND?

With the following rules:
https://github.com/cerbos/cerbos/blob/3474ef7794d048a2d26f52b6160423bd7651b078/internal/test/testdata/query_planner/policies/resource_policies/leave_request.yaml#L117-L131

I think the following test expectations are correct: https://github.com/cerbos/cerbos/blob/3474ef7794d048a2d26f52b6160423bd7651b078/internal/test/testdata/query_planner/suite/maggie.yaml#L53-L67

If either `R.attr.deleted` or `R.attr.hidden` were true, we'd want to `DENY`. Or: we `ALLOW` if neither or them are true.

So `ALLOW` if `NOT(R.attr.deleted) AND NOT(R.attr.hidden)`. Isn't the equivalent of that: `NOT(R.attr.deleted OR R.attr.hidden)` (e.g. what's in the passing test expectation)?

dbuduev:
> So ALLOW if NOT(R.attr.deleted) AND NOT(R.attr.hidden). Isn't this the equivalent of NOT(R.attr.deleted OR R.attr.hidden), e.g. what's in the passing test expectation?

I missed the top `NOT`. Thanks for the clarification!

dbuduev (approved):

Issue #2469, opened 2025-02-07
Switch to ContextEval to evaluate CEL expressions


dbuduev:
Extracted from https://github.com/cerbos/cerbos/issues/2379

Depends on fixing https://github.com/google/cel-go/issues/1124

PR #2495, opened 2025-02-27
enhancement!: Switch to ContextEval to evaluate CEL expressions

dbuduev:
Fixes https://github.com/cerbos/cerbos/issues/2469.
Also, upgrade cel-go to v0.24.1. That required disabling the `DefaultUTCTimeZone` feature and replacing deprecated declarations.


internal/conditions/cel.go
```diff
@@ -40,24 +41,25 @@ var (
...
+		decls.NewVariable(CELGlobalsAbbrev, celtypes.NewMapType(celtypes.StringType, celtypes.DynType)),
 	}
 
 	StdEnvOptions = []cel.EnvOption{
 		ext.TwoVarComprehensions(),
 		cel.CrossTypeNumericComparisons(true),
 		cel.Types(&enginev1.Request{}, &enginev1.Request_Principal{}, &enginev1.Request_Resource{}, &enginev1.Runtime{}),
-		cel.Declarations(StdEnvDecls...),
+		cel.VariableDecls(StdEnvDecls...),
+		cel.DefaultUTCTimeZone(false),
```
charithe, line 62: Why is it necessary to explicitly disable this? The tests seem to pass without it 🤔 Also, apparently this won't be allowed after 0.25 (https://github.com/google/cel-go/pull/1130).

dbuduev: > Why is it necessary to explicitly disable this?

On my machine, `TestSatisfiesCondition` fails without this line. 🤔 
```
engine_test.go:417:   condition > conditionAll > condition#5 > expr=`timestamp(R.attr.lastAccessed).getHours() == 10`
            activated
            result → false
```

AFAIU, https://github.com/google/cel-go/pull/1130 is part of v0.24.0 release. It turns this feature on by default. My PR turns it off.

> The use of DefaultUTCTimeZone(false) will only be permitted until release v0.25.0

We will have to come up with a solution by then.

dbuduev: The mentioned test seems to be unrelated to this feature. I can investigate why it fails.

charithe: Actually it fails on mine too (I seem to have had a cached test result). It's still worth trying to figure out why because we'll have the same problem again once 0.25 is released.

dbuduev: With the `DefaultUTCTimeZone` enabled, all timestamp getters—getHours, getMinutes, getDayOfWeek, and similar—take a timezone parameter. If omitted, the 'UTC' timezone is used by default.
https://github.com/google/cel-go/blob/3b3a43818bba0cb1011948d29822db0dd7871f33/common/stdlib/standard.go#L567

```repl
-> timestamp("2021-04-20T10:05:20.021-05:00").getHours("UTC")
_ = 15

-> timestamp("2021-04-20T10:05:20.021-05:00").getHours("-05:00")
_ = 10

-> timestamp("2021-04-20T10:05:20.021-05:00").getHours()
_ = 15
```
This is why the test fails.

charithe: Nice! Let's keep default UTC enabled and add a note to the docs about the timezone requirement for time functions.

dbuduev: Done. I reckon this is a breaking change. Should we put a warning?

charithe: Thanks. I see you've already added a note about it. I'll mention it in the release notes as well.

charithe (approved):

PR #2496, opened 2025-02-27
chore: Decouple role policies from scope permissions

Sambigeara:
Role policies are now agnostic to scope permissions. They can reside inside any scope and exhibit the same consistent behaviour: they inherit and narrow their parent roles.

If a parent role points to another role policy, then it can only ALLOW the maximum set of permissions specified in that parent role. This is evaluated recursively.

If a parent is an IdP role, this IdP role can have rules present in the same scope or inside a parent scope. Those policies follow the standard scope permissions based evaluation patterns.

I've also removed the restriction on scopePermission/scope ordering so scopes with `scopePermissions: REQUIRE_PARENTAL_CONSENT` are no longer constrained to the leaf nodes. Conversely, a scope with `OVERRIDE_PARENT` can now theoretically be a descendent of an `REQUIRE_PARENTAL_CONSENT` scope (although I'm unaware of a reasonable use-case for this).

TODO:
- [x] Add a test case for out of order scope permissions scenarios


Sambigeara:
> if the `parentRole` for a role policy doesn't exist, the role policy doesn't apply at all

It does apply, but it would only do so if the named role was a role that had `ALLOW` rules in parent scopes, so using a custom role name that wasn't referenced elsewhere would be a no-op. I just tested by removing the parent role in one of our test policies and changing the `role` to one referenced in a resource policy up the chain, and it behaved as expected (narrowing that role's permissions).

Is this what you meant?

charithe:
I can't reproduce it anymore. I must have done something wrong. Sorry I confused you.

While trying to recreate the issue, I noticed that the response metadata attributes the decision to the resource policy even though it clearly came from the role policy. I think, that's worth fixing in a different PR.


Policies:

`res_foo.yaml`

```yaml
---
apiVersion: api.cerbos.dev/v1
resourcePolicy:
  version: default
  resource: foo
  rules:
    - actions:
        - use
      effect: EFFECT_ALLOW
      roles:
        - user
      condition:
        match:
          all:
            of:
              - expr: 'P.attr.org == "foo"'
              - expr: "P.scope == R.scope"
```

`res_foo.a.yaml`

```yaml
---
apiVersion: api.cerbos.dev/v1
resourcePolicy:
  version: default
  resource: foo
  scope: a
  rules:
    - actions:
        - use
      effect: EFFECT_ALLOW
      roles:
        - user
      condition:
        match:
          all:
            of:
              - expr: 'P.attr.org == "foo"'
              - expr: 'P.scope == "a"'
              - expr: 'R.scope == "a"'
```

`role_user.yaml`

```yaml
---
apiVersion: api.cerbos.dev/v1
rolePolicy:
  role: user
  parentRoles: ["bah"]
  scope: a
  rules:
    - resource: foo
      allowActions:
        - use
      condition:
        match:
          expr: 'R.attr.owner != "john"'
```

Request:

```json
{
    "requestId": "test",
    "principal": {
        "id": "donald_duck",
        "policyVersion": "default",
        "scope": "a",
        "roles": [
            "user"
        ],
        "attr": {
            "org": "foo"
        }
    },
    "resources": [
        {
            "actions": [
                "use",
                "create"
            ],
            "resource": {
                "kind": "foo",
                "scope": "a",
                "policyVersion": "default",
                "id": "XX125",
                "attr": {
                    "owner": "john"
                }
            }
        }
    ],
    "includeMeta": true
}
```

Response:

```json
{
  "requestId": "test",
  "results": [
    {
      "resource": {
        "id": "XX125",
        "kind": "foo",
        "policyVersion": "default",
        "scope": "a"
      },
      "actions": {
        "create": "EFFECT_DENY",
        "use": "EFFECT_DENY"
      },
      "meta": {
        "actions": {
          "create": {
            "matchedPolicy": "NO_MATCH_FOR_SCOPE_PERMISSIONS",
            "matchedScope": "a"
          },
          "use": {
            "matchedPolicy": "resource.foo.vdefault/a"
          }
        }
      }
    }
  ],
  "cerbosCallId": "01JN6C6YTM06CFHQSH7B7JWEMM"
}
```

The `matchedPolicy` for action `use` should be the role policy because it's the one that generated the implicit DENY.

charithe (approved):

PR #2504, opened 2025-03-04
chore: Replace `golang.org/x/exp/maps` with stdlib `maps`

Juneezee:
#### Description

The experimental functions in `golang.org/x/exp/maps` are now available in the standard library in Go 1.21.

Reference: https://go.dev/doc/go1.21#maps

#### Checklist 

### DIFF
--- a/internal/storage/util.go
+++ b/internal/storage/util.go
@@ -11,13 +11,6 @@ import (
 
 const MaxPoliciesInBatch = 25
 
-func minInt(a, b int) int {
-	if a < b {
-		return a
-	}
-	return b
-}
-
 func BatchLoadPolicy(
 	ctx context.Context,
 	maxPoliciesInBatch int,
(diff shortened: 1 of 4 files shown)


internal/storage/util.go
```diff
@@ -27,7 +20,7 @@ func BatchLoadPolicy(
 ) error {
 	for idx := range ids {
 		if idx%maxPoliciesInBatch == 0 {
-			idxEnd := minInt(idx+maxPoliciesInBatch, len(ids))
+			idxEnd := min(idx+maxPoliciesInBatch, len(ids))
```
Juneezee, line 23: Go 1.21 added a new built-in `min` function so we don't need the `minInt` helper function anymore

https://go.dev/ref/spec#Min_and_max

haines (approved):
Nice one, thanks!

PR #2515, opened 2025-03-14
chore: Add gopls's modernizer to linters

haines:
https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize


dbuduev (approved):
👍🏻

PR #2521, opened 2025-03-19
feat(plan): Cross-scope query plan

dbuduev:


api/public/cerbos/engine/v1/engine.proto
```diff
@@ -59,6 +59,56 @@ message PlanResourcesInput {
   bool include_meta = 6;
 }
 
+message CrossScopePlanResourcesInput {
```
charithe, line 62: I think we could probably add an extra `repeated string scopes` field to the existing `PlanResourcesInput.Resource` message and make it mutually exclusive with the `scope` field (or merge both fields when processing the request). That way, we  can keep using the same `PlanResources` RPC endpoint  -- which is simpler for the users to understand.

internal/engine/planner/merge.go
```diff
@@ -0,0 +1,114 @@
...
+		output := outputs[scope]
+		switch output.Filter.Kind {
+		case enginev1.PlanResourcesFilter_KIND_ALWAYS_ALLOWED:
+			expressions = append(expressions, mkScopeOp(Equals, scope))
+		case enginev1.PlanResourcesFilter_KIND_ALWAYS_DENIED:
+			nots = append(nots, scope)
+		case enginev1.PlanResourcesFilter_KIND_CONDITIONAL:
+			conds[output.FilterDebug] = append(conds[output.FilterDebug], scope)
+		case enginev1.PlanResourcesFilter_KIND_UNSPECIFIED:
+			panic("unspecified filter kind")
```
charithe, line 34: Let's return an error here instead of panicking because it's easy to forget to update this code path during a change and it'd go unnoticed for a while until someone hits it in production.

charithe (commented):
LGTM. But, let's try to use the same RPC endpoint instead of multiple, specialised ones.

api/public/cerbos/engine/v1/engine.proto
dbuduev: What to do with metadata? For a cross-scope plan, I currently return a map—an instance of PlanResources metadata for each scope in the request.

charithe: The `filter_debug` field would be the combined plan. We can introduce an extra `matched_scopes` field to the `Meta` message to contain the set of matched scopes.

PR #2538, opened 2025-04-02
fix: Purge rule table on index build failure

Sambigeara:
When an index build fails, ensure the rule table is purged to avoid operating on stale data. Adds EventDisableRuleTable event type and makes rule table rebuild from scratch on next policy event.

Also handles an issue in mutable stores where intended no-ops were generating invalid queries and breaking evaluations.

### DIFF
--- a/internal/engine/ruletable/rule_table.go
+++ b/internal/engine/ruletable/rule_table.go
@@ -48,6 +48,7 @@ type RuleTable struct {
 	// role policies are per-scope, so the maps takes the form `map[scope]map[role][]roles`
 	parentRoles              map[string]map[string][]string
 	parentRoleAncestorsCache map[string]map[string][]string
+	awaitingHealthyIndex     chan struct{}
 	mu                       sync.RWMutex
 }
 
@@ -110,6 +111,7 @@ func NewRuleTable(policyLoader policyloader.PolicyLoader) *RuleTable {
 		scopeScopePermissions:    make(map[string]policyv1.ScopePermissions),
 		parentRoles:              make(map[string]map[string][]string),
 		parentRoleAncestorsCache: make(map[string]map[string][]string),
+		awaitingHealthyIndex:     make(chan struct{}, 1),
 		policyLoader:             policyLoader,
 	}
 }
@@ -955,6 +957,9 @@ func (rt *RuleTable) SubscriberID() string {
 func (rt *RuleTable) OnStorageEvent(events ...storage.Event) {
 	for _, evt := range events {
 		switch evt.Kind {
+		case storage.EventDisableRuleTable:
+			rt.log.Debug("Attempting to disable rule table")
+			rt.setIndexUnhealthy()
 		case storage.EventReload:
 			rt.log.Info("Purging ruletable")
 			rt.purge()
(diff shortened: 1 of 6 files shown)


charithe (commented):
Just looking at this on a mobile at the moment so apologies for asking questions instead of testing them out. Don't we have a similar problem with database (i.e. non-indexed) stores as well? It's possible to save an invalid policy to a database as well so how are those stores immune from the problem? If they are, can we use the same mechanism for indexed stores as well?

Sambigeara:
> Just looking at this on a mobile at the moment so apologies for asking questions instead of testing them out. Don't we have a similar problem with database (i.e. non-indexed) stores as well? It's possible to save an invalid policy to a database as well so how are those stores immune from the problem? If they are, can we use the same mechanism for indexed stores as well?

Yes! I've been working on this along with another issue related to mutable stores, I'm going to bundle up both fixes into this PR--hopefully won't be long. Apologies, forgot to put this into Draft.

Sambigeara:
I've checked this against disk, git, blob, and all three mutable stores. All now handle invalid indexes correctly. Non-mutable stores continue to return/act on the previous valid index state, mutable stores return errors until the policy set is fixed (I presume this was pre-existing behaviour?).

internal/engine/ruletable/rule_table.go
```diff
@@ -48,6 +48,7 @@ type RuleTable struct {
 	// role policies are per-scope, so the maps takes the form `map[scope]map[role][]roles`
 	parentRoles              map[string]map[string][]string
 	parentRoleAncestorsCache map[string]map[string][]string
+	awaitingHealthyIndex     chan struct{}
```
charithe, line 51: I think that an [atomic bool](https://pkg.go.dev/sync/atomic#Bool) would do the trick here? Channels tend to fail in surprising ways and I am a bit wary of using them.

charithe (commented):
I see the dilemma with the differing behaviour of stores and the way they deal with them. However, since we are kind of "doing surgery" anyway by introducing a new event as a stop gap measure, it's probably worth taking a look at fixing the underlying issue itself? I don't think we have a show-stopper that requires an immediate fix in prod so we can take a bit more time to sort things out.

From a cursory look, I think it might be the case that we could forward on the event from all sources as long as we add a new field to the event to indicate the state of the store (compiling/not-compiling/unknown -- I think it needs to be ternary because not all stores compile policies on the fly). The subscribers can inspect that field if they are doing anything that requires a valid store. It should be fairly easy to find all the subscription points by looking up implementations of `OnStorageEvent` so I am fairly confident that we can locate and patch all the places 🤞🏽 WDYT?

Sambigeara:
> "doing surgery"
😆 

I entirely agree! I'd thought that the fix was more time pressured given it's in production, but if that's not the case then absolutely, I'll dig in. There's a chance that the rule table might not need to know the state of the index health and forwarding events might be enough. Let's see 🤷‍♂️

Sambigeara:
I don't think there was need for ternary state. In practice, only the schema manager and the rule table subscribed to the compile manager. I handled appropriately in the rule table and bypassed those events in the schema manager to maintain behaviour.

internal/schema/schema.go
```diff
@@ -259,6 +259,9 @@ func (m *manager) SubscriberID() string {
 
 func (m *manager) OnStorageEvent(events ...storage.Event) {
 	for _, event := range events {
+		if event.IndexUnhealthy {
```
charithe, line 262: I don't think this is necessary because we can still continue to load schemas even if the index is unhealthy. 

- If it's an update to an existing schema, we should apply it because we may never get the schema update event again because those are rare.
- If it's a new schema, there's no harm in adding it to the schema cache.

charithe (commented):
I think there's a subtle bug in the schema manager. Other than that, LGTM 👍🏽

charithe (approved):

charithe (approved):

PR #2543, opened 2025-04-07
feat(plan): Add support for multiple actions

dbuduev:
Add support for multiple actions in the query plan API.  The produced query plan is a logical AND of the individual query plans for each action. It can be used to obtain a list of resources on which a principal can perform all the listed actions.

In other words, the combined query plan represents the intersection of permissions across all requested actions.


api/public/cerbos/request/v1/request.proto
```diff
@@ -30,15 +30,34 @@ message PlanResourcesRequest {
...
+      min_items: 0
+      unique_items: true
+      example: "[\"view:public\", \"edit:profile\"]"
+    }
+  ];
+  option (buf.validate.message).cel = {
+    id: "exclusiveFieldsActionOrActions"
+    expression: "has(this.action) && !has(this.actions) || !has(this.action) && size(this.actions) > 0"
+    message: "Exactly one of 'action' or 'actions' field must be set"
+  };
```
Sambigeara, line 59: I did not know you could do this!! Cool

Sambigeara (commented):
I think there's an opportunity here to pass the list of actions to the `EvaluateRuleTableQueryPlan` and handle the intersection of the results in there, rather than generating separate requests and handling the intersection afterwards.

I made some further changes to the method in my [latest PR](https://github.com/cerbos/cerbos/pull/2544). I think (once merged) we'd need only add an outer loop for actions, evaluate each, and `AND` them at the end. This would mean we wouldn't need the additional merge logic, and would also be considerably less expensive as many actions would still only be a single evaluation.

What do you think?

api/public/cerbos/request/v1/request.proto
```diff
@@ -30,15 +30,34 @@ message PlanResourcesRequest {
...
   string action = 2 [
-    (buf.validate.field).required = true,
-    (buf.validate.field).string.min_len = 1,
-    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).string.min_len = 0,
     (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
       description: "Action to be applied to each resource in the list."
       example: "\"view:public\""
     }
   ];
```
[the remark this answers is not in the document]
dbuduev: The existing API works for most users, so we want to keep it while adding support for more advanced use cases of many actions and scopes.

dbuduev:
> I think there's an opportunity here to pass the list of actions to the `EvaluateRuleTableQueryPlan` and handle the intersection of the results in there, rather than generating separate requests and handling the intersection afterwards.
> 
> I made some further changes to the method in my [latest PR](https://github.com/cerbos/cerbos/pull/2544). I think (once merged) we'd need only add an outer loop for actions, evaluate each, and `AND` them at the end. This would mean we wouldn't need the additional merge logic, and would also be considerably less expensive as many actions would still only be a single evaluation.
> 
> What do you think?

I like it in principle, but the merge logic not only `ANDs` the plans but attempts to simplify the final one. Currently, it only eliminates duplicates, but other tactics might be employed later. If the simplification fits in the `EvaluateRuleTableQueryPlan`  I see no cons in moving the merge logic.

What about cross-scope evaluation?

api/public/cerbos/request/v1/request.proto
```diff
@@ -30,15 +30,34 @@ message PlanResourcesRequest {
...
+  repeated string actions = 7 [
+    (buf.validate.field).repeated = {
+      unique: true
+      items: {
+        string: {min_len: 1}
+      }
+      min_items: 0
+    },
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+      description: "Actions to be applied to each resource in the list. Use this field instead of the deprecated action field. Must contain at least one action and all actions must be unique."
```
charithe, line 49:
```suggestion
      description: "List of actions to generate the query plan for. Mutually exclusive with the singular action field. Must contain at least one action and all actions must be unique."
```

api/public/cerbos/request/v1/request.proto
```diff
@@ -30,15 +30,34 @@ message PlanResourcesRequest {
...
     }
   ];
 
+  repeated string actions = 7 [
+    (buf.validate.field).repeated = {
+      unique: true
+      items: {
+        string: {min_len: 1}
+      }
+      min_items: 0
```
charithe, line 46: Let's add a `max_items` restriction as well. About 20 seems reasonable.

api/public/cerbos/response/v1/response.proto
```diff
@@ -34,6 +34,7 @@ message PlanResourcesResponse {
       description: "Policy scope that matched to produce this effect."
       example: "\"acme.corp.base\""
     }];
+    map<string, string> matched_scope_per_action = 3 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Matched scope for each action."}];
```
charithe, line 37:
```suggestion
    map<string, string> matched_scopes = 3 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Matched scope for each action."}];
```

I think we should populate this field even when it's a singular `action` in the request. Then we can slowly deprecate the `matched_scope` field and fully transition to this new field at some point. In fact, in the follow-up task of updating the SDKs, we should start using the new fields instead of the old ones.

api/public/cerbos/request/v1/request.proto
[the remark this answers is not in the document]
charithe: Sorry, it's entirely my fault that I waffled on this. Since `actions` can be used in place of `action`, I think it'd be clearer if we add the `deprecated` label. It won't affect the existing users in any way unless they generate gRPC code themselves. It'd be more of a reminder to ourselves to clean it up later.

Sambigeara:
> If the simplification fits in the EvaluateRuleTableQueryPlan I see no cons in moving the merge logic.

For parity with this implementation (multiple actions alone), I think we could adapt and apply your function at the merge point in the action-scope of the traversal (by inspecting for constant boolean nodes and making decisions based on truthiness etc). Might even be able to apply to other node merge points too, but that's probably not a "now" problem 😆

> What about cross-scope evaluation?

I **_think_** we can do something pretty neat here, too 🤔. To begin with, we could add an additional loop just outside the existing (inner) scope loop, something like `scopeSets` or `crossScopes`; representing a full traversal of each scope for each cross-scope passed in, OR'ing the output nodes from each..?

Actually, what's really exciting and also terrifying is that I think there's opportunity to optimise further by only evaluating unique branches in the scope tree once, so sharing partial scope-chain evaluations _across_ cross-scopes (but I think this could be a later optimisation because the thought of it is already shivering my timbers).

Sambigeara:
The change in question has been merged now FYI: https://github.com/cerbos/cerbos/pull/2544

dbuduev:
@Sambigeara, please have another look. I've moved the loop over actions to `EvaluateRuleTableQueryPlan `.

Sambigeara (approved):
Nice!

Sambigeara (approved):

PR #2544, opened 2025-04-09
feat: Add principal policy support to rule table

Sambigeara:
Extend rule table implementation to handle principal policies by adding a new policy type, separate scope maps for resource and principal policies, and methods to load and evaluate principal policies alongside resource policies.

There's decent scope for further streamlining, but I don't want to make this PR more difficult to parse than it already is. Ongoing refactors can be done in future PRs. Re-introducing tracing is one of these to-dos, for example.


charithe (approved):
LGTM. I didn't do any tests because I have spotty Wifi but I trust you did 😄

charithe (approved):

PR #2549, opened 2025-04-14
fix: Return appropriate backoff in logcap ingest error path

Sambigeara:
Ensure logcap ingest errors always return with an appropriate backoff, either server-provided or a client enforced exponential backoff for severe failures like network issues.

Add upper limit to audit ingest MaxBatchSize.

Requires https://github.com/cerbos/cloud-api/pull/284


Sambigeara:
Lint will fail until the cloud API PR is merged and I can target the release.

internal/audit/hub/ingest.go
```diff
@@ -50,13 +52,36 @@ func NewIngestSyncer(logger *zap.Logger) (*Impl, error) {
...
+					underlying: err,
+					Backoff:    backoff,
+				}
+			} else if err != nil {
+				// Apply fixed default backoff for severe failures (typically network issues)
+				// where the server couldn't communicate a specific backoff duration.
+				// Using a conservative fixed backoff rather than exponential backoff as
+				// these errors are expected to be transient network issues and retrying
+				// too aggressively could exacerbate the problem (and because an exponential
+				// backoff here feels entirely overkill).
```
haines, line 71: I'm not totally convinced by this line of reasoning; the point of exponential backoff strategies is to mitigate exactly the issue you're describing, isn't it?

We can tune the parameters of the exponential backoff so that there is a sensible minimum interval, but in the face of an outage we don't want every PDP connected to Hub to be constantly retrying at the same rate, so an exponential backoff with jitter is definitely appropriate here.

Sambigeara: You're right. I'd estimated the maintenance complexity probably wasn't worth the effort given our current usage and mechanisms, but it's actually not that complex and is sensible. Our couple of cows today could become a thundering herd tomorrow 🙂

PR #2558, opened 2025-04-16
feat(audit): Add size-based batch limiting to audit log hub

Sambigeara:
Added batch size limiting by bytes to prevent oversized ingest batches. This implementation tracks accumulated entry sizes and creates new batches when the configured maximum batch size is reached.

Oversized logs are an unlikely case, and therefore, initial batching is still carried out according to `nEntries`. Byte size batching is carried out after on these pre-batched lists of entries. An argument can be had for a larger refactor to simultaneously batch by `nEntries` or by `byteSize`, however the cost:reward ratio seems heavily skewed in favour of the former. Byte size batching is more a protection rather than a primary batching mechanism.

It's difficult to know how best to handle oversized entries. Any log above a (unreasonably large) certain size can't be processed by the backend, so attempting to `Sync` them is pointless. I've introduced a new metric that users can make use of for alerting purposes.


haines:
I feel like number-of-entries batching is not terribly useful, why don't we just get rid of it altogether and only batch by size? Basically as an operator I feel like I'm only really interested in two things: my audit logs get flushed in a timely fashion (hence I want the flush interval not to be too long) and my batches aren't overwhelmingly large when sent over the wire. Why would the number of entries in the batch matter?

Sambigeara:
I think that inspecting the size of each log is an unfortunate additional computation which we didn't have to carry out before. I thought about trying to put in a preliminary size check on the entire batch to avoid iterating over each log, but that would require additional refactoring (again, it comes back to the `cost:reward` ratio). I think it's prudent to keep performance at the forefront when weighing up these implementation decisions.

I think this change is sound enough--we're protected from oversize batches and can consider further improvements down the road (ideally one that doesn't increase CPU time). It's low cost and doesn't introduce any notable technical debt.

internal/audit/hub/conf.go
```diff
@@ -40,6 +43,8 @@ type Conf struct {
 type IngestConf struct {
 	// MaxBatchSize defines the max number of log entries to send in each Ingest request.
 	MaxBatchSize uint `yaml:"maxBatchSize" conf:",example=32"`
+	// MaxBatchSizeBytes defines the max cumulative size in bytes for a batch of log entries.
+	MaxBatchSizeBytes uint `yaml:"maxBatchSizeBytes" conf:",example=2097152"`
```
charithe, line 47: I don't think it's useful to have this configurable by the end users because the limit is ultimately dictated by us. It could be a hidden config such as [`engine.numWorkers`](https://github.com/cerbos/cerbos/blob/380d8f6553a2dccfbd52b8a91823d06231799485/internal/engine/conf.go#L27) that we can use as an escape hatch though.

haines:
Sorry, I don't understand. Isn't the number of `SizeVT()` calls we have to make the same in either case? I feel like having both number and size based batching just adds complexity; and isn't there a performance cost to sending many small requests vs a few bigger ones?

charithe (commented):
How about storing additional metadata on Badger when we actually persist the audit entries locally? I think we need to store the size of each entry and a flag to indicate very large entries. That way, when batching up the logs for sending, we can looks at the size metadata that's already stored. (In the future, we could even do something clever with it like bin packing.) If there's an oversized value that we cannot accept in the backend, we should mark it as skipped so that we don't keep on trying to send it over and over again each time. (In a future improvement, we could try to update the RPC to accept chunked entries so that logs can be just split into even-sized chunks and sent over the wire to be re-assembled at the backend.)

Sambigeara:
> Sorry, I don't understand. Isn't the number of `SizeVT()` calls we have to make the same in either case?

Yeah, but the solution we have(?) to move to for now is a degradation in performance, so ripping out the older code seems counter-intuitive given we could come back and improve on it down the line. Removing the `nEntries` batching would give negligible (if any) performance gains, for more work.

> How about storing additional metadata on Badger when we actually persist the audit entries locally? I think we need to store the size of each entry and a flag to indicate very large entries. That way, when batching up the logs for sending, we can looks at the size metadata that's already stored.

This is a clever idea! I'd be concerned about implications on more critical query time performance, but actually the logs get written in a separate goroutine so hopefully will largely occur during idle times?

We could take it a step further and append the byte-size to the end of the `Sync` keys, then at prefix-scan time, aggregate the total size and cut early if we exceed the limit. This way, we still only need a single write transaction with Badger for a single, pre-calculated batch. We could also just omit any single records that exceed the limit before writing the Sync key to Badger at all (the record itself will persist, but it'll bypass Hub sync).

charithe:
How about filtering the entries at the time of writing to the local log? I don't think we need to preserve the original because the users don't really go into the local store and look at the raw data.

Sambigeara:
I think there's an edge case that might catch us out if we did that:

> but there may be preexisting logs that exist at PDP start time (retrieved from disk) after masking configuration has changed. We could then also filter at sync time, but by that point, we'd already have skipped those newly reduced logs.

I might be overcomplicating, though 🤔

charithe:
I don't think we need to worry about that. It's reasonable to say that we apply the masking rules that were in effect at the time the audit entry was generated. We don't want to end up in a situation where we have to constantly scan through all the rejected records and see if the new masking rules make them transmittable again. 

One thing we could do here to avoid complete "data loss" is to send a truncated version of the oversized entry so that at least some of the information is still recorded by the backend. 
- Add a new field to the log entry proto to indicate that it's oversized. We can use that field to highlight the problem in the UI.
- The truncation algorithm could be one of:
   - Send only the "important" fields like decision, resource, principal and strip out everything else.
   - Starting with the largest populated field of the entry, drop fields until the whole entry is small enough to send.

Sambigeara:
I've chosen some arbitrary (but I believe sensical) masking defaults for the oversized entries. I opted to just reuse the existing filter mechanism.

We can always revisit down the line if we needed something more sophisticated (opt-in or sized based field removals).

internal/audit/hub/conf.go
```diff
@@ -42,6 +47,8 @@ type Conf struct {
 type IngestConf struct {
 	// MaxBatchSize defines the max number of log entries to send in each Ingest request.
 	MaxBatchSize uint `yaml:"maxBatchSize" conf:",example=32"`
```
charithe, line 49: Shall we hide this as well because it doesn't make much sense for the users to set it? In fact, I think we should ignore it completely in the code because trying to satisfy both the `batchSize` and 'batchSizeBytes` constraints makes it difficult to do things efficiently. In the case of a busy PDP, trying to stick to the `batchSize` limit means we'd be making a lot more API calls than necessary.

internal/audit/hub/hub.go
```diff
@@ -237,28 +309,40 @@ func (l *Log) streamPrefix(ctx context.Context, kind logsv1.IngestBatch_EntryKin
...
 		}
 		defer keysPool.Put(&keys)
 
 		var i int
+		var batchSizeBytes uint64
 		for it.Seek(prefix); it.ValidForPrefix(prefix); it.Next() {
 			item := it.Item()
 
+			// retrieve byte-size from key
+			size, err := strconv.ParseUint(string(item.Key()[local.KeyByteSizeStart:local.KeyByteSizeEnd]), 16, 64)
```
charithe, line 324: Won't this fail if there are existing, unsynced keys that don't have the size encoded in the key?

internal/audit/local/badgerdb.go
```diff
@@ -402,6 +416,14 @@ func GenKey(prefix []byte, id audit.IDBytes) []byte {
 	return key[:]
 }
 
+func GenKeyWithByteSize(prefix []byte, id audit.IDBytes, nBytes int) []byte {
+	key := GenKey(prefix, id)
+	hex := fmt.Sprintf("%0*x", keyByteSizeLen, nBytes)
```
charithe, line 421: I am curious why the size is encoded as hex. Couldn't we just store it encoded as Big Endian using [encoding/binary](https://pkg.go.dev/encoding/binary)?

internal/audit/hub/filter.go
```diff
@@ -69,6 +70,29 @@ func NewAuditLogFilter(conf MaskConf) (*AuditLogFilter, error) {
 	}, nil
 }
 
+func NewOversizedLogFilter() (*AuditLogFilter, error) {
+	return NewAuditLogFilter(MaskConf{
```
charithe, line 74: Unless I am mistaken, doesn't this strip out pretty much all the information from the message? I think that if we just strip `inputs[*].resource.attr`, `inputs[*].principal.attr` and `ouputs[*].filterDebug`, the chances are quite high that the message size would come back down to reasonable levels.

internal/audit/hub/conf.go
Sambigeara: Might be worth still enforcing a hard limit in the code, at the very least, so we can [reuse the same memory for batching](https://github.com/cerbos/cerbos/blob/main/internal/audit/hub/hub.go#L236-L241). IIRC, we implemented this because it quite significantly reduced allocations and GC overhead 🤔

charithe: Actually, I forgot there's a [limit on the number of entries](https://github.com/cerbos/cloud-api/blob/706862567a9d26847c736aa7e4e24e2e768a9661/protos/cerbos/cloud/logs/v1/logs.proto#L41-L44) so we'd still need to ensure that the outgoing request doesn't exceed that. Let's just treat that value as a constant and not a user-configurable value.

internal/audit/hub/hub.go
Sambigeara: Hmm, good point. Not the easiest to switch between cases either, given the key length doesn't change 🤔

We _could_ set a new (increased?) key len, and make assertions based on the key length, but we would be putting a lot of reliance in a potentially unsafe assertion. The alternative is storing metadata in a dedicated key in Badger. 

Storing required info in the key did seem pretty neat (without the need for an additional key), though 🙂

internal/audit/local/badgerdb.go
Sambigeara: I suppose because when I initially implemented it, human-readability seemed like a good idea 😆. The `encoding` lib makes more sense 👍

internal/audit/hub/hub.go
charithe: Hmm... I am probably missing something here. The new keys can't be the same size as the old keys, can they? The old ones were `prefix:timestamp:id` the new ones should be `prefix:timestamp:id:size` which should make them 4 bytes longer (assuming that size is a 32-bit int). 

Isn't the `GenKeyWithByteSize` function currently overwriting part of the audit ID with the message size?

Sambigeara: Aaah, I got horribly mixed up and treated the timestamp portion of the ULID as the whole key 🤦‍♂️. OK, I'll just add the extra bytes to the existing 20-byte key and we should be all good.

internal/audit/hub/hub.go
```diff
@@ -336,18 +337,69 @@ func (l *Log) streamPrefix(ctx context.Context, kind logsv1.IngestBatch_EntryKin
...
+			var size int
+			if k := item.Key(); len(k) == local.KeyByteSizeStart { //nolint:nestif
+				entries, err := l.getIngestBatchEntries([][]byte{item.Key()}, kind)
+				if err != nil {
+					return err
+				}
+				if len(entries) > 0 {
+					entry := entries[0]
+					size = entry.SizeVT()
+					if size > l.maxBatchSizeBytes {
```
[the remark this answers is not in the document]
Sambigeara: They should still be processed? If they're legacy keys and _not_ oversized, they fall through to line 402 (and beyond) for normal processing 🤔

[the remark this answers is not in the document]
charithe: Oh of course! 🤦🏽‍♂️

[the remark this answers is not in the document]
Sambigeara: Oh sorry I missed your followup message, was mind mapping myself 😆

PR #2569, opened 2025-05-07
feat: Cerbosctl commands to interact with Hub store

charithe:


cmd/cerbosctl/hub/store/add_files.go
```diff
@@ -0,0 +1,204 @@
...
+	- 5: Command didn't change the remote store because it's already at the desired state
+	- 6: The version condition supplied using --version-must-eq wasn't satisfied
+
+# Upload foo.yaml and all files in the bar directory
+
+cerbosctl hub store add-files foo.yaml bar
+
+# Upload bar.yaml, renaming it to foo/bar.yaml in the store
+
+cerbosctl hub store add-files --mesage="Adding foo/bar.yaml" foo/bar.yaml=bar.yaml
```
Sambigeara, line 33:
```suggestion
cerbosctl hub store add-files --message="Adding foo/bar.yaml" foo/bar.yaml=bar.yaml
```

cmd/cerbosctl/hub/store/add_files.go
```diff
@@ -0,0 +1,204 @@
...
+`
+
+type AddFilesCmd struct {
+	filesToAdd    map[string]string
+	Output        `embed:""`
+	Message       string   `help:"Commit message for this change" default:"Uploaded using cerbosctl"`
+	Paths         []string `arg:"" help:"List of files or the directories to add to the store. To rename how the file appears in the store, use store_path=actual_path as the input format." required:""`
```
Sambigeara, line 40:
```suggestion
	Paths         []string `arg:"" help:"List of files or directories to add to the store. To rename how the file appears in the store, use store_path=actual_path as the input format." required:""`
```

cmd/cerbosctl/hub/store/add_files.go
```diff
@@ -0,0 +1,204 @@
...
+
+	"github.com/alecthomas/kong"
+
+	"github.com/cerbos/cerbos-sdk-go/cerbos/hub"
+	"github.com/cerbos/cerbos/internal/util"
+	storev1 "github.com/cerbos/cloud-api/genpb/cerbos/cloud/store/v1"
+)
+
+const addFilesHelp = `
+The following exit codes have a special meaning.
```
alexolivier, line 23: Is a non-zero exit code going to complicate running this in CI?

I don't know what is best practice here, but I wouldn't expect my release pipeline to fail if the policies haven't changed.

tcolgate: Seconded. 5 here would flag  by lots of things (normal CI usage, shells). 6 is OK as things should fail in that case. If you want behaviour like 5, I'd suggest adding an extra flag "--error-on-unchanged", possibly taking an optional code to use as the exit (could default to 5).

tcolgate: Another alternative would be to have an standardised yaml output that could include other fields to indicate state beyond exit. However that would probably be something you;d want to be consistant across all commands

cmd/cerbosctl/hub/store/replace_files.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/alecthomas/kong"
+
+	"github.com/cerbos/cerbos-sdk-go/cerbos/hub"
+)
+
+const replaceFilesHelp = `
+Replaces or deletes all files in the remote store so that it only contains the files provided.
+
+The following exit codes have a special meaning.
+	- 5: Command didn't change the remote store because it's already at the desired state
```
tcolgate, line 24: Same concerns here re: exit code.

tcolgate (dismissed):
The "things haven't changed" case seems like it will be very common if these tools are run from CI or scripts (which is going to be a major use case. Non-zero exit will interfete with `set -e`, shells that indicate error'd commands, and most default CI configurations. if we want to give the option of error'ing on unchanged, stick it behind a non-default flag.

internal/util/filesystem.go
```diff
@@ -29,6 +29,8 @@ const SchemasDirectory = "_schemas"
 // TestDataDirectory is the name of the special directory containing test fixtures. It is defined here to avoid an import loop.
 const TestDataDirectory = "testdata"
 
+const pathSeparator = string(filepath.Separator)
```
Sambigeara, line 32: This is also defined in `cmd/cerbosctl/hub/store/store.go`, _could_ make it public and share the same type 🤷‍♂️

Sambigeara (commented):
Reaaaally nice UX 👏

tcolgate (commented):
Have you checked the usage of path separator is valid on Windows? archive/zip always uses `/` (according to docs), os.Separator is OS specific, and then our internal API probably has a set assumption.

cmd/cerbosctl/hub/store/add_files.go
```diff
@@ -0,0 +1,204 @@
...
+
+			if len(contents) > maxFileSize {
+				yield(nil, fmt.Errorf("file too large: %s", actualPath))
+				return
+			}
+
+			batch = append(batch, &storev1.FileOp{
+				Op: &storev1.FileOp_AddOrUpdate{
+					AddOrUpdate: &storev1.File{
+						Path:     storePath,
```
tcolgate, line 184: storePath here is constructed with filepath.Join, which is OS dependent.

cmd/cerbosctl/hub/store/add_files.go
charithe: The no-op case is not straightforward to handle because the API doesn't return the store version when it's a no-op. The output from the command wouldn't be consistent in that case anyway. I'll figure out a fix for that in a different PR. Until then, you can do `cerbosctl ... || true` as a workaround.

cmd/cerbosctl/hub/store/add_files.go
```diff
@@ -0,0 +1,204 @@
...
+			storePath = filepath.Base(path)
+		}
+
+		var err error
+		filePath, err = filepath.Abs(filePath)
+		if err != nil {
+			return fmt.Errorf("failed to find absolute path of %s: %w", filePath, err)
+		}
+		storePath = filepath.Clean(storePath)
+		if storePath == "" || util.PathIsHidden(storePath) || !util.IsSupportedFileType(storePath) {
```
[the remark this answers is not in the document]
charithe: The use is explicitly providing the list of files here so it's better to fail rather than silently ignoring them.

cmd/cerbosctl/hub/store/add_files.go
charithe: I don't think it is? It's either `filepath.Base` of the path or the user-provided value. I've added a `filepath.ToSlash` above as a precaution. 

We don't officially support Windows anyway and there are probably a whole bunch of things that don't work properly on pure Windows.

charithe: I am going to merge this now because it's needed for a demo. If you spotted anything that needs fixing, please keep commenting and I'll fix them.

PR #2570, opened 2025-05-09
enhancement: Simplify plan with exists operation

dbuduev:
In case a collection is a known value, then `collection.exists(t, <bodyExpr>) can be simplified as a logical OR of the body expr evaluated for each value of `t` in the collection. The collection can be either a map or a list. Both single- and two-var operations are supported.

### DIFF
--- a/internal/engine/planner/struct_matcher.go
+++ b/internal/engine/planner/struct_matcher.go
@@ -353,7 +353,6 @@ func (l *lambdaMatcher) Process(ctx context.Context, e celast.Expr) (bool, celas
 		output := mkLogicalOr(opts)
 		internal.ZeroIDs(output)
 		output.RenumberIDs(internal.NewIDGen().Remap)
-		Print(output)
 		return true, output, nil
 	default:
 		return false, nil, nil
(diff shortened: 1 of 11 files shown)


internal/engine/planner/struct_matcher_test.go
```diff
@@ -0,0 +1,100 @@
...
+		},
+		{
+			expr: "4 == {\"a\": 3}[R.attr.Id]", // can't swap args
+		},
+		{
+			expr: "P.attr.anyMap[R.attr.Id] == R.attr.value",
+		},
+		{
+			expr: `{"a1": {"role": "OWNER"}}[R.id].role == "OWNER"`,
+			want: `R.id == "a1" && "OWNER" == {"role": "OWNER"}.role`,
```
charithe, line 35: If we can determine that `R.id` must be `a1` because there's only one key in the map, can't we apply the same logic to the inner map and realise that it's effectively comparing `"OWNER" == "OWNER"` and eliminate that altogether?

internal/engine/planner/struct_matcher.go
```diff
@@ -0,0 +1,447 @@
...
+	if err != nil {
+		return false, nil, err
+	}
+	l.iterVar = lambda.iterVar
+	l.iterVar2 = lambda.iterVar2
+
+	knownVars := make(map[string]any, len(l.partialEvaluator.knownVars)+nLambdaVars)
+	maps.Copy(knownVars, l.partialEvaluator.knownVars)
+
+	const maxItems = 25
```
charithe, line 341: Sorry, I know we talked about 25 being a "reasonable" number. But, now that I think about it, most query languages have an efficient operator for comparing against a list of items so it's probably better to make the cut-off a bit smaller to about 10 items. If the lambda expression is particularly complicated, a 25 of those OR'ed together is probably less efficient to evaluate compared to the equivalent array expression in the target query language.

internal/engine/planner/struct_matcher_test.go
dbuduev: This code recognises specific patterns in partially evaluated expressions and transforms them. If transformation happens, the result is going through the partial evaluator again, so I guess this is where this subexpression will be eliminated.

charithe (approved):

PR #2602, opened 2025-06-13
fix: Honour compile cache duration in rule table

### LINKED ISSUE #2601: Policies are not updated in case a policy is updated via another instance
Let's say we have pdp instance A and B, and both of them use the same database instance as its store.

If we update a policy via the Admin API of the instance B, the policy gets updated in the underlying database store. But the instance A, even if the cache is expires, will not be aware of this.

To remedy the situation;
- Go back to the cerbos pdp `v0.40.0`.
- Call reload store RPC of every cerbos PDP instance when policies are updated.

charithe:
Fixes #2601

### DIFF
--- a/internal/compile/manager.go
+++ b/internal/compile/manager.go
@@ -197,6 +197,10 @@ func (c *Manager) evict(modID namer.ModuleID) {
 	c.cache.Remove(modID)
 }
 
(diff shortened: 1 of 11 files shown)


oguzhand95 (approved):
LGTM

haines (approved):

PR #2609, opened 2025-06-16
docs: Generate llm.txt and llm-full.txt

alexolivier:
Installs `@cerbos/antora-llm-generator` extension in the antora container and adds the configuration to the playbook to run only for the latest version and non-legacy docs.

Adds an HTTP header in Netlify, but can't test this locally


charithe (commented):
This extension runs for all versions of the docs and duplicates the content over and over again for each version. Let's restrict it to the release version only.

Rather than building a new container, can't we just mount the extension as a volume? It looks like it's possible to specify the extension as an NPM package as well. See https://github.com/spring-io/antora-extensions.

charithe (approved):

PR #2617, opened 2025-06-27
feat: Decouple rule table

Sambigeara:
This is the first part of an extended effort to provide a dedicated package for rule table management and consumption (primarily to avoid a single ginormous PR).

The focus of this first change is to:
- Separate the rule table logic into it's own dedicated package
- To introduce an object that accepts precompiled rule tables and provides structured APIs for `Check` and `Plan` requests.

The following items will be tackled in followup PRs:
- Elegant handling of policy store update events (perhaps via a rule table patching mechanism) to avoid full lazy refreshes.
- Further simplification of the `Check` and `Plan` methods, by:
  - Removing the dependency on the PolicyLoader--the ruletable `Manager` should only take static, precompiled rule tables as input and should not be responsible for refreshing it's own state)
  - Integrating the schema manager logic into the `Manager`. Schema state can be baked into the rule table proto also
  - Etc
- Compilation to WASM
  - Necessary optimisations/trimming of superfluous logic/dependencies
  - Separation of PDP<->ePDP functionality (traces, logging etc) via build time constraints
  - Etc


internal/compile/conf.go
```diff
@@ -4,49 +4,19 @@
...
-)
+const confKey = "compile"
 
 // Conf is optional configuration for caches.
 type Conf struct {
-	// CacheSize is the number of compiled policies to cache in memory.
-	CacheSize uint `yaml:"cacheSize" conf:",example=1024"`
-	// CacheDuration is the duration to cache an entry.
-	CacheDuration time.Duration `yaml:"cacheDuration" conf:",example=60s"`
+	// [DEPRECATED] CacheSize is the number of compiled policies to cache in memory.
```
charithe, line 16: It would be good if we can warn the users who still have these settings configured.
We have a deprecation log method at https://github.com/cerbos/cerbos/blob/98d31453522dadc19223b3a3aaeb8b03718bc71a/internal/util/app.go#L81. We'd probably need a second one to say that the configuration has been removed completely and is not going to be effective.

internal/compile/manager.go
```diff
@@ -35,32 +32,15 @@ type Manager struct {
 	sf          singleflight.Group
```
charithe: This was needed before to handle the rush of concurrent requests that'd hit the PDP if it was restarted. Without it, each concurrent request resulted in a new compile operation that wasted time and resources. 

I assume that it's no longer needed because the rule table is loaded all at once on startup? If so, this struct member can be removed.

internal/ruletable/ruletable.go
```diff
@@ -0,0 +1,1991 @@
...
+	for _, evt := range events {
+		switch evt.Kind {
+		case storage.EventReload, storage.EventAddOrUpdatePolicy, storage.EventDeleteOrDisablePolicy:
+			// as a temporary measure, to avoid recompiling the whole rule table on updates when no Check
+			// or Plan calls are made, we toggle a flag which is checked at query time and lazily refresh
+			// the table then. This is far from ideal as it can increase the latency of the first query, but
+			// this is a temporary measure (and the lesser of two evils).
+			// TODO(saml) remove this once rule table patching is implemented
+			mgr.log.Info("Scheduling ruletable reload")
+			mgr.isStale.Store(true)
```
charithe, line 947: I know it's temporary and soon to be fixed. I'd say that recompiling in the background as soon as a storage event is detected is better than waiting for the next request because the latency would be noticeable and might even trigger alerts (we have an existing customer that actually does that).

internal/compile/conf.go
```diff
@@ -20,3 +25,19 @@ type Conf struct {
 func (c *Conf) Key() string {
 	return confKey
 }
+
+func (c *Conf) SetDefaults() {}
+
+func (c *Conf) Validate() (outErr error) {
+	if c.CacheSize != 0 {
+		util.DeprecationWarning("compile.cacheSize")
+		outErr = multierr.Append(outErr, errors.New("compile.cacheSize section is no longer supported"))
```
charithe: I don't think we need to throw an error at this point. We can just ignore the setting after printing the warning message.

charithe (approved):

internal/ruletable/ruletable.go
Sambigeara: Might need a better way of handling this. I implemented this suggestion and now we're hitting store related issues (e.g. SQLite lock contention during the `add_or_update` tests).

Not sure how hard I should be fighting on this though given it'll be refactored out before the branch is merged 🤔

charithe: Yep, I should have been clearer. I meant that we should have a better solution before it's merged to `main`. Doesn't have to be done in this PR.

charithe (approved):

PR #2624, opened 2025-07-03
feat: Include policy source in audit logs

haines:
This PR adds static information about the store config to the audit logs. Dynamic information (e.g. bundle ID, Git commit hash) to come later.


api/public/cerbos/audit/v1/audit.proto
```diff
@@ -68,3 +70,42 @@ message Peer {
...
+      DRIVER_UNSPECIFIED = 0;
+      DRIVER_MYSQL = 1;
+      DRIVER_POSTGRES = 2;
+      DRIVER_SQLITE3 = 3;
+    }
+
+    Driver driver = 1;
+  }
+
+  message Disk {}
```
charithe: Should this contain the path of the policy directory because it could change?

api/public/cerbos/audit/v1/audit.proto
```diff
@@ -68,3 +70,42 @@ message Peer {
...
+
+  message Git {
+    string repository_url = 1;
+    string branch = 2;
+    string sub_dir = 3;
+  }
+
+  message Hub {
+    oneof source {
```
charithe, line 106: This could be a local bundle as well and, even though it's rare, might be worth including the path to the local bundle used.

charithe (dismissed):

PR #2634, opened 2025-07-09
enhancement: Static rule table evaluator

Sambigeara:
* Introduces a new `evaluator.Evaluator` interface, with `Check` and `Plan` methods. The engine and (wrapped) static rule table both implement this interface (thus we can test both in the engine test suite).
* Creates a new static schema manager by storing the schema files as raw bytes in the proto type, and compiling into a static manager at start time.
* RuleTableManager now handles store events and updates the rule table accordingly.

### DIFF
--- a/internal/compile/manager.go
+++ b/internal/compile/manager.go
@@ -72,13 +72,47 @@ func (c *Manager) processUpdateQueue(ctx context.Context) {
 		case evt := <-c.updateQueue:
 			c.log.Debugw("Processing storage event", "event", evt)
 			switch evt.Kind {
-			case storage.EventReload, storage.EventAddOrUpdatePolicy, storage.EventDeleteOrDisablePolicy:
+			case storage.EventReload, storage.EventAddOrUpdateSchema, storage.EventDeleteSchema:
+				c.NotifySubscribers(evt)
+			case storage.EventAddOrUpdatePolicy, storage.EventDeleteOrDisablePolicy:
+				if err := c.addEventDependents(&evt); err != nil {
+					c.log.Warnw("Error while retrieving dependendents event", "event", evt, "error", err)
+				}
 				c.NotifySubscribers(evt)
 			default:
 				c.log.Debugw("Ignoring storage event", "event", evt)
 			}
 		}
 	}
+}
+
+func (c *Manager) addEventDependents(evt *storage.Event) error {
+	deps, err := c.getDependents(evt.PolicyID)
+	if err != nil {
+		return err
+	}
+
+	if len(deps) > 0 {
+		evt.Dependents = deps
+	}
+
+	return nil
+}
+
+func (c *Manager) getDependents(modID namer.ModuleID) ([]namer.ModuleID, error) {
+	ctx, cancelFunc := context.WithTimeout(context.Background(), storeFetchTimeout)
+	defer cancelFunc()
+
+	dependents, err := c.store.GetDependents(ctx, modID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find dependents: %w", err)
+	}
+
+	if len(dependents) > 0 {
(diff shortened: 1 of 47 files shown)


internal/compile/manager.go
```diff
@@ -81,9 +83,38 @@ func (c *Manager) processUpdateQueue(ctx context.Context) {
...
+		return dependents[modID], nil
+	}
+
+	return nil, nil
+}
+
 func (c *Manager) compile(unit *policy.CompilationUnit) (*runtimev1.RunnablePolicySet, error) {
 	return metrics.RecordDuration2(metrics.CompileDuration(), func() (*runtimev1.RunnablePolicySet, error) {
-		return Compile(unit, c.schemaMgr)
+		return Compile(unit, nil)
```
Sambigeara, line 117: I'm fairly certain this is fine. The playground uses the `Compile` call with schema validation, but the rule table needs to be able to load policies without failing due to missing schema. In the event of missing schemas, the "missing schema" error is returned in the outputs of `Check` calls.

internal/schema/schema.go
```diff
@@ -73,93 +79,104 @@ func (NopManager) ValidatePlanResourcesInput(_ context.Context, _ *policyv1.Sche
...
-	mgr := &manager{
-		conf:     conf,
-		log:      zap.L().Named("schema"),
-		cache:    cache.New[string, *cacheEntry]("schema", conf.CacheSize),
-		resolver: defaultResolver(loader),
+	sm := &StaticManager{
+		conf: conf,
+		log:  zap.L().Named("schema"),
 	}
+	sm.loader = sm
```
Sambigeara, line 104: I'm not super happy with this, it feels like I'm trying to fudge in OOP-like behaviour. That said, I'm not sure how else I should do it 🤷‍♂️

internal/ruletable/manager.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/cerbos/cerbos/internal/engine/policyloader"
+	"github.com/cerbos/cerbos/internal/engine/tracer"
+	"github.com/cerbos/cerbos/internal/evaluator"
+	"github.com/cerbos/cerbos/internal/namer"
+	"github.com/cerbos/cerbos/internal/schema"
+	"github.com/cerbos/cerbos/internal/storage"
+	"go.uber.org/zap"
+)
+
+const loaderFetchTimeout = 2 * time.Second
```
charithe: Probably worth making this configurable because there will be slow stores that'll take longer than 2 seconds.

internal/ruletable/manager.go
```diff
@@ -0,0 +1,252 @@
...
+			return fmt.Errorf("failed to load dependent policies: %w", err)
+		}
+
+		// we leave ruletable state static until we're sure all dependents are valid, and then update
+		for _, modID := range evt.Dependents {
+			mgr.deletePolicy(modID)
+		}
+
+		for _, rps := range toReload {
+			if err = mgr.addPolicy(rps); err != nil {
```
charithe, line 168: `addPolicy` calls `indexAndPurgeRules` which effectively goes through all the rules in the rule table and recreates them (temporarily almost doubling the memory ). If the changed policy has a lot of dependents, repeating that in quick succession could make the memory usage balloon up quite quickly and the process could potentially get OOM killed. Is there a better way to handle this that's less memory-intensive?

Sambigeara: Yeah, fair point. I think it would be a fairly simple refactor to return the proto row from `AddPolicy` and index it directly, bypassing the proto `Rules` entirely.

Sambigeara:
I think I can pre-compute the only remaining dynamic map on the rule table (`parentRoleAncestorsCache`) and remove the mutex also. Just having a look at that now.

Edit: done.

PR #2637, opened 2025-07-16
chore: Streamline rule table for embedded builds

Sambigeara:
Refactor rule table and dependencies to reduce binary size and complexity for embedded environments by creating lighter-weight/stub implementations and isolating them via build tags.

This reduced the WASM binary from 56mb to 48mb. Running Binaryen trims off a bit more and lands us at 41mb. Brotli on that: 6.5mb.

We do introduce some maintenance overhead with aspects of this--a good example being the `internal/obervability/logging` implementation, which requires us to manually expose `zap` functions each time we require a new one (and also to implement the stub function in the `!js && !wasm` build). If there are more elegant ways of approaching this, I'm all ears!


charithe (approved):
🎇

PR #2641, opened 2025-07-18
enhancement: Stickier PDP IDs

charithe:
This can help reduce the noise in audit logs because the ID is "sticky".
Otherwise, the same PDP could get different IDs on each restart even
though it's still running on the same hardware. These IDs are still
perfectly opaque and don't reveal anything other than the fact that the
PDP is running on the same anonymous machine every time it's restarted.

### DIFF
--- a/internal/util/app.go
+++ b/internal/util/app.go
@@ -4,7 +4,9 @@
 package util
 
 import (
+	"crypto/hmac"
 	"crypto/md5" //nolint:gosec
+	"crypto/sha256"
 	"fmt"
 	"os"
 	"runtime/debug"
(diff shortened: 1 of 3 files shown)


internal/util/app.go
```diff
@@ -62,9 +69,21 @@ func AppShortVersion() string {
 }
 
 var getPdpID = sync.OnceValue(func() string {
-	//nolint:gosec
-	nodeID := md5.Sum(uuid.NodeID())
-	return fmt.Sprintf("%X-%d", nodeID, os.Getpid())
+	machineID, err := machineid.ID()
```
oguzhand95, line 72: Wouldn't it be easier to use [machineid.ProtectedID(appID string)](https://github.com/keygen-sh/machineid/blob/0bd4ea32a95c643d8bdf26ed4a2f29bf70c322a0/id.go#L39-L45)?

charithe: We need to know the internals of `ProtectedID` to truncate the ID correctly so it's safer to inline it in case it changes in a future version.

PR #2649, opened 2025-07-24
fix: Handle rule-less policies for multi-tenant fallthrough

Sambigeara:
The rule table previously inferred policy existence by searching for matching rules in the given scope (for the resource/principal). However, there is a case where defining policies without rules is useful, such as in a multi-tenant application where some tenants want to override or narrow rules, and others want to assume base rules. In the latter case, with lenientScopeSearch set to false, we need a policy to "exist" in the target scope in order to allow the fallthrough.

Adding a stub row for resource or principal policies with no rules bypasses the lenientScopeSearch+scoping guard at evaluation time, thus allowing parent policies to be evaluated.


internal/engine/ruletable/rule_table.go
```diff
@@ -519,6 +519,24 @@ func (rt *RuleTable) addPrincipalPolicy(rpps *runtimev1.RunnablePrincipalPolicyS
 	}
 	rt.scopeScopePermissions[p.Scope] = scopePermissions
 
+	if len(p.ResourceRules) == 0 {
```
haines, line 522: Can we get some test cases to cover these scenarios? Would be good to prevent future regressions.

PR #2661, opened 2025-08-05
feat: Add AWS Lambda support

dbuduev:
This PR adds: 
- integration with the AWL Lambda Extension API, allowing the Cerbos PDP to run as an AWS Lambda Extension,
- AWS Lambda handler.

### DIFF
--- a/internal/compile/manager.go
+++ b/internal/compile/manager.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"time"
 
 	"go.uber.org/zap"
 
@@ -20,17 +19,9 @@ import (
 	"github.com/cerbos/cerbos/internal/storage"
 )
 
-const (
-	negativeCacheEntryTTL = 10 * time.Second
-	storeFetchTimeout     = 2 * time.Second
-	updateQueueSize       = 32
-)
-
 type Manager struct {
-	store       storage.SourceStore
-	log         *zap.SugaredLogger
-	updateQueue chan storage.Event
-	*storage.SubscriptionManager
+	store storage.SourceStore
+	log   *zap.SugaredLogger
 }
 
 func NewManager(ctx context.Context, store storage.SourceStore) (*Manager, error) {
@@ -39,46 +30,11 @@ func NewManager(ctx context.Context, store storage.SourceStore) (*Manager, error
 	}
 
 	c := &Manager{
-		log:                 zap.S().Named("compiler"),
-		store:               store,
-		updateQueue:         make(chan storage.Event, updateQueueSize),
-		SubscriptionManager: storage.NewSubscriptionManager(ctx),
+		log:   zap.S().Named("compiler"),
+		store: store,
 	}
-
-	go c.processUpdateQueue(ctx)
-	store.Subscribe(c)
 
 	return c, nil
-}
-
-func (c *Manager) SubscriberID() string {
-	return "compile.Manager"
-}
-
-func (c *Manager) OnStorageEvent(events ...storage.Event) {
-	for _, evt := range events {
-		c.log.Debugw("Received storage event", "event", evt)
-		c.updateQueue <- evt
-	}
-}
-
-func (c *Manager) processUpdateQueue(ctx context.Context) {
-	for {
-		select {
-		case <-ctx.Done():
-			return
-		case evt := <-c.updateQueue:
-			c.log.Debugw("Processing storage event", "event", evt)
-			switch evt.Kind {
-			case storage.EventReload:
-				c.NotifySubscribers(evt)
-			case storage.EventAddOrUpdatePolicy, storage.EventDeleteOrDisablePolicy:
-				c.NotifySubscribers(evt)
-			default:
-				c.log.Debugw("Ignoring storage event", "event", evt)
-			}
-		}
-	}
 }
 
 func (c *Manager) compile(unit *policy.CompilationUnit) (*runtimev1.RunnablePolicySet, error) {
--- a/internal/compile/manager_test.go
+++ b/internal/compile/manager_test.go
@@ -17,7 +17,6 @@ import (
 	"github.com/stretchr/testify/require"
 
 	policyv1 "github.com/cerbos/cerbos/api/genpb/cerbos/policy/v1"
-	schemav1 "github.com/cerbos/cerbos/api/genpb/cerbos/schema/v1"
 	"github.com/cerbos/cerbos/internal/compile"
 	"github.com/cerbos/cerbos/internal/namer"
 	"github.com/cerbos/cerbos/internal/policy"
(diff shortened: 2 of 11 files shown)


internal/server/lambda-ext.go
```diff
@@ -0,0 +1,120 @@
...
+	url := fmt.Sprintf("http://%s%s", runtimeAPI, registrationEndpoint)
+
+	registerReq := RegisterRequest{Events: []string{"SHUTDOWN"}}
+
+	reqBody, err := json.Marshal(registerReq)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal register request: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "POST", url, bytes.NewBuffer(reqBody))
```
charithe:
```suggestion
	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewBuffer(reqBody))
```

internal/server/lambda-ext.go
```diff
@@ -0,0 +1,120 @@
...
+
+	l.extensionID = resp.Header.Get(extensionIDHeader)
+
+	return &l, nil
+}
+
+func (l *lambdaExt) checkShutdown(ctx context.Context) (bool, error) {
+	log := zap.L().Named("lambda-ext-impl")
+
+	req, err := http.NewRequestWithContext(ctx, "GET", l.nextEventURL, nil)
```
charithe:
```suggestion
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, l.nextEventURL, nil)
```

internal/server/server.go
```diff
@@ -264,6 +264,9 @@ func (s *Server) Start(ctx context.Context, param Param) error {
 		return err
 	}
 
+	if runtimeAPI := os.Getenv("AWS_LAMBDA_RUNTIME_API"); runtimeAPI != "" {
```
charithe: Let's not assume that the user always wants to run Cerbos as an extension and let them opt into it via a command-line flag/environment variable.

internal/server/lambda-ext.go
```diff
@@ -0,0 +1,120 @@
...
+		var urlErr *url.Error
+		if errors.As(err, &urlErr) && urlErr.Timeout() {
+			log.Debug("Checking next event timed-out")
+			return false, nil
+		}
+		return false, fmt.Errorf("failed to get next event: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
```
charithe: If the status is a 5XX, should we just let the extension stop immediately or should we retry a couple of times before giving up?

dbuduev: We can retry or ignore 5XX errors completely, as this type of failure doesn't affect the PDP directly.

dbuduev: From the docs:
> 500 – Container error. Non-recoverable state. Extension should exit promptly.

dbuduev:
@charithe, please have another look.

deploy/aws/lambda-ext/main.go
```diff
@@ -0,0 +1,105 @@
...
+	"go.uber.org/zap"
+)
+
+func main() {
+	ctx, stopFunc := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stopFunc()
+
+	configPath := os.Getenv("CERBOS_CONFIG")
+	if configPath == "" {
+		configPath = "/conf.yml"
```
charithe: The default config file name we use is `.cerbos.yaml`

```suggestion
		configPath = "/.cerbos.yaml"
```

deploy/aws/lambda-ext/main.go
```diff
@@ -0,0 +1,105 @@
...
+		return cerbos.Serve(ctx,
+			cerbos.WithConfigFile(configPath),
+			cerbos.WithLogLevel(cerbos.LogLevel(logLevel)),
+		)
+	})
+	if err := waitForReady(ctx); err != nil {
+		log.Error("Readiness check failed", zap.Error(err))
+		exit2()
+	}
+	p.Go(func(ctx context.Context) error {
```
charithe: This should go inside a `p.Go` as well because if the cerbos process started above fails for some reason, it won't be aware of it and will hang forever until the entire process is killed by the runtime or by the user.

```suggestion
	p.Go(func(ctx context.Context) error {
	    if err := waitForReady(ctx); err != nil {
		    log.Error("Readiness check failed", zap.Error(err))
		    return fmt.Errorf("readiness check failed: %w", err)
	    }
```

deploy/aws/lambda-ext/main.go
```diff
@@ -0,0 +1,105 @@
...
+				stopFunc()
+				break
+			}
+		}
+		return nil
+	})
+	if err := p.Wait(); err != nil {
+		log.Error("Stopping server due to error", zap.Error(err))
+		exit2()
+	}
```
charithe:
```suggestion
	if err := p.Wait(); err != nil {
	    if !errors.Is(err, context.Canceled) {
		    log.Error("Stopping server due to error", zap.Error(err))
		    exit2()
		}
	}
```

deploy/aws/lambda-ext/main.go
charithe, line 1: How is this file going to be used? It won't be available from the container entrypoint.

internal/server/awslambda/function.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get server configuration: %w", err)
+	}
+
+	reqLimits := svc.RequestLimits{
+		MaxActionsPerResource:  serverConf.RequestLimits.MaxActionsPerResource,
+		MaxResourcesPerRequest: serverConf.RequestLimits.MaxResourcesPerRequest,
+	}
+
+	cerbosSvc := svc.NewCerbosService(eng, auxData, reqLimits)
```
charithe: I am concerned about the duplication of code from the `server` package here. It's very likely to drift apart because anyone working on the code would have to know that this code path exists too. It's not a trivial amount of code either so I'd like to find a way to share it instead of duplicating it.

Is there a big overhead if we just start a normal Cerbos server in a goroutine using `pkg/cerbos` and forward the lambda requests to it over a unix domain socket? I feel like the biggest contributor to the startup time is the initialization steps such as loading the store that we can't avoid in any case. I'd be surprised if the final step of launching the server has any significant impact.

dbuduev: Ouch, I meant to share the code between the lambda and the server. I refactored the code and extracted the common part. 

> Is there a big overhead if we just start a normal Cerbos server in a goroutine using pkg/cerbos and forward the lambda requests to it over a unix domain socket?  I feel like the biggest contributor to the startup time is the initialization steps such as loading the store...

I agree that the startup time will be about the same if we start a server. However, I would like to avoid the extra serialisation and deserialization for each request on both sides of the IPC. CPU seems to be a bottleneck in lambdas.

charithe: Because the Lambda function wrapper is already converting the request and response to protobuf, I won't expect there to be any extra serialization if the gRPC service is called over IPC. However, the refactoring to share the code is fine as well 👍🏽

charithe (approved):

PR #2663, opened 2025-08-08
enhancement: More detailed schema errors

oguzhand95:


internal/test/testdata/compile/missing_schemas.yaml
```diff
@@ -4,15 +4,15 @@ wantErrors:
...
+      Failed to load principal schema "cerbos:///non_existent_principal.json": schema file "non_existent_principal.json" does not exist in the store
     position:
       line: 10
       column: 7
       path: "$.resourcePolicy.schemas.principalSchema.ref"
   - file: resource_policies/leave_request_20210210.yaml
     error: invalid schema
     description: |-
-      Failed to load resource schema "cerbos:///non_existent_leave_request.json": schema "cerbos:///non_existent_leave_request.json" does not exist in the store
+      Failed to load resource schema "cerbos:///non_existent_leave_request.json": schema file "non_existent_leave_request.json" does not exist in the store
```
haines: I like chomping off the `cerbos://` scheme, but I'd be tempted to prepend `_schemas/` so that it has the full path that was expected (otherwise I can see people trying to put schemas in the root of the policies directory).

internal/schema/schema.go
```diff
@@ -74,21 +77,49 @@ func NewEphemeral(resolver Resolver) Manager {
 	return mgr
 }
 
-func (m *manager) LoadSchema(ctx context.Context, url string) (*jsonschema.Schema, error) {
-	entry, ok := m.cache.Get(url)
+func (m *manager) LoadSchema(ctx context.Context, _url string) (*jsonschema.Schema, error) {
```
charithe: Why does `url` have an underscore as a prefix?

internal/schema/schema.go
```diff
@@ -74,21 +77,49 @@ func NewEphemeral(resolver Resolver) Manager {
...
 
-	m.cache.Set(url, e)
-	return e.schema, e.err
+	parsedURL, err := url.Parse(_url)
+	if err != nil {
+		e.err = fmt.Errorf("failed to parse URL: %w", err)
+		m.cache.Set(_url, e)
+		return e.schema, e.err
+	}
+	if parsedURL.Scheme != URLScheme {
```
haines: I feel like we're doing the work to parse the URL and translate into a filesystem path in two places now. I reckon we need to do more of these checks in the resolver instead; we can probably return more useful errors from there to avoid having to do so much work here to figure out what went wrong.

charithe: We do support other schemes such as `http` and any unsupported schemes would be caught by https://github.com/cerbos/cerbos/blob/6b80dba8f39e7c8f8d7248902a4ae79eed460523/internal/schema/schema.go#L145

internal/schema/schema.go
```diff
@@ -74,21 +77,49 @@ func NewEphemeral(resolver Resolver) Manager {
...
+		e.err = fmt.Errorf("invalid URL scheme %q", parsedURL.Scheme)
+		m.cache.Set(_url, e)
+		return e.schema, e.err
+	}
+	parsedURLPath := strings.TrimPrefix(parsedURL.Path, "/")
+
+	pathErrPath := strings.TrimPrefix(pathErr.Path, util.SchemasDirectory+string(os.PathSeparator))
```
charithe: You can only do this if the scheme is `cerbos://`. 
What is the actual error produced by `jsonschema` when a child schema is missing? Hopefully there's a way to extract the missing file information from the error itself.

oguzhand95: The `err` when the file itself is missing:
image

The `err` when the file itself is present but the file referenced is missing:
image

charithe: I think we can throw a custom error from https://github.com/cerbos/cerbos/blob/6b80dba8f39e7c8f8d7248902a4ae79eed460523/internal/schema/schema.go#L142 that tells us exactly which schema reference was not found. To be thorough, we could also replace https://github.com/santhosh-tekuri/jsonschema/blob/16bce71af51f6a4a775f11e649a347a8803940d3/httploader/httploader.go#L23 so that it throws the same custom error if the request sends a 404.

charithe: Just saw @haines review above. I think he's referring to a similar idea.

internal/schema/schema.go
```diff
@@ -130,22 +147,94 @@ type cacheEntry struct {
...
+
+func (e notFoundErr) Error() string {
+	return fmt.Sprintf("schema file %q does not exist in the store", e.Path)
+}
+
 func DefaultResolver(loader Loader) Resolver {
 	return func(ctx context.Context, path string) (io.ReadCloser, error) {
 		u, err := url.Parse(path)
 		if err != nil {
 			return nil, err
```
charithe, line 172:
```suggestion
			return nil, fmt.Errorf("failed to parse URL %q: %w", path, err)
```

internal/schema/schema.go
```diff
@@ -74,20 +79,32 @@ func NewEphemeral(resolver Resolver) Manager {
...
+		parsedSchemaURL, err := url.Parse(schemaURL)
+		switch {
+		case err != nil:
+			e.err = fmt.Errorf("failed to parse URL: %w", err)
+		case notFoundErr.Scheme != URLScheme:
+			e.err = fmt.Errorf("schema file %q does not exist", schemaURL)
+		case strings.TrimPrefix(parsedSchemaURL.Path, "/") != strings.TrimPrefix(notFoundErr.Path, util.SchemasDirectory+string(os.PathSeparator)):
+			e.err = fmt.Errorf("schema file %q referenced by the schema does not exist in the store", notFoundErr.Path)
+		default:
+			e.err = fmt.Errorf("schema file %q does not exist in the store", notFoundErr.Path)
```
charithe: If you change `notFoundErr` to return the full URL it was trying to load and the resolved system path (e.g. `_schemas/foo/bar.json`) you can simplify this as follows.

```suggestion
        if notFoundErr.url != schemaURL {
            e.err = fmt.Errorf("schema %s referenced by %s doesn't exist", notFoundErr.fullPath, schemaURL)
        } else {
            e.err = fmt.Errorf("schema %s doesn't exist", notFoundErr.fullPath)
        }	
```

internal/schema/schema.go
```diff
@@ -130,22 +147,94 @@ type cacheEntry struct {
 	err    error
 }
 
+type notFoundErr struct {
+	Scheme string
+	Path   string
```
charithe:
```suggestion
	url      string // URL that failed to load
	fullPath string // Resolved file system path. For http/https it would be the full URL again.
```

internal/schema/schema.go
```diff
@@ -130,22 +147,94 @@ type cacheEntry struct {
...
+}
+
+func (e notFoundErr) Error() string {
+	return fmt.Sprintf("schema file %q does not exist in the store", e.Path)
```
charithe:
```suggestion
	return fmt.Sprintf("schema %q does not exist", e.fullPath)
```

internal/schema/schema.go
```diff
@@ -87,21 +87,16 @@ func (m *manager) LoadSchema(ctx context.Context, schemaURL string) (*jsonschema
...
+	e.schema, e.err = m.loadSchemaFromStore(ctx, schemaURL)
+	if e.err != nil && errors.As(e.err, &notFoundErr) {
 		switch {
-		case err != nil:
-			e.err = fmt.Errorf("failed to parse URL: %w", err)
-		case notFoundErr.Scheme != URLScheme:
-			e.err = fmt.Errorf("schema file %q does not exist", schemaURL)
-		case strings.TrimPrefix(parsedSchemaURL.Path, "/") != strings.TrimPrefix(notFoundErr.Path, util.SchemasDirectory+string(os.PathSeparator)):
-			e.err = fmt.Errorf("schema file %q referenced by the schema does not exist in the store", notFoundErr.Path)
+		case notFoundErr.scheme != URLScheme:
```
charithe: Isn't it possible for a schema loaded through http to have a reference to a child schema that doesn't exist as well? (Essentially, the same error we are trying to fix here with `cerbos://` schemas.)

PR #2672, opened 2025-08-21
fix: Fix e2e blob test by guaranteeing policy existence

Sambigeara:
Previously, there was a race condition between policies being added to the bucket (and reflected in the rule table via event propagation) and the tests running themselves. This fixes that by ensuring the minio bucket houses the policies prior to the cerbos pod starting, so that the rule table will be initialised with the correct state.

### DIFF
--- a/e2e/blob/blob_test.go
+++ b/e2e/blob/blob_test.go
@@ -7,9 +7,7 @@ package blob_test
 
 import (
 	"fmt"
-	"path/filepath"
 	"testing"
-	"time"
 
 	"github.com/cerbos/cerbos/internal/storage/blob"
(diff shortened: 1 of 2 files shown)


charithe:
Just wondering why it takes longer for the rule table to update? Will that become an issue in production where policies are added/updated to blob store at random?

Sambigeara:
I wonder if it's related to indeterministic ordering of policy additions leading to invalid indexes, which ultimately forces full rebuilds on the rule table 🤔 . In the worst case, this could result in numerous attempts to pull all policies for reindexing. It could have a compounding effect. I'll have a think.

PR #2677, opened 2025-08-26
fix: Batch dependents and deflake CI suites

Sambigeara:
* Storage (blob/db): emit one deduped dependents-only event per batch;
  drop per-policy dependents (there was some duplicated effort here).
* Compile manager: stop enriching all Add/Update and Delete events with
  dependents as we now rely on the storage layers to handle this (there
  was also some duplicated effort here).
* Rule table manager: removed the index-unhealthy gate flag. It's no
  longer necessary since the compile-and-swop mechanism only mutates
  rule table state on successful compilation. Processing dependents
  atomically also relies on all to be available and valid in order for
  mutations to occur (so links are kept consistent). Without this gate,
  we bypass a large number of wasteful full reloads.
* E2E tests: I still needed to add a sleep in between the AdminSuite and
  CheckSuite in order to allow events generated in the mutable stores
  to fully propogate to the rule table before Check requests were fired
  at it. I did a hell of a lot of experimentation, and think that this
  is in part due to differing, more constrained runtime environments
  nudging us into "race" territory (I verified by checking out an old,
  previously passing commit, and saw that fail in Github's actions too).
* Other misc changes.

### DIFF
--- a/e2e/blob/helmfile.yaml.gotmpl
+++ b/e2e/blob/helmfile.yaml.gotmpl
@@ -1,12 +1,14 @@
 repositories:
   - name: bitnami
     url: https://charts.bitnami.com/bitnami
+
 helmDefaults:
   cleanupOnFail: true
   wait: true
   recreatePods: true
   force: true
   createNamespace: true
+
 releases:
   - name: minio
     namespace: '{{ requiredEnv "E2E_NS" }}'
@@ -37,7 +39,8 @@ releases:
           rootPassword: passw0rd
       - defaultBuckets: "cerbos"
       - persistence:
-          enabled: false
+          enabled": false
+
   - name: cerbos
     namespace: '{{ requiredEnv "E2E_NS" }}'
     needs: ["minio"]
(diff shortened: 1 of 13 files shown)


e2e/blob/helmfile.yaml.gotmpl
```diff
@@ -37,7 +39,8 @@ releases:
           rootPassword: passw0rd
       - defaultBuckets: "cerbos"
       - persistence:
-          enabled: false
+          enabled": false
```
charithe: This is a typo that has existed for a while but we should probably fix it.
```suggestion
          enabled: false
```

internal/storage/blob/store.go
```diff
@@ -263,6 +265,18 @@ func (s *Store) updateIndex(ctx context.Context) (err error) {
...
+	}
+
+	depsMap, err := s.idx.GetDependents(modIDs...)
+	if err != nil {
+		return fmt.Errorf("failed to get dependents: %w", err)
+	}
+	for i, e := range evts {
+		if deps, ok := depsMap[e.PolicyID]; ok && len(deps) > 0 {
+			evts[i].Dependents = make([]namer.ModuleID, len(deps))
+			copy(evts[i].Dependents, deps)
```
charithe: Is the copy necessary? Why not just assign `deps` to `evts[i].Dependents`?

e2e/blob/helmfile.yaml.gotmpl
Sambigeara: Lol yes, missed that in the `revert`, thanks

charithe (approved):

PR #2690, opened 2025-09-08
enhancement: Retrieve change details from git repository if available when using replace-files cmd

oguzhand95:


cmd/cerbosctl/hub/store/replace_files.go
```diff
@@ -89,7 +104,50 @@ func (rfc *ReplaceFilesCmd) Run(k *kong.Kong, cmd *Cmd) error {
...
+		changeDetails.WithUploaderDetails(gitChangeDetails.uploader)
+	default:
+		changeDetails.WithUploaderDetails(&storev1.ChangeDetails_Uploader{
+			Name: defaultName,
+		})
+	}
+
+	switch {
+	case rfc.Origin != nil:
+		internal := &storev1.ChangeDetails_Internal{}
```
charithe: This could be a Git change details object as well. Although it doesn't make much sense to do so manually, if the user wants to do it, we shouldn't prevent them.

cmd/cerbosctl/hub/store/replace_files.go
```diff
@@ -7,14 +7,24 @@ import (
...
+	storev1 "github.com/cerbos/cloud-api/genpb/cerbos/cloud/store/v1"
+	"github.com/go-git/go-git/v5"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+const (
+	defaultMessage = "Uploaded using cerbosctl"
+	defaultName    = "cerbos-sdk-go"
+	defaultSource  = "sdk"
```
charithe:
```suggestion
	defaultName    = "cerbosctl"
	defaultSource  = "cerbosctl"
```

charithe (commented):
Please add a line to `cmd/cerbosctl/hub/store/testdata/testscripts/store_ops.txtar` to exercise the new parameters. I know it's not possible to check whether they actually took effect but we'd at least verify that the code is parsing things correctly.

oguzhand95:
Closed and moved to cerbos/cerbos#2695.

PR #2695, opened 2025-09-10
enhancement: Add upload-git cmd and change details from git to replace-files cmd

oguzhand95:


cmd/cerbosctl/hub/store/upload_git.go
```diff
@@ -0,0 +1,298 @@
...
+		return ugc.toCommandError(k.Stderr, err)
+	}
+
+	version := ugc.VersionMustEq
+	for batch, err := range ugc.batch() {
+		if err != nil {
+			return ugc.toCommandError(k.Stderr, err)
+		}
+
+		changeDetails, err := changeDetailsFromHash(repository, ugc.diffToApply.hash)
```
oguzhand95: We are most likely squashing multiple commits. Does generating the change details from the last commit make any sense? (I have a feeling it could confuse people.) @charithe

cmd/cerbosctl/hub/store/upload_git.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/go-git/go-git/v5/plumbing/object"
+
+	"github.com/cerbos/cerbos/internal/util"
+)
+
+const uploadGitHelp = `
+The following exit codes have a special meaning.
+	- 6: The version condition supplied using --version-must-eq wasn't satisfied
+
+# Add, delete or update files from commit 55a4248 to HEAD
```
charithe:
```suggestion
# Apply the file changes recorded in the git repo between commit 55a4248 and HEAD
```

cmd/cerbosctl/hub/store/upload_git.go
```diff
@@ -0,0 +1,298 @@
...
+
+cerbosctl hub store upload-git 55a4248
+cerbosctl hub store upload-git 55a4248 --path path/to/git/repository
+
+# Add, delete or update files from commit 55a4248 to e746228
```
charithe:
```suggestion
# Apply the file changes recorded in the git repo between commit 55a4248 to e746228
```

cmd/cerbosctl/hub/store/upload_git.go
```diff
@@ -0,0 +1,298 @@
...
+cerbosctl hub store upload-git 55a4248 e746228
+cerbosctl hub store upload-git 55a4248 e746228 --path path/to/git/repository
+`
+
+type UploadGitCmd struct {
+	diffToApply   *diff
+	Output        `embed:""`
+	From          string `arg:"" help:"Git revision to start from when generating the diff"`
+	To            string `arg:"" help:"Git revision to end when generating the diff (The resolved reference must be the ancestor of the from argument)" default:"HEAD"`
+	Path          string `help:"Path to the git repository" default:"."`
```
charithe, line 43: We should also add an argument to specify a particular directory such as `policies` so that only the changes that happened under that path are uploaded to the store.

cmd/cerbosctl/hub/store/upload_git.go
charithe: We should allow the user to provide an optional message and other details like the rest of the commands. If none is provided, I think it's OK to take the details from the last commit similar to how squashing commits in git works.

cmd/cerbosctl/hub/store/upload_git.go
charithe: We do a similar thing in the `git` store: https://github.com/cerbos/cerbos/blob/1b8e12dcaebd745a50a7831205ade23615b79bbf/internal/storage/git/store.go#L422

oguzhand95: I think if `subdir` flag is provided,
it shouldn't change the directory structure in the store, should it?

Let's say the repository is in `$HOME/repo` and the flag `--subdir policies` is passed to the command.
If the `$HOME/repo` directory looks like this:
```
$HOME/repo/
    policies/
        leave_request.yaml
```

I think the directory structure in the store should look like this:
```
policies/
    leave_request.yaml
```

charithe: Since they are giving us the policy directory, I think it's much more natural to use that as the root. There's no point in mimicking the layout of the entire Git repo because Hub stores are specifically meant for storing policies.

cmd/cerbosctl/hub/store/upload_git.go
```diff
@@ -0,0 +1,314 @@
...
+
+		gitChangeDetails, err := changeDetailsFromHash(repository, ugc.diffToApply.hash)
+		if err != nil {
+			return ugc.toCommandError(k.Stderr, fmt.Errorf("failed to get change details for %q: %w", ugc.diffToApply.hash.String(), err))
+		}
+
+		changeDetails, message, err := ugc.ChangeDetails.ChangeDetails(gitChangeDetails)
+		if err != nil {
+			return ugc.toCommandError(k.Stderr, fmt.Errorf("failed to get change details: %w", err))
+		}
```
charithe: These can be moved outside the loop because they'd generate the same value every time.

cmd/cerbosctl/hub/store/upload_git.go
```diff
@@ -0,0 +1,314 @@
...
+			name = objectChange.To.Name
+			operation = OpAddOrUpdate
+		case from != nil:
+			name = objectChange.From.Name
+			operation = OpDelete
+		}
+
+		if ugc.skipped(name) {
+			continue
+		}
```
charithe: Move this after line 256 because it's safer to do the subdirectory check after the path has been cleaned.

cmd/cerbosctl/hub/store/upload_git.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	changes := make([]*change, 0, len(objectChanges))
+	for _, objectChange := range objectChanges {
+		from, to, err := objectChange.Files()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get file from change: %w", err)
+		}
+
+		var nameOfToBeDeleted string
```
charithe:
```suggestion
		var oldNameOfRenamedFile string
```

cmd/cerbosctl/hub/store/upload_git.go
```diff
@@ -0,0 +1,328 @@
...
+		case from != nil:
+			name = objectChange.From.Name
+			operation = OpDelete
+		}
+
+		name = filepath.Clean(name)
+		path := filepath.Join(pathToRepo, name)
+		if path == "" || util.PathIsHidden(path) || !util.IsSupportedFileType(path) {
+			return nil, fmt.Errorf("invalid file %q: must not be hidden and must be a YAML or JSON file", path)
+		}
```
charithe: Let's just ignore unsupported files because we are looking at git commits and there's no easy fix that the user can do if there are unsupported files in there.

PR #2707, opened 2025-09-16
chore(ci): Publish lambda handler to AWS SAR

dbuduev:
The lambda is published, but not shared publicly.


.github/workflows/release.yaml
```diff
@@ -201,3 +201,69 @@ jobs:
...
+            --output-template-file packaged-template.yml \
+            --region ${{ vars.AWS_REGION }}
+      - name: Publish to AWS Serverless Application Repository
+        run: |-
+          VERSION="${GITHUB_REF#refs/tags/}"
+          cd deploy/awslambda/function
+          sam publish \
+            --template packaged-template.yml \
+            --semantic-version "${VERSION#v}" \
+            --region ${{ vars.AWS_REGION }}
```
charithe: Some of this is already in the `justfile`. If the deployment logic is written as a   `just` recipe, we could just invoke that from the workflow with the correct arguments. That would make it easier to test and fix the script without having to run the workflow. It also gives us the ability to do a manual release if something fails for whatever reason. Otherwise we'd have to re-tag and create an entirely new release of the PDP just to trigger this workflow again.

cmd/awslambda/function/main.go
```diff
@@ -5,16 +5,23 @@ package main
...
 	"github.com/cerbos/cerbos/internal/observability/logging"
 	"github.com/cerbos/cerbos/internal/server/awslambda"
+	"github.com/cerbos/cerbos/internal/util"
 )
 
 func main() {
+	if len(os.Args) > 1 && (os.Args[1] == "--version" || os.Args[1] == "-v") {
+		fmt.Printf("cerbos %s\n", util.Version) //nolint:forbidigo
+		return
+	}
```
charithe: Isn't this only used for the just recipe below? Otherwise, even though it's only a single argument, it's better to use Kong for consistency. It'd also handle unknown arguments properly.

In practice, we should always be doing the release from CI so I'd make the `VERSION` parameter mandatory for the just recipe and fail if the version is not provided.

deploy/awslambda/function/.justfile
```diff
@@ -1,12 +1,96 @@
...
+        --description "Cerbos configuration and policies layer" \
+        --zip-file fileb://{{ LAYER_NAME }}.zip \
+        --compatible-architectures arm64 x86_64
+    
+    # Cleanup
+    rm -rf test-layer-content {{ LAYER_NAME }}.zip
+    
+    echo "Layer {{ LAYER_NAME }} published successfully"
+
+publish-to-sar VERSION='' ARCH=arch() BUCKET="$CERBOS_SAM_PACKAGING_BUCKET": (function-package ARCH)
```
charithe: If the argument has a `$` prefix, it will be made available as an environment variable to the recipe.

```suggestion
publish-to-sar ARCH=arch() $VERSION $CERBOS_SAM_PACKAGING_BUCKET: (function-package ARCH)
```

deploy/awslambda/function/.justfile
```diff
@@ -1,12 +1,96 @@
...
+    set -euo pipefail
+
+    if [[ -z "{{ BUCKET }}" ]]; then
+        echo "Error: BUCKET arg is required"
+        exit 1
+    fi
+    if [[ -z "{{ VERSION }}" ]]; then
+        echo "Error: version is required"
+        exit 1
+    fi
```
charithe: If you leave out the default values as I suggest above, `just` will take care of this.

deploy/awslambda/function/.justfile
```diff
@@ -1,12 +1,96 @@
...
+    
+    arch=$(sed -e 's/aarch64/arm64/' -e 's/amd64/x86_64/' <<< "{{ ARCH }}")
+    app_name="cerbos-lambda-function-${arch/_/-}" # _ isn't valid character in the name
+    
+    # Create template with architecture-specific name, replace arch (noop if arch is arm64)
+    sed -e "s/\"cerbos-lambda-function\"/\"${app_name}\"/" -e "s/- arm64/- ${arch}/" sam.yml > .sam.tmp.yml
+
+    version=${version#v}
```
charithe: This should be derived from the `VERSION` argument, shouldn't it?

```suggestion
    VERSION=${VERSION#v}
```

charithe (approved):

PR #2726, opened 2025-10-03
feat: Add API adapters

haines:
This PR introduces the capability for the PDP to expose other authorization APIs by defining adapters in config.


haines:
I'll look at adding Envoy external authz adapters in a subsequent PR (hence why we have config under `server.adapters.http` rather than just `server.adapters`).

PR #2729, opened 2025-10-08
chore: Support Hub config in Lambda

dbuduev:


cmd/awslambda/extension/main.go
```diff
@@ -41,19 +44,18 @@ func main() {
...
-	var overrides map[string]any
+	overrides := make(map[string]any)
 	if configPath == "" {
-		overrides = make(map[string]any)
-		if err := awslambda.MkConfStorageOverrides("/var/task/policies", overrides); err != nil {
-			log.Error("failed to create storage config overrides", zap.Error(err))
-		}
-		if err := awslambda.MkConfServerOverrides(overrides); err != nil {
-			log.Error("failed to create server config overrides", zap.Error(err))
+		if err := getConfOverrides(overrides); err != nil {
```
charithe: I missed this in the previous review. In the PDP, configuration values provided by `--set` (and associated environment variables) override what's defined in the config file. So, I think the overrides should be computed even when the `configPath` is not empty.

cmd/cerbos/server/server.go
```diff
@@ -66,22 +82,13 @@ func (c *Cmd) Run() error {
...
-			"storage.driver=hub",
-			fmt.Sprintf("storage.hub.remote.deploymentID=%s", c.Hub.DeploymentID),
-		}
-	case c.Hub.PlaygroundID != "":
-		hubOverrides = []string{
-			"storage.driver=hub",
-			fmt.Sprintf("storage.hub.remote.playgroundID=%s", c.Hub.PlaygroundID),
-		}
-	case c.HubBundle != "":
+	if c.HubBundle != "" {
```
charithe: Why leave this out of `MkHubOverrides`? It's also a subtle breaking change because the order of evaluation has changed. It was allowing for the fact that some users might not have cleaned up their old environment variables and just tacked on the new ones on top. The precedence was given to the new v2 configuration in such cases. With this change, that breaks.

dbuduev: I assumed these are mutually exclusive, and Kong would complain if both are set.

HubBundle       string       `help:"[Legacy] Use Cerbos Hub to pull the policy bundle with the given label. Overrides the store defined in the configuration." optional:"" env:"CERBOS_HUB_BUNDLE,CERBOS_CLOUD_BUNDLE" group:"hubv1" **xor:"hub.deployment-id,hub.playground-id"**

cmd/awslambda/extension/main.go
dbuduev: > In the PDP, configuration values provided by --set (and associated environment variables) 

Yes, but Lambda does not support command-line arguments, so the --set option is not available.
Or do you mean that the associated environment variables can provide overrides without `--set`? Can you point me to how overriding is done in this case?

charithe: Sorry, not `set` but args related to Hub. They take precedence over what's defined in the config file.

cmd/cerbos/server/server.go
charithe: I forgot about that. To the other point, though: as long as we're supporting legacy mode the lambda one should do as well.

cmd/awslambda/extension/main.go
dbuduev: Done, except for setting the policy directory in the non-hub scenario, and also the config is missing. It sets the override that makes sense for Lambda.

PR #2743, opened 2025-10-22
chore: Added messaging about Hub to PDP logs

emreb:
Added info message about Cerbos Hub to PDP logs when an instance that is not connected to Cerbos Hub launches.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -449,7 +449,7 @@ func (s *Server) startHTTPServer(ctx context.Context, l net.Listener, grpcSrv *g
 		log.Infof("Starting HTTP server at %s", s.conf.HTTPListenAddr)
 
 		if suggestHub {
-			log.Info("Cerbos Hub offers features like enhanced policy management, " +
+			zap.L().Named("hub").Info("Cerbos Hub offers features like enhanced policy management, " +
 				"continuous deployment pipelines, and enterprise support. " +
 				"Learn more at https://go.cerbos.io/hub")
 		}


pkg/cerbos/serve.go
```diff
@@ -124,6 +126,14 @@ func Serve(ctx context.Context, options ...ServeOption) error {
 		return err
 	}
 
+	if storageConf, err := storage.GetConf(); err == nil && storageConf.Driver != hub.DriverName {
```
haines: We should probably check the audit log backend as well since you can use Hub for that without using it for storage.

emreb: updated.

emreb:
Now the message only shows up once Cerbos is up and running. 
image

internal/server/server.go
```diff
@@ -447,6 +447,12 @@ func (s *Server) startHTTPServer(ctx context.Context, l net.Listener, grpcSrv *g
 
 	s.pool.Go(func(_ context.Context) error {
 		log.Infof("Starting HTTP server at %s", s.conf.HTTPListenAddr)
+
+		if suggestHub {
+			log.Info("Cerbos Hub offers features like enhanced policy management, " +
```
haines: I don't think this should be logged by the HTTP logger although I guess it needs to be here in order to be logged after the server has started.

```suggestion
			zap.L().Named("hub").Info("Cerbos Hub offers features like enhanced policy management, " +
```

PR #2752, opened 2025-11-03
feat: AuthZEN v1.0 API

alexolivier:
Initial support of an AuthZEN-compatible API served on a separate port.

Notes:
- Only evaluation endpoints are implemented. Subject/Resource/Action search endpoints are not implemented and are intentionally omitted from metadata.
- Requests are translated to Cerbos CheckResources calls. Supplying `subject.properties.roles` is **required**; requests without at least one role are rejected with HTTP `400`.
- The AuthZEN subject type is exposed inside the Cerbos principal attributes under `$type` (e.g., `attr["$type"] = "user"`).
- AuthZEN `context`, if present, is embedded under `$context` inside the principal attributes for policy evaluation.
 - If the underlying Cerbos CheckResources result includes `outputs`, they are returned in the AuthZEN decision under `context.outputs` (each item includes `src` and `val`).
- `options.evaluations_semantic` is accepted, but only `"execute_all"` is currently supported; requests specifying any other value are rejected with `400`.
- TLS is optional. When `server.tls` is configured, the AuthZEN server advertises HTTPS endpoints; otherwise, endpoints are served over HTTP, which can be secured by an upstream load balancer or service mesh.

This PR was written with help from ChatGPT Codex


api/public/cerbos/authzen/v1/authzen.proto
```diff
@@ -0,0 +1,251 @@
...
+option go_package = "github.com/cerbos/cerbos/api/genpb/cerbos/authzen/v1;authzenv1";
+option java_package = "dev.cerbos.api.v1.authzen";
+
+// Subject represents the AuthZEN subject entity.
+message Subject {
+  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
+    json_schema: {description: "AuthZEN Subject"}
+  };
+  // Arbitrary key-value properties attached to the subject.
+  google.protobuf.Struct properties = 1 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Arbitrary key-value properties attached to the subject."}];
```
charithe, line 23: We use `map<string, google.protobuf.Value>` everywhere else. IIRC it's because the construction of `google.protobuf.Struct` objects is not straightforward in some languages and and there were some issues with JSON marshalling as well.

api/public/cerbos/authzen/v1/authzen.proto
```diff
@@ -0,0 +1,251 @@
...
+  ];
+  Resource resource = 3 [
+    (buf.validate.field).required = true,
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Resource to evaluate."}
+  ];
+  google.protobuf.Struct context = 4 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Optional evaluation context."}];
+}
+
+// EvaluationsRequest is the batch request form for multiple tuples.
+message AccessEvaluationsRequest {
```
charithe, line 118: This naming is bad because it's just a single letter difference between `AccessEvaluationRequest` and `AccessEvaluationsRequest`.

api/public/cerbos/authzen/v1/authzen.proto
```diff
@@ -0,0 +1,251 @@
...
+  Action action = 3 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Default action for all evaluations; tuples may override."}];
+  // Optional default resource; tuples may override.
+  Resource resource = 4 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Default resource for all evaluations; tuples may override."}];
+  // The list of evaluations to perform.
+  repeated Tuple evaluations = 5 [
+    (buf.validate.field).repeated = {min_items: 0},
+    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Evaluations to perform (order preserved)."}
+  ];
+  // Optional execution options applied to the batch request.
+  AccessEvaluationsOptions options = 6 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Execution options for the batch request."}];
```
charithe, line 136: We don't support this so it should be left out for now.

internal/server/authzen.go
charithe, line 1: Don't need this

internal/server/conf.go
```diff
@@ -148,6 +151,14 @@ type AdvancedConf struct {
 	GRPC AdvancedGRPCConf `yaml:"grpc"`
 }
 
+// AuthZENConf holds configuration for the OpenID AuthZEN translation layer.
+type AuthZENConf struct {
+	// Enabled defines whether the AuthZEN HTTP API is enabled.
+	Enabled bool `yaml:"enabled" conf:",example=false"`
+	// ListenAddr is the dedicated HTTP address for AuthZEN.
+	ListenAddr string `yaml:"listenAddr" conf:",example=\":3595\""`
```
charithe, line 159: We don't need a separate server or port for AuthZen.

PR #2769, opened 2025-11-06
chore: Ruletable generic backends

Sambigeara:
Refactors `RuleTable` indexing logic into a standalone index package to support pluggable storage. The internal data structure has moved from nested maps to a flattened inverted index, resolving rules via set intersections. This change introduces a thread-safe in-memory backend (replacing the previous non-concurrent implementation which was protected by the manager mutex) and an experimental Redis backend.

I tried for a few days to implement a more efficient batching API for the Redis backend, failed, and then decided it was probably premature optimisation for a backend that might not ever be used.

Edit: apologies, this seems like an unnecessarily complex change. However the real functional change for the PDP is that the index has been flattened and the API's changed to suit. This should be an operational no-op. I don't think much attention needs to be given to the Redis backend right now given that it's experimental, and will either disappear, or be iterated on in the future.

### DIFF
--- a/internal/engine/engine_test.go
+++ b/internal/engine/engine_test.go
@@ -63,14 +63,10 @@ func TestCheck(t *testing.T) {
 	client := redis.NewClient(&redis.Options{Addr: mr.Addr()})
 	t.Cleanup(func() { client.Close() })
 
-	redIdx := index.New(client, "test", 0, 0)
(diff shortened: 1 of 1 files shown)


charithe (dismissed):
Sorry, I changed the title to `chore` because it's not a user-visible change. I am aware that the system is broken and I'll look into fixing it soon.

oguzhand95 (approved):

PR #2773, opened 2025-11-10
enhancement: Support for rule table bundles

charithe:
Extends the `hub` storage driver to handle rule table bundles.

### DIFF
--- a/internal/server/common.go
+++ b/internal/server/common.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 
+	runtimev1 "github.com/cerbos/cerbos/api/genpb/cerbos/runtime/v1"
 	audithub "github.com/cerbos/cerbos/internal/audit/hub"
 
 	"github.com/cerbos/cerbos/internal/audit"
@@ -79,22 +80,27 @@ func InitializeCerbosCore(ctx context.Context) (*CoreComponents, error) {
 		return nil, ErrInvalidStore
 	}
 
-	rt := ruletable.NewProtoRuletable()
-
+	var ruleTable *runtimev1.RuleTable
 	//nolint:nestif
 	if ruleTableStore != nil {
-		rt, err = ruleTableStore.GetRuleTable()
+		rt, err := ruleTableStore.GetRuleTable()
 		if err != nil {
 			if !errors.Is(err, storagehub.ErrUnsupportedOperation) {
 				return nil, fmt.Errorf("failed to load rule table: %w", err)
 			}
 
-			if err := ruletable.LoadPolicies(ctx, rt, policyLoader); err != nil {
+			ruleTable = ruletable.NewProtoRuletable()
+			if err := ruletable.LoadPolicies(ctx, ruleTable, policyLoader); err != nil {
 				return nil, fmt.Errorf("failed to load policies: %w", err)
 			}
+		} else {
+			ruleTable = rt
 		}
-	} else if err := ruletable.LoadPolicies(ctx, rt, policyLoader); err != nil {
-		return nil, fmt.Errorf("failed to load policies: %w", err)
+	} else {
+		ruleTable = ruletable.NewProtoRuletable()
+		if err := ruletable.LoadPolicies(ctx, ruleTable, policyLoader); err != nil {
+			return nil, fmt.Errorf("failed to load policies: %w", err)
+		}
 	}
 
 	// create schema manager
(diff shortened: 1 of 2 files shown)


private/verify/verify_test.go
```diff
@@ -38,5 +39,5 @@ func TestBundle(t *testing.T) {
 
 	results, err := verify.Bundle(ctx, params, test.PathToDir(t, "store"))
 	require.NoError(t, err)
-	require.Equal(t, policyv1.TestResults_RESULT_PASSED, results.Summary.OverallResult)
+	require.Equal(t, policyv1.TestResults_RESULT_FAILED, results.Summary.OverallResult)
```
haines, line 42: Why aren't the tests for this bundle passing?

charithe: The policy has changed but the v1 bundle wasn't updated with those changes. When I switched to v2, it had the recent policies and the tests failed as they should.

internal/storage/hub/remote_source.go
```diff
@@ -353,25 +376,24 @@ func (s *RemoteSource) fetchBundleOffline() error {
...
 
-	return s.swapBundle(bdlPath, nil)
+	return s.swapBundle(bdlPath, nil, bundlev2.BundleType_BUNDLE_TYPE_LEGACY)
 }
 
 func (s *RemoteSource) removeBundle(healthy bool) {
+	var oldBundle Bundle
 	s.mu.Lock()
-	oldBundle := s.bundle
+	oldBundle = s.bundle
```
Sambigeara, line 385: Is this to explicitly highlight the type? 🤔

internal/storage/hub/remote_source.go
```diff
@@ -381,26 +403,35 @@ func (s *RemoteSource) swapBundle(bundlePath string, encryptionKey []byte) error
...
 		}
-	default:
-		return fmt.Errorf("unsupported bundle version: %d", s.bundleVersion)
 	}
 
+	var oldBundle Bundle
 	s.mu.Lock()
-	oldBundle := s.bundle
-	s.bundle = bundle
+	oldBundle = s.bundle
```
Sambigeara, line 433: (same confusion as above)

internal/test/testdata/bundle/v2_ruletable/manifest.json
```diff
@@ -0,0 +1,58 @@
+{
```
Sambigeara, line 1: I'm not sure if this is correct (or relevant in `local_source_test.go`). The rule table manifest contains only the bundle ID (unlike legacy bundle types with this schema).

Edit: see other comment

internal/storage/hub/local_source_test.go
```diff
@@ -50,13 +53,27 @@ func TestLocalSource(t *testing.T) {
...
+	})
+
+	tctx = mkTestCtx(t, bundle.Version2, bundlev2.BundleType_BUNDLE_TYPE_RULE_TABLE)
+	lsrt := mkLocalSource(t, tctx)
+	t.Run("ruleTable", func(t *testing.T) {
+		mb, err := os.ReadFile(filepath.Join(tctx.rootDir, "manifest.json"))
+		require.NoError(t, err)
+
+		manifest := &bundlev2.Manifest{}
+		require.NoError(t, protojson.Unmarshal(mb, manifest))
```
Sambigeara, line 68: I'm not sure this is testing the right thing? The [rule table manifest schema](https://github.com/cerbos/cerbos/blob/main/api/private/cerbos/runtime/v1/runtime.proto#L80-L82) only contains `bundle_id` so the policy index won't be available (it's not really relevant anymore, given the rule table itself acts as the source of truth).

For compatibility with this test, you could construct the index by iterating over the rule table, perhaps? I think the newly introduced `ListPolicyIDs` method in `ruletable_bundle.go` might be useful here?

charithe: I am trying to avoid creating a special case for rule tables that would require duplicating the tests. The other bundle tests use the contents of the manifest for checking the results returned by the bundle. I generated a synthetic manifest in the same format so that the existing tests could be reused for rule tables. 

Because we are trying to test `List*` methods of the bundle, it doesn't seem right to use them generate the expectations in the first place. I certainly won't have caught the issue with schema references in the rule table if I had done that. (The rule table schema list contains full URLs whereas the other bundles just contain the file names.)

internal/storage/hub/remote_source.go
charithe: No. It's a leftover from one of my experiments where I tried a different approach. I wasn't very good at reverting it back to the original. However, I think this is slightly more readable compared to the previous  version.

PR #2775, opened 2025-11-11
chore(ci): Fix no such host error in e2e tests

oguzhand95:
- Using fully-qualified DNS resolved the issue.
- Upgraded the telepresence to `v2.25.1`.

### DIFF
--- a/e2e/run.sh
+++ b/e2e/run.sh
@@ -32,7 +32,7 @@ stop_kind() {
 run_tests() {
     (
         cd "$SCRIPT_DIR"
-        telepresence helm upgrade
+        telepresence helm install
         telepresence connect --no-report -- go test -v -failfast -p=1 --tags="tests e2e" "$@"
     )
 }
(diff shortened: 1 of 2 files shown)


e2e/run.sh
```diff
@@ -32,7 +32,7 @@ stop_kind() {
 run_tests() {
     (
         cd "$SCRIPT_DIR"
-        telepresence helm upgrade
+        telepresence helm install
```
oguzhand95, line 35: `telepresence helm upgrade` results in an error unless I am missing another `install` command somewhere else. So, I think this should be `telepresence helm install`. 🤔 @charithe

charithe: If it's an issue with Telepresence, the other tests would have the same problem as well. What I'd do is:

- Undo the `svc.cluster.local` changes in this PR. They shouldn't be necessary.
- Create a branch directly in the cerbos repo with your change to Telepresence. 
- Launch the E2E workflow from that branch and see if it succeeds.

PR #2801, opened 2025-11-24
feat(api): Implement AuthZen PDP metadata endpoint

dbuduev:
https://openid.net/specs/authorization-api-1_0-04.html#name-policy-decision-point-metad


charithe (commented):
Please rebase and add a changelog entry with `just changelog-entry feature "Implement AuthZen specification"` (you can add more details by editing the generated file). With #2804 we are now explicitly creating changelog entries rather than relying on individual commits. The whole AuthZen implementation would become a single changelog entry.

charithe (approved):

PR #2804, opened 2025-11-24
chore: Changelog management

charithe:
We used to generate the changelog from git commits based on conventional commit information but lately that has caused a lot of confusion and despair because some changes span multiple commits. This change introduces a new system where the user-facing changelog entries are maintained separately.

Add a new changelog entry with:

```console
just changelog-entry TYPE DESCRIPTION
```

This creates a file in `.changelog` which can be edited to add as much detail as necessary in Asciidoc format. During a release, the contents of all files added after the previous release will be compiled into the changelog for the new release. This is inspired by https://github.com/hashicorp/go-changelog. I was going to use it initially but it required a few tweaks to match our workflow and I eventually ended up creating a mini clone tailored to this repo.

Adding a changelog entry with each PR is highly recommended but we won't enforce it mechanically because it doesn't make sense to write one for each and every PR (which is what led us here in the first place).

### DIFF
--- a/.justfile
+++ b/.justfile
@@ -20,12 +20,12 @@ align PKG='./...': _betteralign
 
 build: generate lint tests package
 
-changelog-generate NEW_VERSION PREV_VERSION=`git describe --abbrev=0  --match='v*'`:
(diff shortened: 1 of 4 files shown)


charithe:
I was thinking of getting rid of conventional commits because people tend to get the classification wrong far more often than I expected.

Issue #2813, opened 2025-11-25
Principal Policies Stop Working in v0.44.0


eswolinsky3241:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

We are trying to upgrade Cerbos to the newest version. We are currently on version 0.38.1. We found that our principal policies stopped working in version 0.44.0 with no error messages or warnings during compilation or runtime. Policies that worked correctly in v0.43.0 silently fail to authorize actions in v0.44.0.

In v0.44.0, principal policies are ignored during authorization checks. Actions that should be allowed by principal policies are denied, with no indication in logs or compilation output that anything is wrong.

### Expected Behavior

Principal policies should evaluate and allow/deny actions as defined, consistent with behavior in v0.43.0 and earlier versions.

### 1. Create test policy

Here's a minimal principal policy that I wrote to recreate the issue:

```yaml
apiVersion: api.cerbos.dev/v1
principalPolicy:
  principal: testuser
  version: v1
  rules:
  - resource: document:legacy
    actions:
    - action: view
      effect: EFFECT_ALLOW
    - action: edit
      effect: EFFECT_ALLOW
```

### 2. Start Cerbos server with v0.43.0
```
docker run -d --rm --name cerbos-test-043 -p 3592:3592 \
  -v $(PWD)/test-case:/policies \
  ghcr.io/cerbos/cerbos:0.43.0 server --set=storage.disk.directory=/policies
```
### 3. Test principal policy.
```
curl -X POST http://localhost:3592/api/check/resources \
  -H "Content-Type: application/json" \
  -d '{
    "requestId": "test-principal-043",
    "principal": {
      "id": "testuser",
      "policyVersion": "v1",
      "roles": ["user"]
    },
    "resources": [{
      "resource": {
        "kind": "document:legacy",
        "id": "doc123",
        "policyVersion": "v1"
      },
      "actions": ["view", "edit"]
    }]
  }' | jq .
```

This results in both actions being allowed, as expected:
```
{
  "requestId": "test-principal-043",
  "results": [
    {
      "resource": {
        "id": "doc123",
        "kind": "document:legacy",
        "policyVersion": "v1"
      },
      "actions": {
        "edit": "EFFECT_ALLOW",
        "view": "EFFECT_ALLOW"
      }
    }
  ],
  "cerbosCallId": "01KAXYSPSTFWCCQCAQATBCB6BY"
}
```

However, when trying the same command with Cerbos `v0.44.0`:
[...]

PR #2814, opened 2025-11-25
fix: Sanitise resource kinds in principal policy rules

Sambigeara:
Fixes https://github.com/cerbos/cerbos/issues/2813

I've manually tested against the reported case. Our test suite doesn't catch this in it's existing form, I'll add a specific test retrospectively once I've figured out how to do it cleanly.

**Edit**. The plot thickens:

```go
package main

import (
	"github.com/cerbos/cerbos/internal/namer"
)

func main() {
	println(namer.SanitizedResource("foo:bar"))
	println(namer.SanitizedResource("arn:aws:sns:us-east-1:123456789012:topic-a"))
}
```

Outputs:
```sh
$ go run test.go
foo_bar
arn:aws:sns:us-east-1:123456789012:topic-a
```

**Edit 2**:
This is expected behaviour. The `oldNamePattern` matches a very particular case--anything that strictly has a letter following a colon (not a number, as per the second example).

Regardless, the `sanitizeResource` fix here is still correct. We need to sanitise at index time to ensure we query against the same sanitised key at evaluation time.

### DIFF
--- a/cmd/cerbosctl/put/put_test.go
+++ b/cmd/cerbosctl/put/put_test.go
@@ -97,6 +97,7 @@ func testPutCmd(clientCtx *cmdclient.Context, globals *flagset.Globals) func(*te
 					"principal.donald_duck.vdefault/acme.hr",
 					"principal.donald_duck.vdefault/acme.hr.uk",
 					"principal.donald_duck.vdefault/acme.sales",
+					"principal.frodo.vdefault",
 					"principal.scrooge_mcduck.vdefault",
 					"principal.terry_tibbs.vdefault",
 					"resource.account.vdefault",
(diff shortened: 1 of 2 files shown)


Sambigeara:
I'll pick this up in the morning.

.changelog/20251125_1846.txt
```diff
@@ -0,0 +1,2 @@
+type:fix
+sanitise resource kinds in principal policy rules
```
charithe:
```suggestion
Sanitise resource kinds in principal policy rules.
Fixes a regression where resource kinds with special characters in their name were not correctly matched during policy evaluation.
```

PR #2817, opened 2025-11-27
enhancement: Simplify plan with all operation

dbuduev:
In case a collection is a known value, then `collection.all(t,<bodyExpr>)` can be simplified as a logical AND of the body expr evaluated for each value of `t` in the collection. The collection can be either a map or a list. Both single- and two-var operations are supported.


dbuduev:
> I was thinking we should have a limit on the size of the collection or it would result in an absolutely massive plan. But, I suppose the user has already provided the massive list in the request 🤷🏽‍♂️

We have a limit of 10 items.

PR #2831, opened 2025-12-05
fix: Correctly intersect role policy DENYs for multi-role principals

Sambigeara:
Previously, DENYs from role policies were aggregated using a union (OR) strategy, causing a single role's DENY to block access globally even if other roles did not deny it (via `(not (or (not (FOO) (not (BAR)))))`). This change implements intersection (AND) logic for role policies, ensuring a request is only denied if *all* of the principal's active roles explicitly deny the action via their respective role policies (`(not (and (not (FOO) (not (BAR)))))`).


charithe (approved):

charithe:
@Sambigeara Could you please add a changelog entry for this. I forgot about it during the review.

PR #2836, opened 2025-12-08
Introduce InspectPolicies API for rule tables

oguzhand95:
Moving `storage.RuleTableStore` to `ruletable` package doesn't exactly feel right to me. I am open to any other suggestion(s).

### DIFF
--- a/private/plan/plan.go
+++ b/private/plan/plan.go
@@ -12,7 +12,6 @@ import (
 	internalengine "github.com/cerbos/cerbos/internal/engine"
 	"github.com/cerbos/cerbos/internal/evaluator"
 	"github.com/cerbos/cerbos/internal/ruletable"
-	"github.com/cerbos/cerbos/internal/ruletable/index"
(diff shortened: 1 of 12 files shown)


private/plan/plan.go
```diff
@@ -23,17 +25,22 @@ func Resources(ctx context.Context, conf *evaluator.Conf, idx compile.Index, inp
...
-	rt := ruletable.NewProtoRuletable()
+	protoRT := ruletable.NewProtoRuletable()
 
-	if err := ruletable.LoadPolicies(ctx, rt, compiler); err != nil {
-		return nil, err
+	if err := ruletable.LoadPolicies(ctx, protoRT, compiler); err != nil {
+		return nil, fmt.Errorf("failed to load policies: %w", err)
+	}
+
+	ruleTable, err := ruletable.NewRuleTable(index.NewMem(), protoRT)
```
charithe: This pattern keeps repeating in almost every place that requires a rule table. Let's make it into a function and avoid the repetition.

PR #2838, opened 2025-12-09
Options to control instrumentation of in-process service

charithe:
Adds ability to switch off Otel instrumentation and inject a different
Zap core to redirect logging.

### DIFF
--- a/internal/observability/logging/logging.go
+++ b/internal/observability/logging/logging.go
@@ -48,18 +48,18 @@ var (
 )
 
 // InitLogging initializes the global logger.
-func InitLogging(ctx context.Context, level string) {
+func InitLogging(ctx context.Context, level string, zapCore zapcore.Core) {
 	initOnce.Do(func() {
 		if envLevel := os.Getenv("CERBOS_LOG_LEVEL"); envLevel != "" {
-			doInitLogging(ctx, envLevel)
+			doInitLogging(ctx, envLevel, zapCore)
 			return
 		}
 
-		doInitLogging(ctx, level)
+		doInitLogging(ctx, level, zapCore)
 	})
 }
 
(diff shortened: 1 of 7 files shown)


internal/observability/logging/logging.go
```diff
@@ -48,18 +48,18 @@ var (
 )
 
 // InitLogging initializes the global logger.
-func InitLogging(ctx context.Context, level string) {
+func InitLogging(ctx context.Context, level string, zapCore zapcore.Core) {
```
oguzhand95, line 51: It could work better since both `level` and `zapCore` parameters are optional:
```golang
type logOptions struct {
	zapCore zapcore.Core
	level   string
}

type LogOpt func(*logOptions)

func WithLevel(level string) LogOpt {
	return func(lo *logOptions) {
		lo.level = level
	}
}

func WithZapCore(zapCore zapcore.Core) LogOpt {
	return func(lo *logOptions) {
		lo.zapCore = zapCore
	}
}

// InitLogging initializes the global logger.
func InitLogging(ctx context.Context, options ...LogOpt) {
	initOnce.Do(func() {
		opts := logOptions{level: "INFO"}
		for _, option := range options {
			option(&opts)
		}

		if envLevel := os.Getenv("CERBOS_LOG_LEVEL"); envLevel != "" {
			doInitLogging(ctx, envLevel)
			return
		}

		doInitLogging(ctx, opts.level, opts.zapCore)
	})
}
```

charithe: It's indeed much nicer to read. The reason I didn't do it is because it's an internal API and there are currently only two optional things. The functional options pattern requires a lot more code than doing a couple of emptiness checks -- which feels a bit too much in this instance. If we add more options in the future, we should definitely refactor the code in this way.

PR #2842, opened 2025-12-12
fix: Isolate role policy restrictions

Sambigeara:
This regression was introduced in my previous fix
(https://github.com/cerbos/cerbos/pull/2831). After this "fix", role policy restrictions were aggregated and applied globally to the query plan. This caused a regression where restrictions on a specific role incorrectly constrained permissions granted by other unrestricted roles held by the user.

This moves the evaluation logic into the per-role processing loop. Role policy restrictions are now applied as a local filter on the specific role's ALLOW node before merging into the final result, ensuring role independence.

A positive is that I undid a bunch of complexity that was required for the incorrectly applied "global aggregation", so the planner output is simplified in certain circumstances (see the `combined_role_policies` test assertion).

If you diff against the commit prior to my regression (`git diff 61ab8a6 internal/ruletable/ruletable.go`), you'll see a satisfyingly reduced change set.

### DIFF
--- a/.changelog/20251212_1132.txt
+++ b/.changelog/20251212_1132.txt
@@ -1,4 +1,3 @@
 type:fix
 Isolate role policy restrictions in query planner
-This regression was introduced in my previous fix (https://github.com/cerbos/cerbos/pull/2831). After this "fix", role policy restrictions were aggregated and applied globally to the query plan. This caused a regression where restrictions on a specific role incorrectly constrained permissions granted by other unrestricted roles held by the user.
-This moves the evaluation logic into the per-role processing loop. Role policy restrictions are now applied as a local filter on the specific role's ALLOW node before merging into the final result, ensuring role independence.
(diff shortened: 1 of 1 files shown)


Sambigeara:
Just adding the changelog

.changelog/20251212_1132.txt
```diff
@@ -0,0 +1,4 @@
+type:fix
+Isolate role policy restrictions in query planner
+This regression was introduced in my previous fix (https://github.com/cerbos/cerbos/pull/2831). After this "fix", role policy restrictions were aggregated and applied globally to the query plan. This caused a regression where restrictions on a specific role incorrectly constrained permissions granted by other unrestricted roles held by the user.
```
charithe: I think this should be simplified a bit. It exists purely for the benefit of end users who are reading the release notes. So, something along the lines of "fixes a regression introduced as a result of fixing how role policy effects are intersected for principals with multiple roles" should suffice.

PR #2843, opened 2025-12-12
Add default scope configuration to be used when scope is not specified in the request

oguzhand95:
# Description
Cerbos policies have a `scope` field to support a way to model hierarchical relationships that regularly occur in many situations.
By default, when a request does not explicitly specify the scope, the Cerbos engine attempts to find a matching policy that has its scope set to `""`.
You can change this fallback value by setting the `defaultScope`.
For example, you may set the default scope to `acme` with `defaultScope: acme` which will take effect when the request does not explicitly specify `scope` field.

# Usage
```yaml
engine:
  defaultScope: "acme"
```

### DIFF
--- a/.changelog/20251212_1652.txt
+++ b/.changelog/20251212_1652.txt
@@ -1,3 +1,3 @@
 type:feature
-Add default scope configuration to be used when scope is not specified in the request
-When configuring Cerbos, it is now possible to set a default scope with the configuration parameter "engine.defaultScope". When CheckResources or PlanResources requests don't explicitly set scope, the default scope will be used as fallback.
+Ability to configure a default scope for requests
+It is now possible to configure the PDP with a default scope value. If `engine.DefaultScope` is set, any `CheckResources` or `PlanResources` requests without an explicit scope will default to the configured value.


.changelog/20251212_1652.txt
```diff
@@ -0,0 +1,3 @@
+type:feature
+Add default scope configuration to be used when scope is not specified in the request
+When configuring Cerbos, it is now possible to set a default scope with the configuration parameter "engine.defaultScope". When CheckResources or PlanResources requests don't explicitly set scope, the default scope will be used as fallback.
```
charithe:
```suggestion
Ability to configure a default scope for requests
It is now possible to configure the PDP with a default scope value. If `engine.DefaultScope` is set, any `CheckResources` or `PlanResources` requests without an explicit scope will default to the configured value.
```

charithe (dismissed):

haines:
If the default scope is set, does that now mean it is impossible to explicitly set the scope to `""` in the request?

charithe:
That's a very good point. We should figure out a way to allow that.

haines:
I _think_ you can add `optional` to a field and although it changes the generated code it doesn't affect the messages on the wire so it is backwards compatible? Then we can distinguish between the two cases.

PR #2846, opened 2025-12-15
Add dot as an alias to empty scope in requests

oguzhand95:
# Description
Adds `.` as an alias to empty scopes (``) in requests.
This is particularly useful when `engine.defaultScope` is configured in cerbos configuration.

## Updated regex for `scope` field
`^(^$|\.|[0-9a-zA-Z][\w\-]*(\.\w[\w\-]*)*)$`
regex for scope

# Usage
## CheckResources request
```json
{
  "requestId": "test", 
  "principal": {
    "id": "alice", 
    "policyVersion": "default", 
    "scope": ".", 
  },
  "resources": [ 
    {
      "resource": {
        "id": "XX1", 
        "kind": "leave_request", 
        "policyVersion": "default", 
        "scope": "."
      },
      "actions": [ 
        "approve"
      ]
    }
  ]
}
```

### DIFF
--- a/internal/test/testdata/engine/case_35.yaml
+++ b/internal/test/testdata/engine/case_35.yaml
@@ -1,0 +1,141 @@
+# yaml-language-server: $schema=../.jsonschema/EngineTestCase.schema.json
+---
+description: "Use dot alias for scope"
+inputs:
+  [
+    {
+      "requestId": "test",
+      "actions": ["view:public", "approve"],
+      "principal":
+        {
+          "id": "john",
+          "scope": ".",
+          "policyVersion": "20210210",
+          "roles": ["employee"],
+          "attr":
+            { "department": "marketing", "geography": "GB", "team": "design" },
+        },
+      "resource":
+        {
+          "kind": "leave_request",
+          "scope": ".",
+          "policyVersion": "20210210",
+          "id": "XX125",
+          "attr":
+            {
+              "department": "marketing",
+              "geography": "GB",
+              "id": "XX125",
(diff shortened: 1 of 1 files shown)


internal/evaluator/evaluator.go
```diff
@@ -108,5 +109,5 @@ func Scope(scope string, params EvalParams) string {
 		scope = params.DefaultScope
 	}
 
-	return scope
+	return strings.TrimPrefix(scope, ".")
```
charithe, line 112: Shouldn't this be: 
```suggestion
scope = strings.TrimPrefix(scope, ".")
if scope == "" {
scope = params.DefaultScope
}

	return scope
```

internal/ruletable/planner/planner.go
```diff
@@ -733,12 +734,12 @@ func PlanResourcesInputToRequest(input *enginev1.PlanResourcesInput) *enginev1.R
 			Id:    input.Principal.Id,
 			Roles: input.Principal.Roles,
 			Attr:  input.Principal.Attr,
-			Scope: input.Principal.Scope,
+			Scope: strings.TrimPrefix(input.Principal.Scope, "."),
```
charithe: I'd make this into a function so that we don't have to repeat the logic everywhere.


```suggestion
			Scope: namer.ScopeValue(input.Principal.Scope),
```

internal/evaluator/evaluator.go
charithe: Ignore me 🤦🏽

PR #2849, opened 2025-12-15
fix(planner): Union nodes across scope boundaries

Sambigeara:
For a given role, ALLOW_OVERRIDE scope boundaries should `or` the nodes across each scope. Previously, the evaluation was breaking out of the scope loop too eagerly.

### DIFF
--- a/internal/test/testdata/server/plan_resources/plan_case_01_acme.hr.uk.yaml
+++ b/internal/test/testdata/server/plan_resources/plan_case_01_acme.hr.uk.yaml
@@ -38,19 +38,10 @@ planResources:
     "resourceKind": "leave_request",
     "policyVersion": "default",
     "filter": {
-      "kind": "KIND_CONDITIONAL",
-      "condition": {
-        "expression": {
-          "operator": "eq",
-          "operands": [
-            { "variable": "request.resource.attr.owner" },
-            { "value": "harry"}
-          ]
-        }
-      }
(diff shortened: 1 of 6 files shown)


internal/test/testdata/server/plan_resources/plan_case_01_acme.hr.uk.yaml
```diff
@@ -38,19 +38,10 @@ planResources:
...
-      "condition": {
-        "expression": {
-          "operator": "eq",
-          "operands": [
-            { "variable": "request.resource.attr.owner" },
-            { "value": "harry"}
-          ]
-        }
-      }
+      "kind": "KIND_ALWAYS_ALLOWED"
```
charithe, line 41: I was baffled by this for a bit but it does make sense. Because `acme.hr.uk` has no explicit deny, a request would go up to the parent scope and the `acme.hr` policy has a rule that's always true. Therefore the result changes to ALWAYS_ALLOWED.

charithe (approved):
Thanks for looking into it.

internal/test/testdata/server/plan_resources/plan_case_01_acme.hr.uk.yaml
Sambigeara: Yes, exactly that! I added the new `plan_case_02_acme.hr.uk.yaml` case to bypass the `ALWAYS_ALLOW` condition to try to recreate the same assertions that were originally in here.

PR #2881, opened 2026-01-06
enhancement: Batching tests

dbuduev:
Batching tests (multiple actions) into a single check call. 

`CERBOS_BENCH_DIR=/abspath/to/policies go test -bench=BenchmarkVerify -run='^$' ./internal/verify/`
```text
goos: linux
goarch: arm64
         │ before.txt │             after.txt              │
         │   sec/op   │   sec/op    vs base                │
Verify-5   7.433 ± 1%   1.279 ± 8%  -82.79% (p=0.000 n=10)

         │  before.txt  │              after.txt               │
         │     B/op     │     B/op      vs base                │
Verify-5   7.397Gi ± 0%   1.188Gi ± 0%  -83.94% (p=0.000 n=10)

         │  before.txt  │              after.txt              │
         │  allocs/op   │  allocs/op   vs base                │
Verify-5   27.535M ± 0%   4.650M ± 0%  -83.11% (p=0.000 n=10)
```

### DIFF
--- a/internal/verify/run_test_suite.go
+++ b/internal/verify/run_test_suite.go
@@ -90,9 +90,7 @@ func runTestSuite(ctx context.Context, eng Checker, filter *testFilter, file str
 			continue
 		}
 
-		useBatching := !skipBatching &&
-			!expectedOutput &&
-			len(test.Input.Actions) > 1
+		useBatching := !skipBatching && !expectedOutput
 
 		if useBatching {
 			actionResults := runTest(ctx, eng, test, test.Input.Actions, trace)


internal/verify/run_test_suite.go
```diff
@@ -83,8 +90,18 @@ func runTestSuite(ctx context.Context, eng Checker, filter *testFilter, file str
 			continue
 		}
 
-		for _, action := range test.Input.Actions {
-			addResult(results, test.Name, action, runTest(ctx, eng, test, action, trace))
+		useBatching := !skipBatching && !expectedOutput
```
charithe, line 93: Why is batching disabled when there are expected outputs? Is it because it's a bit of a pain to go through the list of outputs and match them to the test?

cmd/cerbos/compile/compile.go
```diff
@@ -59,6 +59,7 @@ type Cmd struct { //betteralign:ignore
 	Tests         string                            `help:"[Deprecated] Path to the directory containing tests. Defaults to policy directory." type:"path"`
 	RunRegexp     string                            `help:"Run only tests that match this regex" name:"run"`
 	SkipTests     bool                              `help:"Skip tests"`
+	SkipBatching  bool                              `help:"Skip batching tests"`
```
charithe, line 62: Do we need this flag?

haines: I think if you are debugging a test the engine traces will be easier to understand with batching disabled. However we could achieve this without the flag by disabling batching if tracing is enabled (which implies that the `--verbose` flag was specified).

haines: I think if you are debugging a test the engine traces will be easier to understand with batching disabled. However we could achieve this without the flag by disabling batching if tracing is enabled (which implies that the `--verbose` flag was specified).

internal/verify/run_test_suite.go
dbuduev: Yes, the outputs are merged when the test matrix is built. Matching output can be done in a separate PR.

cmd/cerbos/compile/compile.go
charithe: Good point! There are instances where we always enable tracing for tests and it would be a shame to lose the performance benefit of batching in those cases. Therefore, having an explicit flag to disable batching seems better to me.

haines: Where do we do this? I think for example in Hub we should disable batching because we have tracing and want to be able to display an easy-to-interpret trace to users.

haines: Where do we do this? I think for example in Hub we should disable batching because we have tracing and want to be able to display an easy-to-interpret trace to users.

charithe: It's a bit complicated. See the original issue and let's continue this discussion there.

PR #2883, opened 2026-01-07
Cache-friendly time decorator, caching CEL Programs

dbuduev:
This PR adds caching for CEL programs. To enable caching, we need to introduce a cache-friendly time decorator. The existing decorator calls _nowFn_ at compilation time and inserts the result as a constant value. The new decorator looks up the  _nowFn_ from the activation at eval time.

This PR removes caching of CEL ASTs. Having the CEL programs cache, the AST cache is no longer needed.

```txt
goos: linux
goarch: arm64
         │  before.txt  │              after.txt              │
         │    sec/op    │   sec/op     vs base                │
Verify-5   1279.4m ± 8%   173.9m ± 2%  -86.41% (p=0.000 n=10)

         │  before.txt   │              after.txt               │
         │     B/op      │     B/op      vs base                │
Verify-5   1216.1Mi ± 0%   153.5Mi ± 0%  -87.38% (p=0.000 n=10)

         │ before.txt  │              after.txt              │
         │  allocs/op  │  allocs/op   vs base                │
Verify-5   4.650M ± 0%   2.341M ± 0%  -49.65% (p=0.000 n=10)
```

### DIFF
--- a/internal/conditions/cerbos_lib.go
+++ b/internal/conditions/cerbos_lib.go
@@ -202,8 +202,8 @@ func (n *nowInterp) Eval(activation interpreter.Activation) ref.Val {
 
 // timeSinceInterp is a custom Interpretable that looks up NowFunc from the activation at eval time.
 type timeSinceInterp struct {
-	id  int64
 	arg interpreter.Interpretable
+	id  int64
 }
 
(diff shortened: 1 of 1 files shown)


charithe (approved):
🙌🏽

PR #2885, opened 2026-01-09
enhancement: Stream test suite results

dbuduev:
Add verify.VerifyStream function that streams test suite results via channel as each suite completes.
By default, it runs tests concurrently.

```txt
goos: linux
goarch: arm64
         │ ../cerbos-main/before.txt │              after.txt              │
         │          sec/op           │   sec/op     vs base                │
Verify-5                 173.9m ± 5%   108.4m ± 3%  -37.65% (p=0.000 n=10)

         │ ../cerbos-main/before.txt │              after.txt              │
         │           B/op            │     B/op      vs base               │
Verify-5                153.5Mi ± 0%   153.0Mi ± 0%  -0.31% (p=0.001 n=10)

         │ ../cerbos-main/before.txt │             after.txt              │
         │         allocs/op         │  allocs/op   vs base               │
Verify-5                 2.342M ± 0%   2.340M ± 0%  -0.07% (p=0.001 n=10)
```

### DIFF
--- a/internal/verify/verify.go
+++ b/internal/verify/verify.go
@@ -65,8 +65,7 @@ func Verify(ctx context.Context, fsys fs.FS, eng Checker, conf Config) (*policyv
 
 // VerifyStream runs test suites and streams results as each suite completes.
 // It returns the number of test suites, a channel of results, and any setup error.
-// Callers must either consume all results from the channel or cancel the context
-// to avoid goroutine leaks.
+// Callers may cancel the context to stop workers early and avoid unnecessary work.
 func VerifyStream(ctx context.Context, fsys fs.FS, eng Checker, conf Config) (int, <-chan SuiteResult, error) {
 	suiteDefs, fixtureDefs, err := discoverTestFiles(ctx, fsys)
 	if err != nil {


internal/verify/verify.go
```diff
@@ -63,97 +181,115 @@ func Verify(ctx context.Context, fsys fs.FS, eng Checker, conf Config) (*policyv
...
 
-		results.Summary.TestsCount += suiteResult.Summary.TestsCount
+func runConcurrent(ctx context.Context, suiteDefs []string, workers int, runSuite func(string) *policyv1.TestResults_Suite, results chan<- SuiteResult) {
+	jobs := make(chan string, len(suiteDefs))
+	var wg sync.WaitGroup
+
+	for range workers {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
```
charithe, line 250:
```suggestion
		wg.Go(func() {
```

internal/verify/verify.go
```diff
@@ -18,26 +20,143 @@ import (
...
+	if err != nil {
+		return 0, nil, err
+	}
+
+	testFilter, err := newTestFilter(&conf)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	results := make(chan SuiteResult)
```
charithe: Might be better to make this a buffered channel because it'd otherwise make things effectively serial.

dbuduev:
@charithe I've reverted the change that used `(*WaitGroup).Go` method, since we haven't updated the project to Go 1.25 yet.

charithe (approved):

PR #2886, opened 2026-01-09
chore: Rule table index refactor

Sambigeara:
Primarily:

* `GetRows` now more generically accepts slices of all each rule category (e.g. optionally many resources, versions etc. rather than just one of each). Empty slices denote "match all" behaviour. It also accepts an optional arg which determines whether or not the function returns literal rows (rather than inferring functional rows which differ from the indexed ones in the case of role policy rows).
* Exposed a public `ListKeys` method to return all keys for a given category from the index.
* Lifted the `AddParentRoles` function from the rule table to the index.

### DIFF
--- a/internal/ruletable/index/index.go
+++ b/internal/ruletable/index/index.go
@@ -54,6 +54,7 @@ type literalMap interface {
 	set(context.Context, string, *rowSet) error
 	get(context.Context, ...string) (map[string]*rowSet, error)
 	getAll(context.Context) (map[string]*rowSet, error)
+	getAllKeys(context.Context) (map[string]struct{}, error)
 	delete(context.Context, ...string) error
 }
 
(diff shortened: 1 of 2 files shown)


internal/ruletable/index/index.go
```diff
@@ -411,6 +447,33 @@ func (m *Impl) IndexRules(ctx context.Context, rows []*Row) error {
 	return nil
 }
 
+func (m *Impl) ListKeys(ctx context.Context, cat CategoryKey) ([]string, error) {
+	var all map[string]*rowSet
+	var err error
+	switch cat {
+	case CategoryKeyActionGlob:
+		all, err = m.actionGlob.getAll(ctx)
```
charithe: Probably worth adding a function to `globMap` that returns just the keys because returning the values (which are promptly discarded) has some overhead. Particularly in the case of the Redis index, it would be more efficient to call `Keys` compared to reading the values over the network.

internal/ruletable/index/index.go
```diff
@@ -492,40 +555,75 @@ func (m *Impl) GetAllRows(ctx context.Context) ([]*Row, error) {
...
+	var versionSets map[string]*rowSet
+	var err error
+	if len(versions) > 0 {
+		versionSets, err = m.version.get(ctx, versions...)
+	} else {
+		versionSets, err = m.version.getAll(ctx)
+		versions = make([]string, 0, len(versionSets))
+		for v := range versionSets {
+			versions = append(versions, v)
+		}
```
charithe:
```suggestion
	versions = slices.Collect(maps.Keys(versionSets))

```

internal/ruletable/index/index.go
```diff
@@ -492,40 +555,75 @@ func (m *Impl) GetAllRows(ctx context.Context) ([]*Row, error) {
...
-	resourceSets, err := m.resourceGlob.getMerged(ctx, resource)
+	var resourceSets map[string]*rowSet
+	if len(resources) > 0 {
+		resourceSets, err = m.resourceGlob.getMerged(ctx, resources...)
+	} else {
+		resourceSets, err = m.resourceGlob.getAll(ctx)
+		resources = make([]string, 0, len(resourceSets))
+		for v := range resourceSets {
+			resources = append(resources, v)
+		}
```
charithe:
```suggestion
		resources = slices.Collect(maps.Keys(resourceSets))

```

charithe: There are few more similar occurrences below as well. Nothing wrong with them except that the single line version is easier to read.

internal/ruletable/index/index.go
Sambigeara: Very good point!

charithe: I think you forgot to change these to `getAllKeys` 🙂

Sambigeara: Oh deary me. Thanks.

PR #2893, opened 2026-01-12
chore: getAllKeys returns slice instead of map

dbuduev:
@Sambigeara, sorry, I was late reviewing your PR https://github.com/cerbos/cerbos/pull/2886

Does getAllKeys need to return a map rather than a slice?


dbuduev:
> Apologies, I should probably have awaited your review before merging the previous PR!

All good, no worries.

PR #2895, opened 2026-01-12
Add DeletePolicy RPC

oguzhand95:
# Description
It is now possible to delete policies from mutable stores. If the to be deleted policies breaks the scope chain or there are dependents to those policies, the API rejects the request with gRPC invalid argument status.

### Response
```json
{
    "code": 3,
    "message": "Failed to delete policies",
    "details": [
        {
            "@type": "type.googleapis.com/cerbos.response.v1.DeletePolicyErrorDetails",
            "errors": {
                "resource.leave_request.vdefault": {
                    "breaksScopeChain": {
                        "descendants": [
                            "resource.leave_request.vdefault/acme",
                            "resource.leave_request.vdefault/acme.hr",
                            "resource.leave_request.vdefault/acme.hr.tr",
                            "resource.leave_request.vdefault/acme.hr.uk"
                        ]
                    }
                }
            }
        }
    ]
}
```

### Logs (simplified for clarity)
```json
{
  "log.level": "error",
  "message": "Failed to delete policies",
  "grpc.method": "DeletePolicy",
  "error": "cannot perform delete operation: breaks scope chains [resource.leave_request.vdefault]"
}
```

### Response
```json
{
    "code": 3,
    "message": "Failed to delete policies",
    "details": [
        {
            "@type": "type.googleapis.com/cerbos.response.v1.DeletePolicyErrorDetails",
            "errors": {
                "derived_roles.common_roles": {
                    "requiredByOtherPolicies": {
                        "dependents": [
                            "resource.leave_request.vdefault",
                            "resource.leave_request.vdefault/acme.hr",
                            "resource.leave_request.vdefault/acme.hr.tr",
                            "resource.leave_request.vdefault/acme.hr.uk",
                            "resource.leave_request.vdefault/acme"
                        ]
                    }
                }
            }
        }
    ]
}
```

### Logs (simplified for clarity)
```json
{
  "log.level": "error",
  "message": "Failed to delete policies",
  "grpc.method": "DeletePolicy",
  "error": "cannot perform delete operation: required by other policies [derived_roles.common_roles]"
}
```

### Response
```json
{
    "code": 3,
    "message": "Failed to delete policies",
    "details": [
        {
            "@type": "type.googleapis.com/cerbos.response.v1.DeletePolicyErrorDetails",
            "errors": {
                "derived_roles.common_roles": {
                    "requiredByOtherPolicies": {
                        "dependents": [
                            "resource.leave_request.vdefault",
                            "resource.leave_request.vdefault/acme.hr",
                            "resource.leave_request.vdefault/acme.hr.tr",
                            "resource.leave_request.vdefault/acme.hr.uk",
                            "resource.leave_request.vdefault/acme"
                        ]
                    }
                },
                "resource.leave_request.vdefault": {
                    "breaksScopeChain": {
                        "descendants": [
                            "resource.leave_request.vdefault/acme",
                            "resource.leave_request.vdefault/acme.hr",
                            "resource.leave_request.vdefault/acme.hr.tr",
                            "resource.leave_request.vdefault/acme.hr.uk"
                        ]
                    }
                }
            }
        }
    ]
}
```

### Logs (simplified for clarity)
```json
{
  "log.level": "error",
  "message": "Failed to delete policies",
  "grpc.method": "DeletePolicy",
  "error": "cannot perform delete operation: breaks scope chains [resource.leave_request.vdefault]: required by other policies [derived_roles.common_roles]"
}
```

### DIFF
--- a/internal/storage/db/internal/tests.go
+++ b/internal/storage/db/internal/tests.go
@@ -394,40 +394,6 @@ func TestSuite(store DBStorage) func(*testing.T) {
 			}
 		})
 
-		t.Run("delete", func(t *testing.T) {
-			checkEvents := storage.TestSubscription(store)
-
-			_, err := store.Delete(ctx, namer.PolicyKeyFromFQN(rpx.FQN))
-			require.NoError(t, err)
-
-			have, err := store.GetCompilationUnits(ctx, rpx.ID)
-			require.NoError(t, err)
-			require.Empty(t, have)
-
-			wantEvents := []storage.Event{
-				{Kind: storage.EventDeleteOrDisablePolicy, PolicyID: rpx.ID},
-			}
-			checkEvents(t, timeout, wantEvents...)
-
-			_, err = store.Delete(ctx, namer.PolicyKeyFromFQN(dr.FQN))
-			var integrityErr *db.IntegrityErr
-			require.ErrorAs(t, err, &integrityErr)
-			require.Contains(t, integrityErr.Errors, namer.PolicyKey(dr.Policy))
-			require.Len(t, integrityErr.Errors, 1)
-			require.Len(t, integrityErr.Errors[namer.PolicyKey(dr.Policy)].RequiredByOtherPolicies.Dependents, 4)
-			require.Contains(t, integrityErr.Errors[namer.PolicyKey(dr.Policy)].RequiredByOtherPolicies.Dependents, namer.PolicyKey(rp.Policy))
-			require.Contains(t, integrityErr.Errors[namer.PolicyKey(dr.Policy)].RequiredByOtherPolicies.Dependents, namer.PolicyKey(rpAcme.Policy))
-			require.Contains(t, integrityErr.Errors[namer.PolicyKey(dr.Policy)].RequiredByOtherPolicies.Dependents, namer.PolicyKey(rpAcmeHR.Policy))
-			require.Contains(t, integrityErr.Errors[namer.PolicyKey(dr.Policy)].RequiredByOtherPolicies.Dependents, namer.PolicyKey(rpAcmeHRUK.Policy))
-
-			_, err = store.Delete(ctx, namer.PolicyKeyFromFQN(rpAcmeHR.FQN))
-			require.ErrorAs(t, err, &integrityErr)
-			require.Contains(t, integrityErr.Errors, namer.PolicyKey(rpAcmeHR.Policy))
-			require.Len(t, integrityErr.Errors, 1)
-			require.Len(t, integrityErr.Errors[namer.PolicyKey(rpAcmeHR.Policy)].BreaksScopeChain.Descendants, 1)
-			require.Contains(t, integrityErr.Errors[namer.PolicyKey(rpAcmeHR.Policy)].BreaksScopeChain.Descendants, namer.PolicyKey(rpAcmeHRUK.Policy))
-		})
-
 		t.Run("add_schema", func(t *testing.T) {
 			checkEvents := storage.TestSubscription(store)
 			require.NoError(t, store.AddOrUpdateSchema(ctx, &schemav1.Schema{Id: schID, Definition: sch}))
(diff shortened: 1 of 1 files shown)


api/public/cerbos/response/v1/response.proto
```diff
@@ -427,6 +427,13 @@ message GetPolicyResponse {
   repeated cerbos.policy.v1.Policy policies = 1;
 }
 
+message DeletePolicyResponse {
```
charithe, line 430: Drive-by comment: it'd be nice if the RPC returns the reasons why a given policy cannot be deleted. For example:

```
- derived_roles.foo:
    reason: REQUIRED_BY_OTHER_POLICIES
    dependents: 
      - resource.bar.vdefault
- resource.baz.vdefault/a.b:
    reason: BREAKS_SCOPE_CHAIN
    descendants:
      - resource.baz.vdefault/a.b.c
      - resource.baz.vdefault/a.b.c.d
``` 
  

As usual, I haven't though through the complications of doing this 😅 Just take a look to see whether we can do it without paying a huge performance/complexity penalty.

oguzhand95: That'd certainly be nice. Here are the steps needed:

In order to determine dependents for `db.BreaksDependentsErr`, 
we need to query the database for the policy key by moduleIDs because `getDependents` work with moduleIDs.(`map[namer.ModuleID][]namer.ModuleID`)

In order to determine descendants for `db.BreaksScopeChainErr`,
we need to either modify the query `HasDescendants` or add a new query such as `GetDescendants` where it returns something like `map[namer.ModuleID][]namer.ModuleID`.
Since it has to work with moduleIDs, we have to resolve the moduleIDs into policyKeys here, too.

charithe: The IDs passed in the `DeletePolicies` request would be policy keys, right? (Because `ListPolicies` on a database store returns policy keys.) We can derive the module ID from that and find the dependants and descendants.

.changelog/20260115_0845.txt
```diff
@@ -0,0 +1,3 @@
+type:feature
+Add DeletePolicies RPC
+It is now possible to delete policies from mutable stores. If the to be deleted policies breaks the scope chain or there are dependents to those policies, the API rejects the request with gRPC invalid argument status.
```
charithe:
```suggestion
Allows safe deletion of policies from database stores via the Admin API. Any deletion that would leave the policy store in an invalid state is rejected. 
```

docs/modules/api/pages/admin_api.adoc
```diff
@@ -200,13 +200,39 @@ curl -k -u cerbos:cerbosAdmin \
...
+
+----
+POST   /admin/policy/delete?id=policy_id
+PUT    /admin/policy/delete?id=policy_id
+----
+
+NOTE: This endpoint requires a mutable storage driver such as xref:configuration:storage.adoc#sqlite3[sqlite3] to be configured.
+
+Issue a POST request to the endpoint with the list of IDs (the `id` query parameter can be repeated multiple times) to delete.
+The ID is of the form `<kind>.<name>.v<version>/<scope>`. A resource policy for `leave_request` with version `default` and scope `acme.hr` would therefore have the ID `resource.leave_request.vdefault/acme.hr`.
```
charithe:
```suggestion
IDs follow the same format as the output of `ListPolicies` (`<kind>.<name>.v<version>/<scope>`). For example, a resource policy for `leave_request` with version `default` and scope `acme.hr` would have the ID `resource.leave_request.vdefault/acme.hr`.
```

internal/storage/db/internal/db.go
```diff
@@ -606,92 +606,302 @@ func (s *dbStorage) getDependents(ctx context.Context, tx *goqu.TxDatabase, ids
...
+
+	// SELECT
+	//	policy_dependency.dependency_id AS policy_id,
+	//	policy_dependency.policy_id AS dependent_id
+	//  policy.kind
+	//  policy.name
+	//  policy.version
+	//  policy.scope
+	// FROM policy_dependency
+	// JOIN policy ON (policy.id = policy_dependency.policy_id)
```
charithe: I think we need to account for disabled policies (although it'd be quite rare in practice).


```suggestion
	// JOIN policy ON (policy.id = policy_dependency.policy_id) AND (policy.disabled = false)
```

internal/storage/db/internal/db.go
```diff
@@ -723,71 +933,88 @@ func (s *dbStorage) Enable(ctx context.Context, policyKey ...string) (uint32, er
...
+) error {
+	descendants, err := s.GetDescendants(ctx, mIDs...)
+	if err != nil {
+		return fmt.Errorf("failed to get descendants for policies: %w", err)
+	}
 
-			if deps, ok := dependents[ids[0]]; ok && len(deps) > 0 {
-				ev.Dependents = deps
+	for mID, policyKey := range mIDPolicyKey {
+		if descs, ok := descendants[mID]; ok && len(descs) > 0 {
```
charithe, line 974: I don't think this is handling the case where all the descendants of the module are included in the delete list?

Say there are three policies `resource.foo.v1`, `resource.foo.v1/a` and `resource.foo.v1/a.b`. If the delete list contains all three of them, then the scope chain isn't broken because they are all getting deleted at the same time.

internal/storage/db/internal/db.go
```diff
@@ -723,71 +933,88 @@ func (s *dbStorage) Enable(ctx context.Context, policyKey ...string) (uint32, er
...
-		}
+// writeRequiredByOtherPoliciesErrors checks whether deleting or disabling the given policies break dependents and writes the relevant information to out map.
+func (s *dbStorage) writeRequiredByOtherPoliciesErrors(
+	out map[string]*responsev1.IntegrityErrors,
+	mIDPolicyKey map[namer.ModuleID]string,
+	dependents map[namer.ModuleID][]namer.Policy,
+) error {
+	for mID, deps := range dependents {
+		if len(deps) > 0 {
+			policyKey, ok := mIDPolicyKey[mID]
```
charithe, line 1005: Same as above; if all the dependants are already included in the delete request, that's fine.

api/public/cerbos/request/v1/request.proto
```diff
@@ -610,6 +610,27 @@ message GetPolicyRequest {
...
 
+message DeletePolicyRequest {
+  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
+    json_schema: {description: "Delete policy request"}
+  };
+  repeated string id = 1 [
+    (google.api.field_behavior) = REQUIRED,
+    (buf.validate.field).repeated = {
+      unique: true
+      min_items: 1
```
charithe, line 621: Let's add a `max_items` as well because the descendants query is going to blow up if the number of policies is too large.

api/public/cerbos/svc/v1/svc.proto
```diff
@@ -155,14 +155,33 @@ service CerbosAdminService {
     };
   }
 
+  rpc DeletePolicy(cerbos.request.v1.DeletePolicyRequest) returns (cerbos.response.v1.DeletePolicyResponse) {
+    option (google.api.http) = {
+      post: "/admin/policy/delete"
+      additional_bindings: {
+        put: "/admin/policy/delete"
+        body: "*"
+      }
```
charithe: I don't think we need `PUT` here. The "correct" RESTful thing would be `DELETE /admin/policy` but there are conflicting views on whether `DELETE` requests should have a body or not so `POST` seems like the safer alternative.

internal/storage/db/store.go
```diff
@@ -8,12 +8,37 @@ package db
...
+		}
+
+		if ie.RequiredByOtherPolicies != nil && len(ie.RequiredByOtherPolicies.Dependents) > 0 {
+			policiesRequiredByOtherPolicies = append(policiesRequiredByOtherPolicies, policyKey)
+		}
+	}
+
+	if len(policiesBreakingScopeChain) > 0 && len(policiesRequiredByOtherPolicies) > 0 {
+		return fmt.Sprintf(
+			"removing the scoped policies [%s] will break the scope chain(s) and removing the policies [%s] will break the dependent policies",
```
charithe: This is a confusing error message because one of the arrays could be empty, which would result in something like "removing scoped policies [] will break scope chains...". 

I'd build up the error string in stages.

```go
sb := new(strings.Builder)
sb.WriteString("cannot perform delete operation:")
if len(policiesBreakingScopeChain) > 0 {
    sb.WriteString("breaks scope chains: [")
    sb.WriteString(strings.Join(policiesBreakingScopeChain, ","))
    sb.WriteString("]")
}
...
```

internal/storage/db/store.go
```diff
@@ -8,12 +8,37 @@ package db
...
+			strings.Join(policiesBreakingScopeChain, ", "),
+			strings.Join(policiesRequiredByOtherPolicies, ", "),
+		)
+	} else if len(policiesBreakingScopeChain) != 0 {
+		return fmt.Sprintf("removing the following scoped policies will break the scope chain: %s", strings.Join(policiesBreakingScopeChain, ", "))
+	}
+
+	return fmt.Sprintf("removing the following policies will break the dependent policies: %s", strings.Join(policiesRequiredByOtherPolicies, ", "))
```
charithe: These are not needed if the error message is constructed as I suggested above.

charithe (commented):
Please add tests to check the following cases:

Setup 
- resources R1 and R2 import derived roles D
- D imports variables V 

Deleting R1, D and V should fail
Deleting R1, R2 and V should fail
Deleting R1, R2 and D should succeed

charithe (approved):

haines:
This seems to be breaking the JS SDK tests

```
failed to write 'breaks scope chain' errors: failed to get descendants for policies: could not execute \"GetDescendants\" query: SQL logic error: no such table: policy_ancestor (1)

rpc error: code = Internal desc = Failed to delete policies
```

The tests are hitting a PDP configured with

```yaml
storage:
  driver: sqlite3
  sqlite3:
    dsn: ":memory:"
```

oguzhand95:
Would you mind trying with this configuration, @haines?

```yaml
storage:
  driver: "sqlite3"
  sqlite3:
    dsn: "file::memory:?cache=shared"
```

haines:
With that configuration it seems to hang on the first `AddOrUpdatePolicy` call and the test times out.

PR #2897, opened 2026-01-13
Add PurgeStoreRevisions RPC

oguzhand95:
# Description
Cerbos stores the history of changes applied to the policies in the `policy_revision` table when mutable stores are used. If `keepLast` query parameter is not specified, `PurgeStoreRevisions` deletes all revisions from the table.
If the query parameter is specified, for each policy it keeps the last `N` revision rows in the table.

# Usage
```
DELETE /admin/store/revisions
```

### DIFF
--- a/.changelog/20260115_1354.txt
+++ b/.changelog/20260115_1354.txt
@@ -1,3 +1,3 @@
 type:feature
 Add PurgeStoreRevisions RPC
-When Cerbos is configured with a database store and the Admin API is used to add/update policies, each operation creates a backup copy of the policies being modified. The `PurgeStoreRevisions` Admin RPC allows deleting those backups to reclaim space.
+When Cerbos is configured with a database store and the Admin API is used to add/update policies, each operation creates a backup copy of the policies being modified. The `PurgeStoreRevisions` Admin RPC allows deleting those backups to reclaim space. Please note that this feature requires the database user used by the PDP to have `DELETE` privileges on the `policy_revisions` table. If you have an existing deployment, this grant needs to be added manually by a database administrator. 


.changelog/20260115_1354.txt
```diff
@@ -0,0 +1,3 @@
+type:feature
+Add PurgeStoreRevisions RPC
+Cerbos stores the history of changes applied to the policies in the policy_revision table when mutable stores are used. PurgeStoreRevisions purges all the rows from this table to clear up resources.
```
charithe:
```suggestion
When Cerbos is configured with a database store and the Admin API is used to add/update policies, each operation creates a backup copy of the policies being modified. The `PurgeStoreRevisions` Admin RPC allows deleting those backups to reclaim space.
```

api/public/cerbos/request/v1/request.proto
```diff
@@ -771,3 +771,9 @@ message ReloadStoreRequest {
     (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Wait until the reloading process finishes"}
   ];
 }
+
+message PurgeStoreRevisionsRequest {
```
charithe, line 798: I think we should let the users define how many revisions they would like to retain. 0 means delete everything, 1 means delete everything but the latest copy and so on.

api/public/cerbos/request/v1/request.proto
```diff
@@ -771,3 +771,11 @@ message ReloadStoreRequest {
...
+  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
+    json_schema: {description: "Purge store revisions request"}
+  };
+
+  uint32 keep_last = 1 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Keep last N revisions. If not specified all revisions are going to be deleted."}];
```
charithe:
```suggestion
  uint32 keep_last = 1 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {description: "Keep last N revisions. If not specified or set to zero, all revisions will be deleted."}];
```

internal/storage/db/internal/db.go
```diff
@@ -1055,3 +1057,59 @@ func (s *dbStorage) CheckSchema(ctx context.Context) error {
...
+		// DELETE
+		// FROM policy_revision
+		if res, err = s.db.Delete(PolicyRevisionTbl).
+			Prepared(true).
+			Executor().
+			ExecContext(ctx); err != nil {
+			return 0, fmt.Errorf("failed to delete revisions: %w", err)
+		}
+	default:
+		// SELECT revision_id
```
charithe: This is not right because we need to keep the last N revisions of _each_ policy. Otherwise we'd just lose data for everything except for the most recently modified policies. 

You'd need a window query to do this. Roughly, something like the following.

```sql
WITH ranked AS (
  SELECT revision_id, ROW_NUMBER() OVER (PARTITION BY id ORDER BY updated_at DESC) AS revision
  FROM policy_revision
)
DELETE FROM policy_revision 
WHERE revision_id IN (SELECT revision_id FROM ranked WHERE revision > ?)
```

Please add a test where you add at least two policies initially. Then update one of those policies once to create a single revision. Update the other one multiple times to create several revisions. When you do a `Purge` with `keep_last=1`, you should end up with two records left in the table with one record each for each of the policies.

internal/storage/db/internal/db.go
```diff
@@ -1320,3 +1323,85 @@ func (s *dbStorage) CheckSchema(ctx context.Context) error {
...
+func (s *dbStorage) ListRevisions(ctx context.Context, ids ...namer.ModuleID) (map[namer.ModuleID]int, error) {
+	// SELECT
+	//  id,
+	//  COUNT(*) AS count
+	// FROM policy_revision
+	// GROUP BY id
+	query := s.db.
+		Select(
+			PolicyRevisionTblIDCol,
+			goqu.COUNT("*").As("count"),
```
charithe:
```suggestion
			goqu.COUNT("1").As("count"),
```

internal/storage/db/internal/db.go
```diff
@@ -1320,3 +1323,85 @@ func (s *dbStorage) CheckSchema(ctx context.Context) error {
...
+    SELECT revision_id
+    FROM (
+        SELECT revision_id, ROW_NUMBER() OVER (PARTITION BY id ORDER BY update_timestamp DESC) AS revision
+        FROM policy_revision
+    ) AS ranked
+    WHERE revision > %d
+);`, keepLast)
+
+		var err error
+		if res, err = s.db.ExecContext(ctx, query); err != nil {
```
charithe: We use `sqlx` so you could update the query to have named parameters like `WHERE revision > :keepLast`. Then, to execute it, do something like the following:

```go
		db, ok := s.db.Db.(*sqlx.DB)
		if !ok {
			// Error
		}

		rows, err := sqlx.NamedQueryContext(ctx, db, query, map[string]any{"keepLast": keepLast})
        ...
```

docs/modules/api/pages/admin_api.adoc
```diff
@@ -398,6 +398,31 @@ curl -k -u cerbos:cerbosAdmin -X DELETE \
...
 
+=== Purge store revisions
+
+----
+DELETE /admin/store/revisions
+----
+
+Issue a DELETE request to the endpoint to delete all rows in the `policy_revisions` table.
+
+Use `keepLast=N` query parameter to keep the last N revisions. If not specified all revisions are going to be deleted.
```
charithe:
```suggestion
Use `keepLast=N` query parameter to keep the last N revisions for each policy. If not specified, the entire table will be truncated.
```

internal/storage/db/internal/db.go
```diff
@@ -1320,3 +1323,85 @@ func (s *dbStorage) CheckSchema(ctx context.Context) error {
...
+
+// PurgeRevisions deletes revisions from the relevant table.
+func (s *dbStorage) PurgeRevisions(ctx context.Context, keepLast uint32) (uint32, error) {
+	var res sql.Result
+	switch keepLast {
+	case 0:
+		var err error
+		// DELETE
+		// FROM policy_revision
+		if res, err = s.db.Delete(PolicyRevisionTbl).
```
charithe, line 1379: Let's do truncate here because it's more efficient than delete.

https://pkg.go.dev/github.com/doug-martin/goqu/v9#Database.Truncate

oguzhand95: `sqlite` doesn't support `Truncate` statements.

.changelog/20260115_1354.txt
```diff
@@ -0,0 +1,3 @@
+type:feature
+Add PurgeStoreRevisions RPC
+When Cerbos is configured with a database store and the Admin API is used to add/update policies, each operation creates a backup copy of the policies being modified. The `PurgeStoreRevisions` Admin RPC allows deleting those backups to reclaim space.
```
charithe:
```suggestion
When Cerbos is configured with a database store and the Admin API is used to add/update policies, each operation creates a backup copy of the policies being modified. The `PurgeStoreRevisions` Admin RPC allows deleting those backups to reclaim space. Please note that this feature requires the database user used by the PDP to have `DELETE` privileges on the `policy_revisions` table. If you have an existing deployment, this grant needs to be added manually by a database administrator. 
```

charithe (dismissed):

PR #2900, opened 2026-01-14
Add action to output entries

dbuduev:


internal/verify/run_test_suite.go
```diff
@@ -304,13 +295,15 @@ func runTest(ctx context.Context, eng Checker, test *policyv1.Test, actions []st
...
-	for _, output := range actual[0].Outputs {
-		actualOutputs[output.Src] = output.Val
-	}
-
 	for _, action := range actions {
-		details := &policyv1.TestResults_Details{EngineTrace: traces}
+		clear(actualOutputs)
+		for _, output := range actual[0].Outputs {
+			if output.Action == action {
+				actualOutputs[output.Src] = output.Val
```
Sambigeara, line 302: I wonder if it would be worth building an ephemeral map: `map[action]output` here, reset for each action iteration and populated each time `output.Action == action` in this initial loop?

Might be overkill, but it could be referenced further down to avoid the second iteration 🤷‍♂️

charithe: Speaking of overkill, we could wrap the creation of `actualOutputs` in `sync.OnceValue` and only create/access it lazily when there are  expected outputs to process. It's probably not worth the added complexity though.

dbuduev: Yeah, it might be cleaner. I'll create a separate PR.

PR #2903, opened 2026-01-14
Replace SHA256 with xxhash function for Row deduplication

dbuduev:
This PR changes the hashing function used for `index.Row` from SHA256 to xxHash. 
⚠️ I assumed the records in Redis are transient and do not require migration.

If cryptographic properties aren't needed, the update brings the following benefits:
- faster map lookups,
- reduction in memory,
- faster hashing function.

Here are the stats for the BenchmarkEvaluator:
```txt
goos: linux
goarch: arm64
pkg: github.com/cerbos/cerbos/internal/engine
            │ ../cerbos-main/before.txt │              after.txt              │
            │          sec/op           │   sec/op     vs base                │
Evaluator-5                 65.12µ ± 5%   55.16µ ± 4%  -15.31% (p=0.000 n=10)

            │ ../cerbos-main/before.txt │              after.txt               │
            │           B/op            │     B/op      vs base                │
Evaluator-5                36.47Ki ± 0%   28.91Ki ± 0%  -20.74% (p=0.000 n=10)

            │ ../cerbos-main/before.txt │           after.txt            │
            │         allocs/op         │ allocs/op   vs base            │
Evaluator-5                  181.0 ± 0%   181.0 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

### DIFF
--- a/internal/ruletable/index/redis.go
+++ b/internal/ruletable/index/redis.go
@@ -74,7 +74,7 @@ func GetExistingRedis(ctx context.Context, client Cmdable, namespace string) (*R
 
 	return &Redis{
 		db:       client,
-		nsKey:    namespace,
+		nsKey:    namespace + ":",
 		sentKey:  sentKey,
 		readOnly: true,
 	}, nil
@@ -185,11 +185,11 @@ type redisMap struct {
 	readOnly         bool
 }
 
-func newRedisMap(db Cmdable, namespace, categoryKey, sentKey string, readOnly bool, sentinelDeadline, dataDeadline time.Time) *redisMap {
+func newRedisMap(db Cmdable, nsKey, categoryKey, sentKey string, readOnly bool, sentinelDeadline, dataDeadline time.Time) *redisMap {
 	return &redisMap{
 		db:               db,
-		nsKey:            namespace + ":",
-		catKey:           namespace + ":" + categoryKey,
+		nsKey:            nsKey,
+		catKey:           nsKey + categoryKey,
 		sentKey:          sentKey,
 		sentinelDeadline: sentinelDeadline,
 		dataDeadline:     dataDeadline,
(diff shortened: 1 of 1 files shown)


internal/ruletable/index/redis.go
```diff
@@ -169,9 +170,9 @@ func (r *Redis) resolveIter(ctx context.Context, rows iter.Seq[*Row]) (iter.Seq[
 	return slices.Values(resolved), nil
 }
 
-func (r *Redis) rowKey(sum string) string {
+func (r *Redis) rowKey(sum uint64) string {
 	// value is the serialised row
-	return r.nsKey + ":" + sum
+	return fmt.Sprintf("%s:%d", r.nsKey, sum)
```
charithe: Probably an unnecessary optimisation in the grand scheme of things 🙂 

If we update `r.nsKey` to include the `:` suffix when it's being initialised, we could do a slightly faster string concatenation here.

```suggestion
	return r.nsKey + strconv.FormatUint(sum ,10) 
```

internal/ruletable/index/redis.go
```diff
@@ -219,13 +220,15 @@ func (rm *redisMap) serialize(rs *rowSet) ([]any, []any, error) {
 	return sums, raws, nil
 }
 
-func (rm *redisMap) rowKey(sum string) string {
+func (rm *redisMap) rowKey(sum uint64) string {
 	// value is the serialised row
-	return rm.nsKey + ":" + sum
+	return fmt.Sprintf("%s:%d", rm.nsKey, sum)
```
charithe: Same as above.

charithe (commented):
LGTM. I don't think a cryptographic hash was needed here but let's wait for Sam to confirm that in case I missed something.

internal/ruletable/index/redis.go
```diff
@@ -169,9 +170,9 @@ func (r *Redis) resolveIter(ctx context.Context, rows iter.Seq[*Row]) (iter.Seq[
 	return slices.Values(resolved), nil
 }
 
-func (r *Redis) rowKey(sum string) string {
+func (r *Redis) rowKey(sum uint64) string {
 	// value is the serialised row
-	return r.nsKey + ":" + sum
+	return r.nsKey + strconv.FormatUint(sum, 10)
```
charithe, line 175: Line 102 needs to be updated to initialise `r.nsKey` as `namespace + ":"` and then lines 225 and 229 need to be updated to match.

Sambigeara:
This looks cool! Apologies, I'm afk until about 10:30 uk time, will take a look then

PR #2908, opened 2026-01-19
GlobMap clean-up

dbuduev:
This PR is a follow-up to https://github.com/cerbos/cerbos/pull/2905. It removes defensive locking and adds a comment on the GlobMap thread-safety.


dbuduev:
> Is there a way we can enforce locking? Do the internal locks add a lot of overhead?

The internal lock overhead was minimal. The reason for this update is consistency with other map fields (`globs`, `literals`, `compiled`). Access to them is regulated by convention - callers are expected to provide external synchronisation for writes. Currently, `memGlobMap`, which wraps `GlobMap`, uses its own `RWMutex` to synchronise all access.

The `matchCache` lock was only there because reads can trigger writes (cache population on miss).

Are you suggesting we protect all map fields internally? That would be a different design - making `GlobMap` fully self-synchronized rather than relying on external synchronization.

This would add overhead to a single-threaded use case (actionMatchedRows in `index.go`), which is on the hot path. 
Also, there are batch operations in `memGlobMap`. For example, `setBatch` holds one lock while calling `Set` multiple times.

> This feels a bit dangerous. Even though the current access patterns always happen inside a lock, it's possible that someone might forget to do so in the future. 

Should we make GlobMap unexported?

charithe:
I just feel uneasy about some fields being protected by a mutex while some aren't. We have to kind of rely on people reading the comment before use and keeping it updated when things change in the implementation. 

> Should we make GlobMap unexported?

Ideally. If it's used exclusively in this context, that'd help because it's clearly not meant to be used as a general purpose data structure.

PR #2915, opened 2026-01-20
Remove DELETE method from DeletePolicy RPC and move it back to DisablePolicy RPC as deprecated

oguzhand95:
# Description

- Reverts the `DELETE /admin/policy` method/endpoint changes.
- Fixes the `cerbosctl disable` test.

## History
- The following PR deprecated `DELETE /admin/policy` endpoint of `DisablePolicies` RPC: cerbos/cerbos#1476.
- The following PR introduced `DeletePolicy` RPC and I thought it was a good idea to use `DELETE /admin/policy` endpoint for `DeletePolicy` RPC (it wasn't): cerbos/cerbos#2895

- This PR removes `DELETE /admin/policy` from `DeletePolicy` RPC, and adds it back to `DisablePolicy` RPC as deprecated again. So basically this is a revert.

### DIFF
--- a/cmd/cerbosctl/disable/disable_test.go
+++ b/cmd/cerbosctl/disable/disable_test.go
@@ -83,7 +83,7 @@ func testDisableCmd(ctx context.Context, cctx *cmdclient.Context, globals *flags
 				}{
 					{
 						policyKey:    "derived_roles.my_derived_roles_1",
-						wantDisabled: true,
+						wantDisabled: false,
 					},
 					{
 						policyKey:    "principal.donald_duck_1.vdefault",


cmd/cerbosctl/disable/disable_test.go
```diff
@@ -83,7 +83,7 @@ func testDisableCmd(ctx context.Context, cctx *cmdclient.Context, globals *flags
 				}{
 					{
 						policyKey:    "derived_roles.my_derived_roles_1",
-						wantDisabled: true,
+						wantDisabled: false,
```
haines, line 86: Why has this changed?

oguzhand95: `derived_roles.my_derived_roles_1` is imported by other policies and disabling that policy breaks the store and it is not allowed anymore.

PR #2916, opened 2026-01-21
Add test filter to run selected tests only

dbuduev:
The PR adds a test filter for the `cerbos compile` command.

`cerbos compile --test-filter='suite=MySuite;test=album*;principal=alice;resource=my_album;action=view' /path/to/policy/repo`


charithe (commented):
LGTM. Let's mark the `run` flag as deprecated and remove references to it from the documentation. Please add a changelog entry as well because this is a new feature.

Also, as discussed, let's avoid printing skipped tests when they are skipped due to a filter.

dbuduev:
Should we also hide the deprecated options?

dbuduev:
I hid the `run` option. I'll change the number of tests in a separate PR.

PR #2917, opened 2026-01-21
enhancement: Add role policy versions support

Sambigeara:
Role policies now accept explicit versions, rather than implicitly assuming version as "default". This is an optional field for the time being. In a future release, it will be made a required field. If not provided, role policies assume version as "default" as before.

I been quite defensive (opting to cover the default-to-default case in a few locations), but we can pull these out when we eventually make the field required!

### DIFF
--- a/docs/modules/policies/pages/role_policies.adoc
+++ b/docs/modules/policies/pages/role_policies.adoc
@@ -39,7 +39,7 @@ rolePolicy:
       allowActions: ["create"]
 ----
 <1> The role to which this policy applies.
-<2> Version of this policy. Policies are uniquely identified by the role name and version pair. You can have multiple policy versions for the same role (e.g. production vs. staging). The version value `default` is special as it is the default fallback when no version is specified in the request. Currently optional (defaults to `default` if not specified), but will become required in a future release.
+<2> Version of this policy. Policies are uniquely identified by the role name and version pair. You can have multiple policy versions for the same role (e.g. production vs. staging). The version value `default` is special as it is the default fallback when no version is specified in the request. 
 <3> Optional principal xref:scoped_policies.adoc[scope] for this policy.
 <4> The list of parent roles that the custom role inherits.
 <5> The resource to which the following rule applies.


api/public/cerbos/policy/v1/policy.proto
```diff
@@ -131,6 +131,8 @@ message RolePolicy {
     option (buf.validate.oneof).required = true;
     string role = 1 [(buf.validate.field).string = {pattern: "^[^!*?\\[\\]{}]+$"}];
   }
+  // Optional for now; will become required in a future version. Defaults to "default" if not specified.
```
charithe: This comment is getting picked up by the documentation generators. I don't think we need to tell users that the field is optional because we do want anyone writing a new role policy to always define the version. The optionality of the field is just an internal implementation detail to avoid breaking existing users.

charithe (approved):

docs/modules/policies/pages/role_policies.adoc
```diff
@@ -18,30 +18,32 @@ A parent role can be either an arbitrary IdP role or the name of another role po
...
 ----
 <1> The role to which this policy applies.
-<2> Optional principal xref:scoped_policies.adoc[scope] for this policy.
-<3> The list of parent roles that the custom role inherits.
-<4> The resource to which the following rule applies.
-<5> The list of allowable actions that the role can carry out on the given resource.
-<6> Wildcard actions are supported.
-<7> A condition that must be met for the action to be allowed.
-<8> Wildcard resources are also supported.
+<2> Version of this policy. Policies are uniquely identified by the role name and version pair. You can have multiple policy versions for the same role (e.g. production vs. staging). The version value `default` is special as it is the default fallback when no version is specified in the request. Currently optional (defaults to `default` if not specified), but will become required in a future release.
```
charithe:
```suggestion
<2> Version of this policy. Policies are uniquely identified by the role name and version pair. You can have multiple policy versions for the same role (e.g. production vs. staging). The version value `default` is special as it is the default fallback when no version is specified in the request. 
```

internal/compile/compile.go
```diff
@@ -79,6 +79,11 @@ func compileRolePolicySet(modCtx *moduleCtx) *runtimev1.RunnablePolicySet {
 		return nil
 	}
 
+	version := rp.Version
+	if version == "" {
+		version = namer.DefaultVersion
```
haines, line 84: Isn't this a breaking change if I'm running a PDP configured with a different default?

PR #3057, opened 2026-03-27
Pool bitmaps

dbuduev:
Profiling the CPU under load showed GC works hard. To ease off GC pressure, I attempted to reuse bitmaps. The `FastOr` function allocates quite a few intermediate objects. Both `FastAnd` and `FastOr` were replaced with in-place operations, where an accumulator is allocated from the pool. 

The BenchmarkEvaluator has the same performance, but allocations dropped to 82 (from 100) per operation. The difference under load is significant. The RPS metric in the throughput test more than doubled, the response time distribution has a shorter tail.

Edit: @Sambigeara I can postpone this PR if it interferes with your work.

### DIFF
--- a/internal/ruletable/index/bitmap_pool.go
+++ b/internal/ruletable/index/bitmap_pool.go
@@ -13,8 +13,7 @@ var bitmapPool = sync.Pool{
 	New: func() any { return roaring.New() },
 }
 
-// emptyBitmap is a shared, immutable empty bitmap returned when a query has no
-// matches. Callers must not mutate it.
+// emptyBitmap is a shared. Callers must not mutate it.
 var emptyBitmap = roaring.New()
 
 // bitmapArena tracks pooled bitmaps acquired during a single query so they can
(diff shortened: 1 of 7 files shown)


Sambigeara:
> The RPS metric in the throughput test more than doubled

Insane 👏 . Seems natural to turn to GC when we're down to ~5us/req. Where to next? Zig? 😆

I don't think it should interfere with my work. I'm trying to keep serialisation as a bolt on concern, so should be fairly straightforward as an integration. I shall have a proper look at this this afto!

internal/ruletable/index/bitmap_pool.go
```diff
@@ -0,0 +1,100 @@
...
+
+// bitmapArena tracks pooled bitmaps acquired during a single query so they can
+// be released in bulk when the query completes.
+type bitmapArena struct {
+	used []*roaring.Bitmap
+}
+
+var arenaPool = sync.Pool{
+	New: func() any { return &bitmapArena{used: make([]*roaring.Bitmap, 0, 8)} }, //nolint:mnd
+}
```
Sambigeara: Do we gain much by using a sync.Pool for the bitmapArena too? I think each allocation is a slice header (24 bytes?) plus the backing array of 8 pointers (8*8 = 64 bytes) giving us 88 bytes total per query, seems fairly negligible? I might be wrong though, have not load tested.

dbuduev: > Do we gain much by using a sync.Pool for the bitmapArena too?

No gain at all. Currently the bitmap arena is allocated on stack if `acquireArena` is replaced with a simple  `newBitmapArena` function, so GC is free.

Sambigeara (approved):
🚀

PR #3094, opened 2026-04-09
Record store version information such as commit hash or bundle ID in audit log entries

oguzhand95:
# Description

This PR adds new fields `bundleID` (Hub)  and `hash` (Git) to [PolicySource](https://github.com/cerbos/cerbos/blob/946b67e6e0a8cf50de7bd258685447ce48c47a30/api/public/cerbos/audit/v1/audit.proto#L79-L147).


.changelog/20260409_1655.txt
```diff
@@ -0,0 +1,2 @@
+type:enhancement
+Add current bundle ID for hub and commit hash for git stores to policy source information
```
charithe:
```suggestion
Record store version information such as commit hash or bundle ID in audit log entries.
```

internal/storage/git/store.go
```diff
@@ -55,11 +55,12 @@ func init() {
...
-	log  *zap.SugaredLogger
-	conf *Conf
-	idx  index.Index
-	repo *git.Repository
-	sf   singleflight.Group
+	log    *zap.SugaredLogger
+	conf   *Conf
+	idx    index.Index
+	repo   *git.Repository
+	source *auditv1.PolicySource
```
charithe: This has a subtle issue. Every audit log entry gets this same object as their policy source. So, if the HEAD changes after processing a request but before its audit log entry is written, the audit entry will incorrectly report the new hash as its hash.

Instead of storing the protobuf, add a string field to store just the hash and protect it with a mutex. When the `Source` method is called, return a brand new protobuf object back to the caller to ensure that it doesn't get mutated unexpectedly.

internal/storage/hub/remote_source.go
```diff
@@ -259,7 +259,11 @@ func (s *RemoteSource) Init(ctx context.Context) error {
 		var source bundleapiv2.Source
 		switch {
 		case s.conf.Remote.DeploymentID != "":
-			hub.Source = &auditv1.PolicySource_Hub_DeploymentId{DeploymentId: s.conf.Remote.DeploymentID}
+			hub.Source = &auditv1.PolicySource_Hub_RemoteBundle_{
```
charithe: Same issue as above here. Now this information is dynamic and a call to `Source` should return a brand new protobuf object to avoid race conditions.

internal/storage/git/store.go
```diff
@@ -200,6 +216,10 @@ func (s *Store) Reload(ctx context.Context) error {
 		return fmt.Errorf("failed to pull: %w", err)
 	}
 
+	if err := s.setHash(); err != nil {
+		return fmt.Errorf("failed to set hash: %w", err)
```
charithe: We don't need to fail because of this. It's unlikely to happen anyway because the function above would have already failed if it couldn't get the HEAD.

internal/storage/git/store.go
```diff
@@ -218,11 +226,35 @@ func (s *Store) Source() *auditv1.PolicySource {
...
+func (s *Store) setCurrentCommitHash() {
+	if s.repo == nil {
+		return
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	head, err := s.repo.Head()
+	if err != nil {
```
charithe, line 251: Clear `currCommitHash` or set it to `UNKNOWN` because we don't know whether the current value is the right one or not.

charithe (approved):

PR #3095, opened 2026-04-09
Serializable bitmap index

Sambigeara:
Adds a wire format for the bitmap index so a pre-computed index can be persisted and loaded back without rebuilding from policies.


internal/inspect/ruletable.go
```diff
@@ -184,3 +187,108 @@ func getOrInitStringSet(m map[string]util.StringSet, key string) util.StringSet
 	m[key] = s
 	return s
 }
+
+func listActions(b *index.Binding) []string {
+	var actions []string
```
charithe: Probably worth pre-allocating here.

internal/inspect/ruletable.go
```diff
@@ -184,3 +187,108 @@ func getOrInitStringSet(m map[string]util.StringSet, key string) util.StringSet
...
+	}
+
+	if len(actions) > 1 {
+		slices.Sort(actions)
+	}
+
+	return actions
+}
+
+func listConstants(b *index.Binding) ([]*responsev1.InspectPoliciesResponse_Constant, error) {
```
charithe: Would it make sense to lazily cache this work with `sync.OnceValues` perhaps?

dbuduev:
I think we can reduce a number of allocations (not the number of allocated bytes) performed during index marshalling operation. It calls `(*Bitmap).MarshalBinary() ([]byte, error)` many times and every call allocates `[]byte`. We can follow `vtprotobuf` approach instead: introduce `(*Bitmap).MarshalBinaryTo(buf []byte) (int, err)` and `(*Bitmap).SizeBinary() int`, calculate the total index size upfront, allocate one big `[]byte` slice and marshal bitmaps into it.

This optimisation, if needed, can come in a separate PR, so LGTM.

internal/ruletable/index/marshal.go
```diff
@@ -0,0 +1,447 @@
...
+		coreIndex[c] = uint32(i)
+		pc, err := marshalCore(c)
+		if err != nil {
+			return nil, fmt.Errorf("marshaling core %d: %w", i, err)
+		}
+		pbCores[i] = pc
+	}
+
+	pbBindings := make([]*runtimev1.BitmapIndex_Binding, len(bi.bindings))
+	for i, b := range bi.bindings {
```
dbuduev: `b` can be nil, right?

internal/ruletable/index/marshal.go
```diff
@@ -0,0 +1,447 @@
...
+	runtimev1 "github.com/cerbos/cerbos/api/genpb/cerbos/runtime/v1"
+	"github.com/cerbos/cerbos/internal/util"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+func (m *Index) Marshal() ([]byte, error) {
+	bi := m.bi
+
+	cores := make([]*FunctionalCore, 0, len(bi.coresBySum))
+	for _, c := range bi.coresBySum {
```
dbuduev, line 21: Maps iteration order is non-deterministic, so is marshalling/binary representation, are we okay with that?

dbuduev:
It's not obvious that the restored index can't be mutated. Or can it?

internal/ruletable/index/marshal.go
Sambigeara: In it's primary use case, it can't (the intention is for this serialisation mechanism to produce "immutable" artefacts). That said, this is a public API and it's a cheap check so I'll throw it in.

internal/ruletable/index/marshal.go
Sambigeara: Yeah, think we're safe here. No downstream reliance on consistent hashes or anything like that.

Sambigeara:
I'll chuck a comment above `Marshal` explicitly detailing it's intent to produce artefacts that can't (correctly) be mutated.

internal/inspect/ruletable.go
```diff
@@ -184,3 +192,127 @@ func getOrInitStringSet(m map[string]util.StringSet, key string) util.StringSet
...
+		slices.Sort(actions)
+	}
+
+	return actions
+}
+
+// Results are cached per-Core because many bindings share the same Core pointer
+// (they differ only in routing dimensions) and structpb conversion is non-trivial.
+func (rt *RuleTable) listConstants(b *index.Binding) ([]*responsev1.InspectPoliciesResponse_Constant, error) {
+	if cached, ok := rt.constantsCache[b.Core]; ok {
```
charithe, line 220: Because these cache maps are not synchronised, I think there's a potential for multiple goroutines to do the same work concurrently?

Sambigeara: I don't think there is in this circumstance 🤔. From what I can tell, a fresh `inspect.RuleTable` is [constructed per call](https://github.com/Sambigeara/cerbos/blob/710a1a58d617bfe44e0af9d99d9c5ab2ea8bee5b/internal/storage/hub/ruletable_bundle.go#L116), and `Inspect()`'s [row loop is serial](https://github.com/Sambigeara/cerbos/blob/710a1a58d617bfe44e0af9d99d9c5ab2ea8bee5b/internal/inspect/ruletable.go#L55) so the caches only ever see one goroutine. On top of that, `CerbosAdminService.InspectPolicies` (the only public entry-point?) coalesces concurrent calls via [singleflight](https://github.com/Sambigeara/cerbos/blob/710a1a58d617bfe44e0af9d99d9c5ab2ea8bee5b/internal/svc/admin_svc.go#L135), so I _think_ we only ever have one inspection running at a time anyway.

If you were concerned about the implicit risks here, I'd be happy to put them behind a mutex (or `sync.OnceValues` keyed on `*FunctionalCore` or whatevs)?

charithe: Oh I didn't clock that these were only for the benefit of the `Inspect` API. I thought they would be in the hot path for processing engine requests 🤦🏽‍♂️

charithe (approved):

internal/inspect/ruletable.go
Sambigeara: Yeah, `Inspect`s inclusion in the diff was misleading anyway - I only moved it to avoid an import cycle. Should have mentioned in the PR summary, my bad.

PR #3116, opened 2026-04-22
Fix git cloning issue

oguzhand95:


Sambigeara:
This'll need to be coordinated with https://github.com/cerbos/cerbos/pull/3117 (when you update the downstream dependency, it'll break without the aforementioned being merged)

charithe (approved):

PR #3128, opened 2026-04-28
Remove GitHub Actions permissions monitor

charithe:
This is no longer needed. It also seems to interfere with `gsutil`
because it forces everything to go through a proxy with self-signed
certificates.


oguzhand95 (approved):

charithe:
Forgot about #3120 which already includes these changes.

PR #3133, opened 2026-04-28
Add Helm release workflow and bump chart version to 0.52.1

charithe:
The Helm chart needs to be published with the correct `values.yaml` file
from #3131 but it's not worth doing a full Cerbos release. The plan is
to release just the Helm chart with its version bumped.

### DIFF
--- a/.github/workflows/release-helm-chart.yaml
+++ b/.github/workflows/release-helm-chart.yaml
@@ -1,0 +1,56 @@
+---
+name: Release Helm chart
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - helm/v*
+
+jobs:
+  publishHelm:
+    name: Publish Helm chart
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      id-token: write
+      contents: read
+    env:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-helm-chart.yaml
```diff
@@ -0,0 +1,56 @@
+---
+name: Release Helm chart
+on:
+  workflow_dispatch:
```
haines, line 4: Perhaps we could add `workflow_call` here and call this workflow from the `publishHelm` job in the main release workflow to avoid them diverging in future?

haines (dismissed):

.github/workflows/release-helm-chart.yaml
charithe: Yes. The Helm release runs as a parallel job anyway, so I'll just remove it from the main `release` workflow and update the script to also tag `helm/vX.Y.Z`.

charithe:
Sorry, the version should be 0.52.1 not 0.53.1 🤦🏽

haines:
> Sorry, the version should be 0.52.1 not 0.53.1 🤦🏽

I probably should've spotted that 😅

haines (approved):

PR #3135, opened 2026-04-28
Move Helm release to its own workflow

charithe:
Avoid duplicating the steps in the `release` workflow and let the Helm
publish workflow be driven by the presence of `helm/vX.Y.Z` tags.

### DIFF
--- a/hack/scripts/prep-release.sh
+++ b/hack/scripts/prep-release.sh
@@ -84,6 +84,7 @@ set_branch "HEAD"
 git -C "$PROJECT_DIR" commit -s -a -m "chore(release): Prepare release $VERSION"
 git tag "v${VERSION}" -m "v${VERSION}"
 git tag "api/genpb/v${VERSION}" -m "api/genpb/v${VERSION}"
+git tag "helm/v${VERSION}" -m "helm/v${VERSION}"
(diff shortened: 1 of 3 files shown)


hack/scripts/prep-release.sh
```diff
@@ -84,6 +84,7 @@ set_branch "HEAD"
 git -C "$PROJECT_DIR" commit -s -a -m "chore(release): Prepare release $VERSION"
 git tag "v${VERSION}" -m "v${VERSION}"
 git tag "api/genpb/v${VERSION}" -m "api/genpb/v${VERSION}"
+git tag "helm/v${VERSION}" -m "helm/v${VERSION}"
```
haines, line 87: What happens e.g. if we need to release Cerbos v0.52.1 so that we'd need to bump the chart version and app version separately? Would we just handle that manually?

haines (approved):

hack/scripts/prep-release.sh
charithe: Yeah. Unfortunately, doing point releases is not quite streamlined and I try to avoid them if possible. The whole process needs to be improved but it doesn't come up very often so it keeps getting pushed back.

PR #3137, opened 2026-04-29
fix planner ignoring OVERRIDE_PARENT for parent DENYs

Sambigeara:
The actual fix was to gate parent-scope DENYs by `NOT childOverrideAllow` so an ALLOW from an OVERRIDE_PARENT child scope suppresses them. The planner was previously OR'ing DENYs flat across scopes with no OVERRIDE_PARENT awareness at all. The early loop return is just an optimisation. I reckon it was introduced in https://github.com/cerbos/cerbos/pull/2849

Fixes https://github.com/cerbos/cerbos-private/issues/1883


haines (commented):
I think this needs a changelog entry.

charithe (approved):

PR #3152, opened 2026-05-05
Ignore `helm/*` tags when building snapshots

haines:
Fixes

```
current tag is not semver error=invalid semantic version tag=helm/v0.53.0
```


charithe (approved):
Oops! 🤦🏽

PR #3154, opened 2026-05-06
Compact fqnBindings dimension

dbuduev:
fqnBindings doesn't have to use a bitmap. Each bitmap in the dimension is extremely sparse (3 bits on average) and doesn't participate in bitmap operations, so we can store actual IDs in a slice instead.


Sambigeara (commented):
Isn't the number of bindings going to be the cartesian product of all potential rules in the policy 🤔

If we have 10 rules, each applying to (say) 2 roles and 3 actions, that's already `5*2*3 = 60` bindings.

Still super low numbers tbf, but might there not be cases where people author policies with loads of rules and a high level of cardinality--in the unusual case, it doesn't feel like it's unreasonable to expect bitmaps with 100s of bits set?

In which case, I would counter with: is the benefit of storing bindings in this bitmap justifiable? I guess the downsides of a bitmap is that you pre-allocate a map that's nBindings long by default, which for a len(bm)==3 case is clearly wasteful, _coupled_ with iterations over that being a little more expensive (although the meta-layer you introduced goes a hell of a long way to remedying this). Also, as a _counter_ counter, is pre-allocating a full bitmap for fqnBindings _that_ bad if it's only a few bytes long?

I might be completely wrong here but figured it'd be worth raising 🤷‍♂️.

Edit: that said, I guess there's very little point in the bitmap at a conceptual layer, because every binding is unique to a single FQN.

dbuduev:
> Isn't the number of bindings going to be the cartesian product of all potential rules in the policy 🤔
> 
> If we have 10 rules, each applying to (say) 2 roles and 3 actions, that's already `5*2*3 = 60` bindings.
> 
> Still super low numbers tbf,

The number of bindings per policy does not determine the bitmap width. It's the greatest binding number of the policy matters, which in turn is influenced by the number of bindings. In general, the fqnBindings bitmap memory size is O(n^2).  At 40K policies it was ~200 MiB.

Sambigeara:
Yes, I think we converged on the same idea from different angles. No two fqnBinding bitmaps will intersect (given each bitmap maps to a policy), so the bitmaps aren't doing anything useful at all (and take up a lot of space in the process 😆). Makes sense.

Sambigeara (approved):

PR #3192, opened 2026-05-21
Surface errors evaluating outputs

### LINKED ISSUE #3191: Non-existent attribute access in outputs returns `outputs` without `val`
### Current Behavior

For the following policy, output uses `P.attr.department`

```yaml
---
apiVersion: api.cerbos.dev/v1
resourcePolicy:
  version: "default"
  resource: "expense"

  rules:
    - actions: ["approve"]
      effect: EFFECT_DENY
      roles: ["USER"]
      condition:
        match:
          all:
            of:
              - expr: >
                  !("MANAGER" in P.roles)
              - expr: request.resource.attr.amount > 10000
      output:
        when:
          conditionNotMet: |-
            {"department": P.attr.department }
```

If the check request does not contain `P.attr.department`, the returned output has no `val`.

### Expected Behavior

I think it makes sense that referring to non existent attributes cause error when computing output val but in that case including an error message could be helpful.

haines:
Fixes #3191 

This PR adds an `error` field to the `OutputEntry` to surface any errors that were encountered while evaluating the expression rather than either [returning nil or an angle-bracketed string as we do currently](https://github.com/cerbos/cerbos/blob/1fcf1426b0a67c1ac94f613b7ef80b6683994c05/internal/ruletable/check.go#L739-L761).

It makes sense for this information to be included in the test results and audit logs. I am less sure about whether to actually include it in the RPC response; it seems like it could leak a little bit of information about the policy. I wonder if we should redact the error message in the response so that it says `"failed to evaluate output expression, see audit logs for details"` or something? If someone doesn't have audit logs enabled this doesn't help though...


charithe (dismissed):
Our model depends on trusting the calling application so I don't think there's any harm in surfacing the error in the RPC response. If the application then decides to forward that on, that'd be a deliberate decision by the implementer.

charithe (approved):

PR #3194, opened 2026-05-22
Return an error when empty expressions are provided in outputs

### LINKED ISSUE #3190: Empty expressions in outputs are ignored
### Current Behavior

It's a compile error to have a condition with an empty expression

```yaml
apiVersion: api.cerbos.dev/v1
resourcePolicy:
  resource: example
  version: default
  rules:
    - actions:
        - view
      effect: EFFECT_ALLOW
      roles:
        - user
      condition:
        match:
          expr:
```

```
Invalid expression ``: [Syntax error: mismatched input '<EOF>' expecting {'[', '{', '(', '.', '-', '!', 'true', 'false', 'null', NUM_FLOAT, NUM_INT, NUM_UINT, STRING, BYTES, IDENTIFIER}]
```

However the same does not currently apply to outputs; all of these are treated as valid and equivalent to not specifying `output` at all:

```yaml
      output: {}
```

```yaml
      output:
        when:
          ruleActivated:
```

```yaml
      output:
        when:
          conditionNotMet:
```

```yaml
      output:
        expr:
```

### Expected Behavior

I think that these should be compile errors as well.

oguzhand95:
# Description
Previously, output blocks with no expressions set were silently ignored during compilation. Now, a compile-time error is returned when an output block is provided with no expressions.


api/public/cerbos/policy/v1/policy.proto
```diff
@@ -323,11 +323,11 @@ message Match {
 
 message Output {
   message When {
-    string rule_activated = 1;
-    string condition_not_met = 2;
+    optional string rule_activated = 1;
```
haines: I don't think it's necessary to do this because there is no semantic difference between "not set" and "set to empty string".

internal/compile/compile.go
```diff
@@ -432,19 +432,26 @@ func compileOutput(modCtx *moduleCtx, path string, out *policyv1.Output) *runtim
 		return nil
 	}
 
-	when := &runtimev1.Output_When{}
-	// TODO: Remove this block when output.expr field no longer exists
 	//nolint:staticcheck
-	if out.Expr != "" {
-		when.RuleActivated = compileCELExpr(modCtx, path+".output.expr", out.Expr, true)
+	if out.Expr == nil && out.When == nil {
```
haines: We should also consider the possibility of `out.When` being present but `out.When.RuleActivated` and `out.When.ConditionNotMet` both being empty.

internal/test/testdata/compile/invalid_output.yaml.input
```diff
@@ -0,0 +1,57 @@
...
+
+    - actions:
+        - something
+      effect: EFFECT_ALLOW
+      roles:
+        - user
+      output:
+        expr: |-
+          {"principal": P.id}
+        when: {}
```
oguzhand95: Does this case look OK @haines?
```yaml
  - file: <file>
    error: output with no field set
    description: output.when must have at least one field set
    position:
      line: <line>
      column: <column>
      path: |-
        $.resourcePolicy.rules[<ruleIdx>].output.when
```

internal/compile/errors.go
```diff
@@ -30,6 +30,7 @@ var (
 	errInvalidSchema          = errors.New("invalid schema")
 	errMissingDefinition      = errors.New("missing policy definition")
 	errScriptsUnsupported     = errors.New("scripts in conditions are no longer supported")
+	errOutputWithNoFieldSet   = errors.New("output with no field set")
```
haines:
```suggestion
	errEmptyOutput   = errors.New("empty output")
```

internal/compile/compile.go
```diff
@@ -432,19 +432,31 @@ func compileOutput(modCtx *moduleCtx, path string, out *policyv1.Output) *runtim
...
 	//nolint:staticcheck
 	if out.Expr != "" {
-		when.RuleActivated = compileCELExpr(modCtx, path+".output.expr", out.Expr, true)
+		when.RuleActivated = compileCELExpr(modCtx, path+".output.expr", out.GetExpr(), true)
 	}
 
-	if out.When != nil && out.When.RuleActivated != "" {
-		when.RuleActivated = compileCELExpr(modCtx, path+".output.when.rule_activated", out.When.RuleActivated, true)
-	}
+	if out.When != nil {
```
haines: We could simplify the structure to avoid duplicate checks

```go
	when := &runtimev1.Output_When{}
	empty := true

	if out.Expr != "" {
		when.RuleActivated = compileCELExpr(modCtx, path+".output.expr", out.Expr, true)
		empty = false
	}

	if expr := out.When.GetRuleActivated(); expr != "" {
		when.RuleActivated = compileCELExpr(modCtx, path+".output.when.rule_activated", expr, true)
		empty = false
	}

	if expr := out.When.GetConditionNotMet(); expr != "" {
		when.ConditionNotMet = compileCELExpr(modCtx, path+".output.when.condition_not_met", expr, true)
		empty = false
	}

	if empty {
		modCtx.addErrForValueAtProtoPath(path+".output", errEmptyOutput, "output must have at least one expression")
	}

	return &runtimev1.Output{When: when}
```

haines (approved):

PR #3200, opened 2026-05-27
Lazy load sparse bitmaps

dbuduev:
Memory on the loadtest set (8,000 policies). Current version (main) does not force a GC at startup and this PR does, so the table shows main both as shipped and with a GC  forced for a like-for-like comparison.

  | Metric (MiB) | Before | Before, forced GC | After |
  |:---|---:|---:|---:|
  | RSS | 1488.9 | 906.6 | 646.4 |
  | heap_alloc (live) | 1079.1 | 88.4 | 48.4 |
  | heap_inuse | 1344.7 | 729.5 | 538.8 |
  | fragmentation slack | 265.6 | 641.1 | 490.4 |

  ### Lazy bitmap loading
- Sparse dimension entries are stored as a sorted `[]uint32` and materialised to a cached `Bitmap` on first query. The non-requested entries remain in the original format.
- Whether a dimension entry is sparse or not is decided on the `compact` step after build or reload. It keeps the smaller of slice or bitmap per entry.
- Materialisation is an atomic  CAS, safe under the rule table read lock. 
- Stored index format is unchanged.

### Set all `*CheckedExpr` in the rule table to nil
  The `CheckedExpr` trees are the largest retained structure and are unused at runtime. After indexing, `init` nils every `Expr.Checked` and calls `debug.FreeOSMemory`. Conditions, outputs and variables recompile from `Expr.Original` on demand.

  ### Pool allocation in the storage index
This targets memory fragmentation, not total bytes.

  ### Hide NotifySubscribers from the store API
By not embedding `*storage.SubscriptionManager` into stores.


dbuduev:
This could be four different PRs. I'm happy to split it if necessary.

Sambigeara:
Ha sorry Dennis, no need at all, I'm working my way through it as we speak.

Sambigeara (approved):

PR #3208, opened 2026-06-01
docs(tutorial): fix typo in sentence

spaceone:
spelling fixes and enhancements:
* `which` instead of `with`
* Clarify that a rule matches only when its condition is true.

### DIFF
--- a/docs/modules/policies/pages/evaluation.adoc
+++ b/docs/modules/policies/pages/evaluation.adoc
@@ -36,7 +36,7 @@ For each action in the request, Cerbos evaluates all rules in the matched policy
 
 . The rule's `actions` match the requested action. Wildcards are supported and respect the `:` delimiter (e.g. `view:*` matches `view:public` but not `view`).
 . The rule's `roles` or `derivedRoles` include at least one of the principal's static roles or activated derived roles.
-. If the rule has a xref:conditions.adoc[condition], it evaluates to `true`.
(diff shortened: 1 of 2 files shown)


docs/modules/policies/pages/evaluation.adoc
```diff
@@ -36,7 +36,7 @@ For each action in the request, Cerbos evaluates all rules in the matched policy
 
 . The rule's `actions` match the requested action. Wildcards are supported and respect the `:` delimiter (e.g. `view:*` matches `view:public` but not `view`).
 . The rule's `roles` or `derivedRoles` include at least one of the principal's static roles or activated derived roles.
-. If the rule has a xref:conditions.adoc[condition], it evaluates to `true`.
+. If the rule has a xref:conditions.adoc[condition], that evaluates to `true`.
```
haines: I agree that `it` is not ideal because it's slightly ambiguous whether it's referring to the rule or the condition, but I'm also not sure that `that` works.

Perhaps it's best just to be explicit?

```suggestion
. If the rule has a xref:conditions.adoc[condition], the condition evaluates to `true`.
```

haines (commented):
Thanks for the PR!

docs/modules/policies/pages/evaluation.adoc
spaceone: done :-)

haines (approved):

PR #3209, opened 2026-06-02
De-duplicate and re-allocate rule table objects

dbuduev:
This PR aims to reduce memory consumed by the rule table in the steady state. The key metrics are: **heap_alloc** (heap taken by live objects) and **heap_inuse** (heap taken by the process). The observed ~10x difference between these two is caused by memory fragmentation. This PR reduced the heap_alloc and memory fragmentation, thus reducing the heap_inuse as measured at the steady state after loading all policies.

Here are metrics for the classic policy set with 8000 policies.

| Metric | baseline | after | delta |
|:---|---:|---:|---:|
| heap_alloc (live) | 48.4 MiB | 51.3 MiB | +2.9 MiB (+6.0%) |
| heap_inuse | 538.8 MiB | 343.6 MiB | -195.2 MiB (-36.2%) |
| RSS | 646.4 MiB | 456.4 MiB | -190.0 MiB (-29.4%) |

The load test showed improved tail latency in the sustained-rate test (5K RPS): p99 down from 31 ms to 9 ms. The throughput test up from 6469 PRS to ~7200 RPS.

This PR de-duplicates and re-allocates strings, maps and derived roles when the rule table is initialised. It does not change the incremental path, which can be done later.

The latest released Cerbos version' metrics for the same policy set: RSS - 906.6 MiB, heap_inuse - 729.5 MiB, heap_alloc - 88.4 MiB.


internal/ruletable/index/string_dedup.go
```diff
@@ -0,0 +1,157 @@
...
+	seen map[string]string
+}
+
+func NewStringDeduper() *StringDeduper {
+	return &StringDeduper{seen: make(map[string]string)}
+}
+
+// Intern replaces *p with the first re-allocated copy of the same value, so
+// equal strings share one freshly-allocated backing array.
+func (s *StringDeduper) Intern(p *string) {
```
charithe, line 24: Just curious: have you considered using [the `unique` package](https://pkg.go.dev/unique) here? IIRC, it has reference counting that helps the garbage collector get rid of unused values.

Sambigeara:
This is starting to remind me of that developer who continually fit the game "Snake" into fewer and fewer bytes. I strongly suspect that the PDP will fit on a floppy disk after a few more months of Dennis' iterations.

internal/ruletable/index/string_dedup.go
dbuduev: I think it requires storing unique.Handle(s) to keep canonical values alive between GC runs, so fields of type `string` need to become `unique.Handle[string]`. That's a bigger change.

charithe: Replacing all occurrences of `string` with `unique.Handle` would be too disruptive indeed. However, what if we start interning strings in the `unmarshal*` functions themselves? Instead of changing the type of `string` fields, just assign the `unique.Handle.Value` to them (they are just pointers). Maintain a list of `unique.Handle`s obtained during this process as an extra field of the `Index` object so that those strings are guaranteed to not get GC'ed while the index is active.

If it works, I think it'd improve memory usage for the entire life time of the PDP because the interned strings can survive between bundle reloads. Right now the `StringDeduper` is scoped to a single index and we lose them when building a new index. Because `unique` maintains a global cache, when a new bundle is loaded, it will already have most of those strings already interned and ready to use. 

(It's a shame that protobuf doesn't use `unique` during unmarshal 😞)

dbuduev: Yes, I think it's a great option for cross-bundle interning.

I measured StringDeduper implementations: hand-rolled, unique package with handles discarded/kept-alive.

| Metric | hand-rolled, discard (n=3) | `unique`, discard (n=5) | `unique`, keep-alive (n=5) |
|:---|---:|---:|---:|
| heap_alloc (live) | 42.3 MiB (±0.1) | 49.5 MiB (±0.2) | 51.3 MiB (±0.2) |
| heap_inuse | 326.6 MiB (±8.3) | 337.8 MiB (±4.6) | 343.6 MiB (±9.3) |
| RSS | 436.6 MiB (±8.5) | 451.0 MiB (±5.2) | 456.4 MiB (±10.1) |

While the unique package implementations help to reduce memory fragmentation, it grows the live heap past the baseline.

I'd use the unique package implementation only for the `unmarshal` scenario.

I have another PR in the works that requires a custom implementation: replacing strings with integers, interning user/request generated strings - can't put these in the same map.

charithe: Thank you for this. It's really useful information. As we discussed during our chat, I am leaning towards using the `unique` package even though it's not as good as your original implementation. Primarily because it's in the standard library and that means we can share the interned strings with other libraries such as vtprotobuf as well. Furthermore, it'd become more and more useful over the typical life cycle of the PDP because almost all bundle updates are incremental and the overlap of strings between them would be very high.

dbuduev: @charithe I switched the implementation to the `unique` package. The string handles are deduped and stored in the index. Is there anything else you want me to add here?

charithe (approved):
🙌🏽

PR #3224, opened 2026-06-12
Pace GC during rule table builds

dbuduev:
This change sets GOGC to 10 (from the default 100) for the duration of the build and restores the previous value afterwards. Frequent GC runs recycle the holes left by transient allocations while the build is still running, so survivors occupy fewer spans.
Set `CERBOS_RULE_TABLE_GC_PERCENT` to control GC pacing during rule table builds: 1–100 paces the GC at that value. If it is unset or can't be parsed, defaults to 10. Any other value disables pacing.

On a test store with 8,000 policy files with default GC pacing:
| Metric | main | gc-pacing | delta |
|---|--:|--:|--:|
| heap_alloc (MiB) | 51.2 | 51.0 | ~0 |
| heap_inuse (MiB) | 342.6 | 281.8 | -60.8 (-17.8%) |
| RSS (MiB) | 455.2 | 378.4 | -76.8 (-16.9%) |
| startup --> healthy (s) | 54.4 / 48.6 | 55.5 / 55.2 | ~+3.8 (+7%) |

### DIFF
--- a/internal/ruletable/ruletable.go
+++ b/internal/ruletable/ruletable.go
@@ -507,9 +507,35 @@ func compileFromSource(src string) (cel.Program, error) {
 // reduce garbage/survivor interleaving in build-era spans.
 const buildGCPercent = 10
 
+var buildGCPacer struct {
+	sync.Mutex
+	depth int
+	prev  int
+}
+
+// paceBuildGC lowers the GC percent for the duration of a rule table build and
+// returns a function restoring the original value. It is safe for concurrent or
+// nested builds.
+func paceBuildGC() (restore func()) {
+	buildGCPacer.Lock()
+	if buildGCPacer.depth == 0 {
+		buildGCPacer.prev = debug.SetGCPercent(buildGCPercent)
+	}
+	buildGCPacer.depth++
+	buildGCPacer.Unlock()
+
+	return func() {
+		buildGCPacer.Lock()
+		buildGCPacer.depth--
+		if buildGCPacer.depth == 0 {
+			debug.SetGCPercent(buildGCPacer.prev)
+		}
+		buildGCPacer.Unlock()
+	}
+}
+
 func NewRuleTableFromLoader(ctx context.Context, policyLoader policyloader.PolicyLoader) (*RuleTable, error) {
-	prev := debug.SetGCPercent(buildGCPercent)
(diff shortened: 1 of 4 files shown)


internal/ruletable/ruletable.go
```diff
@@ -503,7 +503,40 @@ func compileFromSource(src string) (cel.Program, error) {
 	)
 }
 
+// buildGCPercent paces the GC during the load+build phase: frequent collections
+// reduce garbage/survivor interleaving in build-era spans.
+const buildGCPercent = 10
```
charithe, line 510: Let's make this configurable using an environment variable like `CERBOS_RULE_TABLE_GC_PERCENT` in case this change has unforeseen issues in someone's environment and we need to make it less aggressive.

internal/ruletable/ruletable.go
```diff
@@ -503,7 +503,40 @@ func compileFromSource(src string) (cel.Program, error) {
...
+
+var buildGCPacer struct {
+	sync.Mutex
```
charithe, line 513: I thought mutexes had to be used inside pointer variables to avoid copying but if the compiler and the linters are not complaining, I guess it's OK? 🤷🏽

docs/modules/configuration/examples/mysql_schema.sql
charithe, line 1: What happened here? I don't understand this diff 🤔

internal/ruletable/ruletable.go
dbuduev: It's a singleton. it is never copied.

docs/modules/configuration/examples/mysql_schema.sql
dbuduev: I recreated these symlinks to fix "too many levels of symbolic links". The root cause is my file-system layout. I'll revert this.

internal/ruletable/ruletable.go
dbuduev: @charithe Should we describe this var in the doc?

charithe (approved):

internal/ruletable/ruletable.go
charithe: No. It's only an escape hatch for extreme cases.

PR #3228, opened 2026-06-15
Shrink rule table memory with interned bindings and structured eval keys

dbuduev:
This PR does three things to reduce memory consumed by the rule table:
  - Stores binding fields as interned unique.Handle values instead of plain strings.
  - Replaces the formatted evaluation-key string with a structured `EvaluationKeyTuple` whose components are interned, but to avoid increasing the binding size, the eval keys are moved out to the parent structure. `RuleTable_RuleRow.EvaluationKey` is preserved for forward-compatibility.
  - Deep-clones source attributes and constants during the dedup pass.

On a test store with 8,000 policy files:
| Metric | main (`72a094b6`) | binding-handle | delta |
  |---|--:|--:|--:|
  | heap_alloc (MiB) | 51.2 | 44.2 | -7.0 (-13.7%) |
  | heap_inuse (MiB) | 276.9 | 210.5 | -66.4 (-24.0%) |
  | RSS (MiB) | 373.1 | 305.7 | -67.4 (-18.1%) |
  | startup -> healthy (s) | 54.6 | 58.0 | +3.4 (within run-to-run spread) |

### DIFF
--- a/internal/ruletable/index/core.go
+++ b/internal/ruletable/index/core.go
@@ -124,7 +124,7 @@ func makeStringHandle(s string) unique.Handle[string] {
 	return unique.Make(s)
 }
 
-// TODO: replace with HandleStr
+// TODO: replace with HandleStr.
 func stringHandleValue(h unique.Handle[string]) string {
 	if h == EmptyHandle {
 		return ""
--- a/internal/ruletable/index/core.go
+++ b/internal/ruletable/index/core.go
@@ -73,12 +73,11 @@ type BindingHandle struct {
 	OriginFqn                  unique.Handle[string]
 	OriginDerivedRole          unique.Handle[string]
 	Name                       unique.Handle[string]
-	EvaluationKey              EvaluationKeyTuple
 	ID                         uint32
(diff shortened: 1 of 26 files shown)


internal/ruletable/index/core.go
```diff
@@ -40,11 +56,112 @@ type Binding struct {
...
+		OriginFqn:                  HandleStr(b.OriginFqn),
+		OriginDerivedRole:          HandleStr(b.OriginDerivedRole),
+		Name:                       HandleStr(b.Name),
+		EvaluationKey:              evalKey,
+		ID:                         b.ID,
+		NoMatchForScopePermissions: b.NoMatchForScopePermissions,
+	}
+}
+
+// IsZero reports whether the tuple is empty, which equivalent to the old empty
```
Sambigeara:
```suggestion
// IsZero reports whether the tuple is empty, which is equivalent to the old empty
```

Sambigeara (dismissed):
🚀

Sambigeara (approved):

PR #3236, opened 2026-06-16
Update to go1.26

oguzhand95:
# Description

This PR updates Go to `1.26`.
All occurences of `errors.As` are replaced with newly introduced `errors.AsType`.
Additionally, `new()` is used to create pointer to types when possible.


haines:
> We upgraded Go version to `1.26` in a previous PR

We shouldn't have. We need to stay on 1.25 for now (the EPDP doesn't build on 1.26 yet).

PR #3244, opened 2026-06-23
Support compilation of policies in a database store via cerbos compile command

oguzhand95:


cmd/cerbos/compile/compile.go
```diff
@@ -70,24 +91,146 @@ type Cmd struct { //betteralign:ignore
...
+}
+
+func (c *Cmd) byConfig(ctx context.Context, p *printer.Printer, colorLevel outputcolor.Level) error {
+	cm, disable, err := c.compileManager(ctx)
+	if err != nil {
+		return err
+	}
+
+	var compErr *compile.ErrorSet
+	if _, err = cm.GetAll(ctx); err != nil && !errors.As(err, &compErr) {
```
charithe: There's a `GetAllStreaming` method now which would be more memory efficient on a large database.

cmd/cerbos/compile/compile.go
```diff
@@ -70,24 +91,146 @@ type Cmd struct { //betteralign:ignore
...
-
 	if colorLevel.Enabled() {
 		pterm.EnableColor()
 	} else {
 		pterm.DisableColor()
 	}
 
 	p := printer.New(k.Stdout, k.Stderr)
+	if c.Config != "" {
+		return c.byConfig(ctx, p, colorLevel)
```
charithe: This only handles databases and fails if any other store is configured. If we allow a generic `--config' flag, it should support all store types except the ones that are already compiled (hub). That being said, I think this flag needs to move to a sub command. See my top comment.

cmd/cerbos/compile/compile.go
```diff
@@ -70,24 +91,146 @@ type Cmd struct { //betteralign:ignore
...
+			}
+		}
+	}
+
+	disabledPolicies, err := disable(ctx, policyKeys.ToSlice()...)
+	if err != nil || disabledPolicies == 0 {
+		return fmt.Errorf("failed to disable policies: %w", err)
+	}
+
+	p.Printf("Successfully disabled %d policies breaking the store\n", disabledPolicies)
```
charithe: This should be a message listing all the policies that were disabled.

charithe (commented):
I think the new flags make it confusing to use the `compile` command. The `--fix` flag only applies to databases and only works if used with `--config` (which, confusingly only supports database stores). Let's move this logic to a sub command that's dedicated to databases. Something like `cerbos compile database [--disable-invalid] path/to/config.file`.

PR #3251, opened 2026-06-25
Add cerbos compile-store command

oguzhand95:
# Description

This PR adds `cerbos compile-store` command.

Adds ability to use the PDP configuration file and directly compile the contents of the configured policy store. This is useful for diagnosing issues with stores that are not stored on disk storage such as `blob` and `database` stores. With the optional `--disable-invalid` flag, it's possible to automatically disable invalid policies on database stores as well.

## Usage

```
cerbos compile-store ~/cerbos/config.yaml
cerbos compile-store --disable-invalid ~/cerbos/config.yaml
cerbos compile-store --assume-yes --disable-invalid ~/cerbos/config.yaml

cerbos compile-store --output=json ~/cerbos/config.yaml
cerbos compile-store --disable-invalid --output=json  ~/cerbos/config.yaml
cerbos compile-store --assume-yes --disable-invalid --output=json ~/cerbos/config.yaml
```


.changelog/20260625_1455.txt
```diff
@@ -0,0 +1,3 @@
+type:feature
+Add cerbos compile-store command
+Pass the Cerbos configuration file as an argument to compile policies in the store. Add `--disable-invalid` flag to automatically disable any invalid policies on database stores, even when the Cerbos PDP cannot start.
```
charithe:
```suggestion
Adds ability to use the PDP configuration file and directly compile the contents of the configured policy store. This is useful for diagnosing issues with stores that are not stored on disk storage such as `blob` and `database` stores. With the optional `--disable-invalid` flag, it's possible to automatically disable invalid policies on database stores as well.   
```

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,252 @@
...
+
+# Compile database store and disable invalid policies 
+
+cerbos compile-store --disable-invalid /path/to/cerbos/config.yaml
+`
+)
+
+//nolint:govet
+type Cmd struct { //betteralign:ignore
+	Config         string   `help:"Path to config file" arg:"" required:"" type:"path" placeholder:".cerbos.yaml" env:"CERBOS_CONFIG"`
```
charithe:
```suggestion
	Config         string   `help:"Path to config file" arg:"" required:"" type:"existingfile" placeholder:".cerbos.yaml" env:"CERBOS_CONFIG"`
```

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,252 @@
...
+		ss = store
+		fn = store.Disable
+	default:
+		return nil, nil, fmt.Errorf("unsupported driver: %s", conf.Driver)
+	}
+
+	cm, err := compile.NewManager(ctx, ss)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create compile manager: %w", err)
+	}
```
charithe: You can use `store.New` to get the store and do a cast check.

```go
s, err := store.New(ctx)
if err != nil { ... }
switch st :=  s.(type) {
   case store.BinaryStore:
      return errors.New("the configured store is a binary store that's already compiled")
   case store.MutableStore:
      disableFn = st.Disable
}
```

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,252 @@
...
+	return cm, fn, nil
+}
+
+func (c *Cmd) disableInvalidPolicies(ctx context.Context, p *printer.Printer, disable disableFn, policyKeys util.StringSet) error {
+	var integrityErr *db.IntegrityErr
+	pk := policyKeys.ToSlice()
+	if disabledPolicies, err := disable(ctx, pk...); err != nil && !errors.As(err, &integrityErr) {
+		return fmt.Errorf("failed to disable policies: %w", err)
+	} else if integrityErr == nil {
+		p.Printf("Successfully disabled %d policies breaking the store: %s\n", disabledPolicies, strings.Join(pk, ", "))
```
charithe: Iterate through the list of policies and print them one-by-one so that the user is aware of what got disabled. It should respect the output format (e.g `json`). You should do that at the end of this function so that the list contains everything that was disabled. It'd be good to explain the reason as well (invalid policy, dependent of invalid policy and so on).

docs/modules/cli/pages/cerbos.adoc
```diff
@@ -77,17 +78,64 @@ Flags:
       --version                    Show cerbos version
 
       --ignore-schemas             Ignore schemas during compilation
-      --test-filter=TEST-FILTER    Filter tests by dimensions (suite, test, principal, resource, action).
-                                   Format: 'dimension=glob1,glob2;...'. Can be specified multiple times.
+      --tests=STRING               [Deprecated] Path to the directory containing tests. Defaults to policy directory.
```
charithe: This has been deprecated for a while so let's remove it from the docs.
```suggestion
```

internal/storage/blob/store.go
```diff
@@ -200,7 +168,43 @@ func (s *Store) Unsubscribe(sub storage.Subscriber) {
 	s.subs.Unsubscribe(sub)
 }
 
-func NewStore(ctx context.Context, conf *Conf, workFS FS, cloner bucketCloner, symlink symlinker, source *auditv1.PolicySource) (*Store, error) {
+func NewStore(ctx context.Context, conf *Conf) (*Store, error) {
```
charithe: This change is probably not necessary with the change I suggested above.

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	var compErr *compile.ErrorSet
+	if err := cm.GetAllStreaming(ctx, func(_ *runtimev1.RunnablePolicySet) error {
+		return nil
+	}); err != nil && !errors.As(err, &compErr) {
+		return fmt.Errorf("failed to compile policies: %w", err)
+	} else if err == nil {
+		return nil
+	}
```
charithe: Hmm... I know I told you to use this function but it's odd because I think it returns after the first error. That means we don't get all the compilation errors in the store.

The function signature is odd too. It looks like an iterator but it's not. I'd change it to a proper iterator like `func (c *Manager) GetAllStream(ctx context.Context) iter.Seq2[*runtimev1.RunnablePolicySet, error]`. I think we need to do that change first in a different PR in order to get the correct results here.

charithe: I'll take a look at that PR

charithe: I added a `CompileAll` method to `CompileManager` in #3253. It'll give you the same result as `compile.BatchCompile` does here: https://github.com/cerbos/cerbos/blob/18f118cae1391266df275b8e29f766c11b36d458/cmd/cerbos/compile/compile.go#L115.

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create compile manager: %w", err)
+	}
+
+	return cm, fn, nil
+}
+
+func (c *Cmd) disableInvalidPolicies(ctx context.Context, p *printer.Printer, colorLevel outputcolor.Level, disable disableFn, policyKeys map[string][]errWithDesc) error {
+	var integrityErr *db.IntegrityErr
+	if _, err := disable(ctx, c.toSlice(policyKeys)...); err != nil && !errors.As(err, &integrityErr) {
```
charithe:
```suggestion
	if _, err := disable(ctx, slices.Collect(maps.Keys(policyKey))...); err != nil && !errors.As(err, &integrityErr) {
```

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,200 @@
...
+		return fmt.Errorf("failed to disable policies: %w", err)
+	} else if integrityErr == nil {
+		return display(p, c.Format, colorLevel, policyKeys)
+	}
+
+	for _, ierr := range integrityErr.Errors {
```
charithe: Don't ignore the key here because the error message should contain the name of the policy that's at the root of the issue.


```suggestion
	for invalidPolicyKey, ierr := range integrityErr.Errors {
```

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,200 @@
...
+		if ierr.GetBreaksScopeChain() != nil {
+			for _, descendant := range ierr.GetBreaksScopeChain().GetDescendants() {
+				policyKeys[descendant] = append(policyKeys[descendant], errWithDesc{Err: "breaks scope chain", Description: "Disabling an invalid policy was breaking the scope chain unless both were disabled"})
```
charithe:
```suggestion
				policyKeys[descendant] = append(policyKeys[descendant], errWithDesc{Err: "Descendant of invalid scope policy, Description: fmt.Sprintf("Policy %s is invalid and all of its descendants should be disabled to avoid breaking the scope chain", invalidPolicyKey)})
```

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+		}
+
+		if ierr.GetRequiredByOtherPolicies() != nil {
+			for _, dependents := range ierr.GetRequiredByOtherPolicies().GetDependents() {
+				policyKeys[dependents] = append(policyKeys[dependents], errWithDesc{Err: "required by other policies", Description: "Disabling an invalid policy was breaking this dependent policy unless both were disabled"})
```
charithe:
```suggestion
				policyKeys[dependents] = append(policyKeys[dependents], errWithDesc{Err: "Dependant of invalid policy", Description: fmt.Sprintf("This policy depends on %s which is invalid", invalidPolicyKey)})
```

cmd/cerbos/compilestore/compilestore.go
charithe: If we get here we are doing a big change because it involves disabling valid policies (and probably quite a lot of them as well). The user might not want that so we should ask for confirmation. Display the list of policies that will be disabled and ask the user to respond with yes/no if they want to go ahead. For non-interactive use cases, we should add a flag called `--assume-yes`. If that flag is set, don't ask for confirmation.

charithe (commented):
In the JSON output I am seeing the following:

```json
{
  "compileErrors": {
    "errors": [
      {
        "file": "\u003cresource.calendar_entry.vdefault/acme.sales\u003e",
        "error": "missing policy definition",
        "description": "Missing ancestor policy \"resource.calendar_entry.vdefault\""
      },
      {
        "file": "\u003cresource.calendar_entry.vdefault/acme.sales\u003e",
        "error": "missing policy definition",
        "description": "Missing ancestor policy \"resource.calendar_entry.vdefault/acme\""
      }
    ]
  }
}
```

The value of `file` field that you're extracting probably has angle brackets like `<foo.yaml>` which is getting encoded as `\u003c` and `\u003e`.

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,250 @@
...
+				colored.Header(fmt.Sprintf("Continuing with this command going to disable the following %d policies, and more if those policies are required by other policies or the scope chain is broken:", len(policyKeys))),
+				strings.Join(slices.Collect(func(yield func(string) bool) {
+					for k := range policyKeys {
+						if !yield(colored.PolicyKey(k)) {
+							return
+						}
+					}
+				}), "\n"),
+			),
+		)
```
charithe: Why not just pass the `policyKeys` map to the `confirm` function and print them there? It's only used by this code path so it doesn't need to be reusable.



```suggestion
	if !c.AssumeYes {
		confirmed = confirm(ctx, p, "The following policies will be disabled", policyKeys)
```

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,250 @@
...
+		if ierr.GetRequiredByOtherPolicies() != nil {
+			for _, dependents := range ierr.GetRequiredByOtherPolicies().GetDependents() {
+				policyKeys[dependents] = append(policyKeys[dependents], errWithDesc{
+					Err:         "dependant of invalid policy",
+					Description: fmt.Sprintf("This policy depends on %s which is invalid", invalidPolicyKey),
+				})
+			}
+		}
+	}
+
```
charithe, line 207: There should be another confirmation here for the extra policies.


```suggestion
if !c.AssumeYes {
  if confirm(ctx, p, "The following additional policies will be disabled", policyKeys) {
      ...
  }   
}
```

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,250 @@
...
+	cm, err := compile.NewManager(ctx, ss)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create compile manager: %w", err)
+	}
+
+	return cm, fn, nil
+}
+
+func (c *Cmd) disableInvalidPolicies(ctx context.Context, p *printer.Printer, colorLevel outputcolor.Level, disable disableFn, policyKeys map[string][]errWithDesc) error {
```
charithe, line 159: I think we have to do this whole function in a loop until there are no more integrity errors.

- Confirm disabling of the initial invalid set
- If disabling failed because of dependencies
    - Loop
    - Confirm whether the additional policies should be disabled
    - Try to disable
    - If there are more errors due to dependencies, go back to loop

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,250 @@
...
+
+	go func() {
+		var input string
+		if _, err := fmt.Fscan(os.Stdin, &input); err != nil {
+			panic(fmt.Errorf("failed to read confirmation input: %w", err))
+		}
+
+		ch <- input
+		close(ch)
+	}()
```
charithe:
```suggestion
```suggestion
func confirm(ctx context.Context, p *printer.Printer, policyKeys map[string][]errWithDesc, title string) bool {
    p.Println(colored.Header(title)
    for policyKey, policyErr := range policyKeys {
      p.Println(colored.PolicyKey(policyKey)
      for _, perr := range policyErr {
         p.Printf(" - %s: %s\n", perr.Err, perr.Description)
      }
    }
	p.Printf("\n\nDo you want to continue? [y/n]: ")

	ch := make(chan string, 1)

	go func() {
	    defer close(ch)
		var input string
		if _, err := fmt.Scanf("%s", &input); err != nil {
		    input = "no"
		}

		ch <- input
	}()
```

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,252 @@
...
+		return nil
+	}
+
+	if !c.DisableInvalid {
+		return internalcompile.Display(p, *compErr, c.Format, colorLevel)
+	}
+
+	policyKeys := make(map[string][]errWithDesc)
+	for _, err := range compErr.Errors().GetErrors() {
+		key := strings.TrimSuffix(strings.TrimPrefix(err.GetFile(), "<"), ">")
```
charithe: The trimming of `<` and `>` is no longer needed, right?

cmd/cerbos/compilestore/compilestore.go
```diff
@@ -0,0 +1,252 @@
...
+		return nil
+	}
+
+	for {
+		var integrityErr *db.IntegrityErr
+		if _, err := disable(ctx, slices.Collect(maps.Keys(policyKeys))...); err != nil && !errors.As(err, &integrityErr) {
+			return fmt.Errorf("failed to disable policies: %w", err)
+		} else if integrityErr == nil {
+			return display(p, c.Format, colorLevel, policyKeys)
+		}
```
charithe:
```suggestion
		_, err := disable(ctx, slices.Collect(maps.Keys(policyKeys))...)
		if err == nil {
		     return display(p, c.Format, colorLevel, policyKeys) 
		}
		 
		if !errors.As(err, &integrityErr) {
			return fmt.Errorf("failed to disable policies: %w", err)
		}
```

charithe (approved):

PR #3250, opened 2026-06-24
Sort validation errors encountered when parsing files

oguzhand95:
# Description

One of the `internal/verify` tests were sporidically failing with the error below due to `protovalidate` returning validation errors in random order.

With this PR, we are now sorting the validation errors to fix this behaviour.

<details>
<summary>Error</summary>

```
Error Trace:	/home/runner/work/cerbos/cerbos/internal/verify/verify_test.go:67
        	Error:      	Should be empty, but was   (*policyv1.TestResults)(Inverse(protocmp.Transform, protocmp.Message{
        	            	  	"@type": s"cerbos.policy.v1.TestResults",
        	            	  	"suites": []protocmp.Message(Inverse(cmpopts.SortSlices, []protocmp.Message{
        	            	  		{
        	            	  			"@type": s"cerbos.policy.v1.TestResults.Suite",
        	            	  			"error": (
        	            	  				"""
        	            	  				failed to load test suite:
        	            	- 				20:24 tests[0].expected[0].actions["view:public"]: must be one of [EFFECT_ALLOW, EFFECT_DENY]
        	            	  				21:20 tests[0].expected[0].actions["approve"]: must be one of [EFFECT_ALLOW, EFFECT_DENY]
        	            	+ 				20:24 tests[0].expected[0].actions["view:public"]: must be one of [EFFECT_ALLOW, EFFECT_DENY]
        	            	  				"""
        	            	  			),
        	            	  			"file":    string("suite_test.yaml"),
        	            	  			"name":    string("Unknown"),
        	            	  			"summary": protocmp.Message{"@type": s"cerbos.policy.v1.TestResults.Summary", "overall_result": s"RESULT_ERRORED"},
        	            	  		},
        	            	  		{"@type": s"cerbos.policy.v1.TestResults.Suite", "description": string("Tests for verifying something"), "error": string(`Failed to load the test suite: invalid test "John and his leave `...), "file": string("another_suite_test.yaml"), ...},
        	            	  	})),
        	            	  	"summary": protocmp.Message{"@type": s"cerbos.policy.v1.TestResults.Summary", "overall_result": s"RESULT_ERRORED"},
        	            	  }))
        	Test:       	TestVerify/case_014
```
</details>

PR #3252, opened 2026-06-25
Sort validation errors by position

oguzhand95:
# Description

The PR cerbos/cerbos#3250 sorted the validation errors by field names.
I was aiming to sort them by positions when I started working on it, but -somehow- missed the way to do it properly and thought it'd inefficient to do it. Thanks to [the comment](https://github.com/cerbos/cerbos/pull/3250/changes#r3472908808) I fixed it after putting some thought into it.

### DIFF
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -1175,23 +1175,34 @@ func (u *unmarshaler[T]) validate(uctx *unmarshalCtx, msg T) (outErr error) {
 		return err
 	}
 
-	slices.SortFunc(verrs.Violations, func(a, b *protovalidate.Violation) int {
-		if a.FieldDescriptor != nil && b.FieldDescriptor != nil {
-			if nameComparison := cmp.Compare(a.FieldDescriptor.FullName(), b.FieldDescriptor.FullName()); nameComparison != 0 {
-				return nameComparison
+	errs := make([]UnmarshalError, len(verrs.Violations))
+	for idx, v := range verrs.Violations {
+		path := fieldPathString(v.Proto.GetField().GetElements())
+		errs[idx] = uctx.verrorf(path, v.Proto.GetMessage())
+	}
+
+	slices.SortFunc(errs, func(a, b UnmarshalError) int {
+		posA := a.Err.GetPosition()
+		posB := b.Err.GetPosition()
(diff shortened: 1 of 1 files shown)


internal/parser/parser.go
```diff
@@ -1175,23 +1175,40 @@ func (u *unmarshaler[T]) validate(uctx *unmarshalCtx, msg T) (outErr error) {
...
+
+	slices.SortFunc(errs, func(a, b UnmarshalError) int {
+		if a.Err == nil || b.Err == nil {
+			return 0
+		}
+
+		posA := a.Err.GetPosition()
+		posB := b.Err.GetPosition()
+		if posA == nil || posB == nil {
+			return 0
```
haines: These `return 0`s look a bit suspicious to me. When is `Err` or `Err.GetPosition()` nil? If one of them has a position and the other doesn't, I don't think they should be considered equal. Even if they both lack a position, we should probably have a way of breaking the tie, otherwise the ordering isn't deterministic.

oguzhand95: `Err` always seems to have a value.

`Position` has two places where it could potentially be nil, although I am not sure how it'd be in practice:
https://github.com/cerbos/cerbos/blob/4ff0525a8aad327b7001259bfd8ee174a17d0846/internal/parser/parser.go#L246-L255

https://github.com/cerbos/cerbos/blob/4ff0525a8aad327b7001259bfd8ee174a17d0846/internal/parser/parser.go#L1383-L1394

internal/parser/parser.go
```diff
@@ -1175,23 +1175,34 @@ func (u *unmarshaler[T]) validate(uctx *unmarshalCtx, msg T) (outErr error) {
...
+				return colCmp
+			}
+
+			if pathCmp := cmp.Compare(posA.GetPath(), posB.GetPath()); pathCmp != 0 {
+				return pathCmp
+			}
 		}
 
-		return 0
+		return cmp.Compare(a.Err.GetKind(), b.Err.GetKind())
```
haines, line 1206: I think there needs to be a case here where only one of `posA` or `posB` is nil - e.g. we sort errors with positions before errors without.

haines (approved):

PR #3258, opened 2026-07-01
Extend the load-test framework to measure PDP memory demand and GC behaviour under load

dbuduev:
This PR adds a memory provisioning sweep to the load-test framework. `sweep.sh` runs the load test across a sweep of the `GOGC` and `GOMEMLIMIT` parameters to explore GC behaviour and memory demand under load.

To support it, the PDP metrics handler now also exports the Go `/cpu/classes/*` runtime metrics (`go_cpu_classes_*`), so GC CPU cost can be measured from a counter diff. The harness gains per-phase GC-counter reporting, an RPS=auto mode (target set to a fixed fraction of measured throughput).

The `loadtest-classic` report is reworked: instantaneous memory gauges are dropped (they were sampled at arbitrary points in the GC cycle) in favour of peak RSS (`VmHWM`) and before/after GC-counter diffs. It now includes recommendations for provisioning memory and choosing GC settings, with the raw per-run sweep tables checked in alongside it. 

The `loadtest-multitenant` report is removed since it measures within a few percent of classic and paints the same picture.


hack/loadtest/gcp/reports/loadtest-classic.md
```diff
@@ -1,172 +1,80 @@
...
+**Cerbos version:** latest
+**Config:** default runtime (`GOGC=100`, no `GOMEMLIMIT`); concurrency=100, connections=5; mixed requests (cr_req01 + cr_req02); throughput probe (max), then sustained-rate at ~85% of the measured ceiling for 120s.
 
 ## Summary
 
-- Throughput degrades gradually from 800 to 80K policies -- no cliff.
-- p99 latency roughly doubles per 10x policy increase under sustained load at 80% of capacity.
-- Memory grows worse than linearly with policy count (O(n²) bitmap index).
-- Only 80K policies exhibit stalls and throughput gaps from GC pressure.
+- Throughput degrades gradually from 800 to 40K policies: max RPS falls from 8.6K to 7.7K to 5.7K.
```
charithe: I assume 7.7K is for 8K policies? 


```suggestion
- Throughput degrades gradually from 800 to 40K policies: max RPS falls from 8.6K (800 policies) to 7.7K (8K policies) to 5.7K (40K policies).
```

charithe (dismissed):
🙌🏽

charithe (approved):

PR #3260, opened 2026-07-03
Log CEL runtime errors and record them in audit entries

Sambigeara:
CEL runtime errors raised during policy evaluation are currently swallowed: the expression is treated as not satisfied, so DENY rules fail open. v0.55 will change DENY to fail closed; this PR makes the errors visible so affected policies can be fixed before then.

- Every CEL runtime error is logged with the offending expression, at a level configurable via `engine.celErrorLogLevel` (default `warn`, `none` to silence). A once-per-process hint explains the setting and the upcoming v0.55 change.
- Errors are deduplicated per request and recorded on `CheckOutput`/`PlanResourcesOutput` via a new `cel_errors` field, flowing into audit log entries but not API responses.

Evaluation semantics are unchanged.


.changelog/20260703_1028.txt
```diff
@@ -0,0 +1,3 @@
+type:enhancement
+Log CEL runtime errors raised during policy evaluation and include them in audit log entries
+When a condition or variable expression raises a runtime error at evaluation time (for example, comparing incompatible types or referencing a missing attribute), the expression is treated as not satisfied. Cerbos now logs each such error and includes it in the audit log entry for the request, so that affected policies can be found and corrected. The log level can be configured with the xref:configuration:engine.adoc#cel_error_log_level[`engine.celErrorLogLevel`] setting (default `warn`; `none` disables the log messages but the errors are still recorded in audit log entries). From Cerbos v0.55, a DENY rule whose condition raises a runtime error will be applied instead of skipped.
```
charithe:
```suggestion
When a condition or variable expression raises a runtime error at evaluation time (for example, comparing incompatible types or referencing a missing attribute), the expression is treated as not satisfied. Cerbos now logs each such error and includes it in the audit log entry for the request, so that affected policies can be found and corrected. The log level can be configured with the xref:configuration:engine.adoc#cel_error_log_level[`engine.celErrorLogLevel`] setting (default `warn`; `none` disables the log messages but the errors are still recorded in audit log entries). From Cerbos v0.55, a DENY rule whose condition raises a runtime error will be considered as satisfying the DENY condition.
```

api/public/cerbos/engine/v1/engine.proto
```diff
@@ -122,6 +122,7 @@ message PlanResourcesOutput {
   repeated cerbos.schema.v1.ValidationError validation_errors = 8;
   repeated string actions = 9;
   map<string, string> matched_scopes = 10;
+  repeated CELError cel_errors = 11;
```
charithe: Let's make this less specific so that we can use it for other evaluation errors as well. See my suggestion below.

```suggestion
  repeated EvaluationError evaluation_errors = 11;
```

api/public/cerbos/engine/v1/engine.proto
```diff
@@ -157,6 +158,12 @@ message CheckOutput {
   repeated string effective_derived_roles = 4;
   repeated cerbos.schema.v1.ValidationError validation_errors = 5;
   repeated OutputEntry outputs = 6;
+  repeated CELError cel_errors = 7;
```
charithe:
```suggestion
  repeated EvaluationError evaluation_errors = 7;
```

api/public/cerbos/engine/v1/engine.proto
```diff
@@ -157,6 +158,12 @@ message CheckOutput {
...
+}
+
+message CELError {
+  string expression = 1;
+  string message = 2;
 }
```
charithe, line 172:
```suggestion
message EvaluationError {
  message CELError {
    string expression = 1;
    string message = 2;
  }
  oneof error {
    CELError cel_error = 1;
  }
}
```

internal/evaluator/cel_errors.go
```diff
@@ -0,0 +1,103 @@
...
+
+type celError struct {
+	expression string
+	message    string
+}
+
+type CELErrors struct {
+	errors map[celError]struct{}
+	level  CELErrorLogLevel
+}
```
charithe, line 34: Just a minor optimisation: we could do the `switch` on level only once by using a log function.

```suggestion
type CELErrors struct {
	errors map[celError]struct{}
	log logFunc
}

type logFunc func(context.Context, string, error)

func getLogFunc(level CELErrorLogLevel) logFunc {
    switch level {
        case CELErrorLogLevelNone:
            return func(context.Context,string,error){}
        case CELErrorLogLevelDebug:
            return func(ctx context.Context, expr string, err error){                logging.FromContext(ctx).Debug(...)}
        ...
    }
}

charithe (approved):

PR #3263, opened 2026-07-06
Reduce memory usage when compiling rule table bundles

dbuduev:
This PR introduces `ruletable.LoadPoliciesIter`, a streaming equivalent of `LoadPolicies`. It follows the approach of `newRuleTableFromIterableLoader`:
- Consumes compiled policy sets one at a time via `Manager.Iter`, so only a single policy set is resident at any moment.
- Releases the `CheckedExpr` of every ingested rule row. The serialised bytes of a bundle changed (smaller, no `CheckedExpr`s), so its hash is changed too.

`private/ruletable/compile.Compile` is changed to use `ruletable.LoadPoliciesIter` instead of `LoadPolicies`.

Cerbos Hub's bundle generator on a 8K-policy store:

|  | Before | After | Delta |
| --- | ---: | ---: | ---: |
| Peak RSS (VmHWM) | ~1.52 GiB | ~180 MiB |  -88% |
| Bundle size  | 43.2 MiB | 18.7 MiB | -57% |
| Time to generate | 9.1-12.5 s | 7.1 s | ~-34% |


charithe (approved):
🙌🏽

PR #3270, opened 2026-07-10
Enable incremental rule table marshalling

dbuduev:
This PR enabled incremental table marshalling (see `TestStreamedWireCompatibility`) for memory efficiency.

Currently to create a rule table bundle the entire rule table needs to be build, then it is marshalled into a buffer. For a brief moment both the rule table and its serialised form are being held in memory. 

This changes enables streaming compilation with the `CompileStream` function, when each rule isn't appended to the rule table, but processed by a callback function.

### DIFF
--- a/internal/ruletable/index/index.go
+++ b/internal/ruletable/index/index.go
@@ -807,9 +807,11 @@ func (m *Index) DeletePolicy(fqn string) error {
 		m.bi.removeBinding(b)
 
 		if len(b.Core.origins) == 0 {
-			delete(m.bi.coresBySum, b.Core.sum)
-			b.Core.Params.removeRef(m.paramsCache)
-			b.Core.DerivedRoleParams.removeRef(m.drParamsCache)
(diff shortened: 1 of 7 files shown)


internal/ruletable/index/index.go
```diff
@@ -793,6 +808,8 @@ func (m *Index) DeletePolicy(fqn string) error {
 
 		if len(b.Core.origins) == 0 {
 			delete(m.bi.coresBySum, b.Core.sum)
+			b.Core.Params.removeRef(m.paramsCache)
```
Sambigeara: won't this decrement once per binding rather than once per core? a core can have multiple bindings from the same fqn (rule with two roles -> two rows -> one core), so after the first binding empties `origins` the rest hit this block again and `removeRef` fires more than once per `addRef` 🤔

dbuduev: Indeed. Thanks!

Sambigeara (approved):

PR #3283, opened 2026-07-21
Support multiple JWTs

charithe:
Adds ability to send more than one JWT with each request.


docs/modules/api/pages/index.adoc
```diff
@@ -234,9 +249,19 @@ Produces a query plan that can be used to obtain a list of resources that a prin
   },
   "includeMeta": true, <13>
   "auxData": { <14>
-    "jwt": {
-      "token": "xxx.yyy.zzz", <15>
-      "keySetId": "ks-1" <16>
+    "jwt": { <15>
+      "token": "xxx.yyy.zzz", <16>
+      "keySetId": "ks-1" <17>
```
oguzhand95: Should this be `ks-1` -> `ks1` like the others?

api/public/cerbos/request/v1/request.proto
```diff
@@ -257,6 +257,11 @@ message AuxData {
     json_schema: {description: "Structured auxiliary data useful for evaluating the request"}
     example: "{}"
   };
+  option (buf.validate.message).cel = {
+    id: "AuxData.only_one_jwt_field"
+    message: "Only one of the jwt or jwts fields must be set"
+    expression: "has(this.jwt) != has(this.jwts)"
```
haines: This is potentially a breaking change because an empty `AuxData` message is no longer valid. I think this would be safer:

```suggestion
    expression: "!(has(this.jwt) && has(this.jwts))"
```

private/verify/verify_test.go
```diff
@@ -23,35 +20,3 @@ func TestFiles(t *testing.T) {
 
 	require.Equal(t, results.Summary.OverallResult, policyv1.TestResults_RESULT_PASSED)
 }
-
-func TestBundle(t *testing.T) {
-	params := engine.BundleParams{
-		BundlePath: filepath.Join(test.PathToDir(t, filepath.Join("bundle", "v2_legacy")), "bundle_unencrypted.crbp"),
```
haines, line 29: Isn't the legacy format still supported? Shouldn't we keep the tests?

charithe: We've had support for rule tables since v0.48 so I think we can now remove legacy bundles.

haines: Isn't the question whether there are still any deployments in Hub whose active bundle is only available in the legacy format? If someone hasn't updated their policies for a while that might be possible.

internal/server/server_test.go
```diff
@@ -114,12 +106,7 @@ func TestServer(t *testing.T) {
 					policyLoader: store,
 					schemaMgr:    schemaMgr,
 				}
-			}
-		}
-
-		t.Run("api", func(t *testing.T) {
-			t.Run("bundlev2", apiTests(tpg(bundlev2.BundleType_BUNDLE_TYPE_LEGACY)))
```
haines, line 121: Should we restore this as well for now, until we properly drop support?

internal/test/testdata/bundle/v2_ruletable/bundle.crrts
haines, line 1: The `.crbp` files are LFSed - should we do the same to the `.crrt` and `.crrts` ones?

api/public/cerbos/engine/v1/engine.proto
```diff
@@ -306,7 +306,12 @@ message AuxData {
     json_schema: {description: "Structured auxiliary data"}
   };
 
+  message JWT {
+    map<string, google.protobuf.Value> claims = 1;
+  }
+
   map<string, google.protobuf.Value> jwt = 1;
+  map<string, JWT> jwts = 2;
```
haines: I think this needs validation like the request version of the message has, so that we can reject invalid test fixtures that have both fields specified (or zero-length keys in the map).

haines:
```suggestion
  map<string, JWT> jwts = 2 [(buf.validate.field).map.keys = {
    string: {min_len: 1}
  }];
```

haines (approved):

PR #3306, opened 2026-07-29
Fix inconsistent behaviour of CEL functions 'except', 'hasIntersection', 'intersect' and 'isSubset'

### LINKED ISSUE #3257: Inconsistent behaviour of CEL set functions
The custom CEL functions for set operations such as `hasIntersection` behave differently based on the size of the set. If the set is less than 3 elements in length, we do an `Equal` against each item which does proper number comparison regardless of the type (e.g. 1.0 is equal to 1). However, when the set is longer, we convert it to a map and do map membership checks to see whether the item exists. This fails to compare ints to floats and vice versa because the interface types are different and produce different hashes. This means `[1.0].intersects([1,2])` is true but `[1.0].intersects[1,2,3,4]` is false.

oguzhand95:
# Description

When the numerical types differed between the two lists (floats against integers), the comparison results were inconsistent depending on the size of the lists.


.changelog/20260729_1453.txt
```diff
@@ -0,0 +1,3 @@
+type:fix
+Fix inconsistent behaviour of CEL functions 'except', 'hasIntersection', 'intersect' and 'isSubset'
+The custom CEL functions were behaving differently based on the size of the set.
```
charithe:
```suggestion
When the numerical types differed between the two lists (floats against integers), the comparison results were inconsistent depending on the size of the lists.
```

charithe (commented):
The set functions now exist in the upstream CEL library. We just have to maintain these ones to avoid breaking existing users. It would have been great if we could just alias our functions to the upstream ones but they are private functions in that module. Therefore, I'd just copy [those definitions](https://github.com/cel-expr/cel-go/blob/a492a702877fee0079ad2863d58b53e5ac79d1da/ext/sets.go#L201-L225) over so that both function calls work the same way.

charithe:
It might be possible to alias the functions by grabbing the [`CompileOptions` from the sets library](https://github.com/cel-expr/cel-go/blob/a492a702877fee0079ad2863d58b53e5ac79d1da/ext/sets.go#L107) and changing the function names to match ours.

oguzhand95:
Naturally, getting rid of the map has some footprint on intersection of 50 items.

## `upstream/main`
```
goos: darwin
goarch: arm64
pkg: github.com/cerbos/cerbos/internal/conditions
cpu: Apple M1
BenchmarkIntersect50
BenchmarkIntersect50-8   	  215205	      5550 ns/op
```

## `fix-inconsistent-behaviour-of-CEL-functions`
```
goos: darwin
goarch: arm64
pkg: github.com/cerbos/cerbos/internal/conditions
cpu: Apple M1
BenchmarkIntersect50
BenchmarkIntersect50-8   	   69804	     15742 ns/op
```

charithe:
> Naturally, getting rid of the map has some footprint on intersection of 50 items.

Add CEL's own `sets.intersects` to the benchmark. They have an [optimization step](https://github.com/cel-expr/cel-go/blob/a492a702877fee0079ad2863d58b53e5ac79d1da/ext/sets.go#L139-L144) which should make it faster than the `hasIntersection` implementation in this PR. I think the optimizations work automatically. If they don't, we should figure out how to enable them as well.

oguzhand95:
Looks like the optimization works without any changes. Added `BenchmarkExtIntersects50` (`sets.intersect`):
```
goos: darwin
goarch: arm64
pkg: github.com/cerbos/cerbos/internal/conditions
cpu: Apple M1
BenchmarkExtIntersects50
BenchmarkExtIntersects50-8   	  828109	      1440 ns/op
```

charithe:
We should be able to apply the same optimization to our code, shouldn't we? It's worth trying.

oguzhand95:
## `upstream/main`
```
BenchmarkExtIntersect5 1862456  611.7 ns/op
BenchmarkIntersect5     1295124	 902.8 ns/op

BenchmarkExtIntersect50  779568  1386 ns/op
BenchmarkIntersect50      349389  3377 ns/op

BenchmarkExtIsSubset5  1452466  835.3 ns/op
BenchmarkIsSubset5     1195750  979.5 ns/op

BenchmarkExtIsSubset50 79873  13454 ns/op
BenchmarkIsSubset50    247238  4800 ns/op
```

## `origin/fix-inconsistent-behaviour-of-CEL-functions`
```
BenchmarkExtIntersect5  1923994  607.0 ns/op
BenchmarkIntersect5   	 1903362  603.8 ns/op

BenchmarkExtIntersect50  697406  1695 ns/op
BenchmarkIntersect50      677774  1518 ns/op

BenchmarkExtIsSubset5  1440811  809.1 ns/op
BenchmarkIsSubset5     1436320  811.6 ns/op

BenchmarkExtIsSubset50  82712  13365 ns/op
BenchmarkIsSubset50     76051  14709 ns/op
```

internal/conditions/cerbos_lib_benchmarks_test.go
```diff
@@ -51,18 +78,30 @@ func generateExpr(size int) string {
...
+	smo, err := ext.NewSetMembershipOptimizer()
+	is.NoError(err)
+
+	staticOptimizer, err := cel.NewStaticOptimizer(smo)
+	is.NoError(err)
+
+	optimizedAST, issues := staticOptimizer.Optimize(env, ast)
+	is.NoError(issues.Err())
+
+	prg, err := env.Program(optimizedAST, cel.CustomDecorator(newTimeDecorator(time.Now)))
```
charithe: This actually halved the times because it optimizes everything -- not just the sets.


```suggestion
	prg, err := env.Program(ast, cel.EvalOptions(cel.OptOptimize), cel.CustomDecorator(newTimeDecorator(time.Now)))
```

oguzhand95:
## `origin/fix-inconsistent-behaviour-of-CEL-functions`
```
BenchmarkExtIntersect5  3456326  324.2 ns/op
BenchmarkIntersect5     3489340  326.8 ns/op

BenchmarkIntersect50     1774432  657.7 ns/op
BenchmarkExtIntersect50  3806365  319.4 ns/op

BenchmarkExtIsSubset5  2255764  524.9 ns/op
BenchmarkIsSubset5     2146543  541.0 ns/op

BenchmarkExtIsSubset50  80192  13314 ns/op
BenchmarkIsSubset50     84522  13953 ns/op
```

charithe (approved):

PR #3312, opened 2026-08-04
Strict evaluation mode

dbuduev:
This PR adds `StrictEvaluation` mode to the engine. Setting `engine.strictEvaluation` to `true` makes policy evaluation errors stop being ignored. 

- When a condition or variable expression raises a runtime error, the affected action is denied instead of the rule being ignored.
- An error in a variable denies only the actions whose conditions reference it.
- An error in an imported derived role denies every action evaluated against the scope importing it. 

The mode applies to both the check and the query planner APIs. Policy tests can be run in this mode as well by passing `--strict-evaluation` flag.


.changelog/20260804_0406.txt
```diff
@@ -0,0 +1,3 @@
+type:enhancement
+Strict evaluation mode
+Enabling `engine.strictEvaluation` makes policy evaluation errors stop being ignored. When a condition or variable expression raises a runtime error, the affected action is denied. An error in a variable denies only the actions whose conditions reference it. An error in an imported derived role denies every action evaluated against the scope importing it. The mode applies to both the check and the query planner APIs. The `cerbos compile` CLI command can run tests in this mode by passing `--strict-evaluation` flag. See xref:configuration:engine.adoc#strict_evaluation[`engine.strictEvaluation`] for details.
```
charithe:
```suggestion
Strict evaluation mode treats runtime errors in policy conditions such as accessing unknown fields or mixing types as terminal failures. 
When a condition or variable expression raises a runtime error, the affected action is denied. An error in a variable denies only the actions whose conditions reference it. An error in an imported derived role denies every action evaluated against the scope importing it. The mode applies to both the check and the query planner APIs. The `cerbos compile` CLI command can run tests in this mode by passing `--strict-evaluation` flag. See xref:configuration:engine.adoc#strict_evaluation[`engine.strictEvaluation`] for details.
```

docs/modules/configuration/pages/engine.adoc
```diff
@@ -91,3 +91,24 @@ NOTE: This setting only affects how Cerbos treats missing leaf scopes when searc
...
+
+By default, when an error is raised during the evaluation of a policy condition (for example, comparing incompatible types or referencing a missing attribute), the affected expression is treated as not satisfied and the evaluation carries on. The error is logged (see <<cel_error_log_level>>) and included in the audit log entry for the request. Because a rule whose condition raises an error never matches, an `EFFECT_DENY` rule could be silently skipped.
+
+Setting `strictEvaluation` to `true` makes such errors deny the affected action. This applies to both the check and the query planner APIs.
+
+The deny is scoped as this:
+
+* An error in a rule condition denies the actions the rule applies to.
+* An error in a variable denies only the actions whose conditions reference that variable.
+* An error in the condition of an imported derived role cannot be attributed to individual actions, so it denies every action evaluated against the policy scope that imports the derived role.
```
charithe: Is this true? The policy authors have to opt in to use a derived role for an action by specifying `derivedRoles` in the rule definition so we should be able to associate derived roles with individual actions, shouldn't we?

Edit: I guess we have to determine the set of derived roles that apply to the resource and if any of them have an error, we don't know whether they'd have been activated or not. However, we should only deny the actions that actually refer to a derived role. Actions that refer only to a role shouldn't be tainted.

docs/modules/configuration/pages/engine.adoc
```diff
@@ -91,3 +91,24 @@ NOTE: This setting only affects how Cerbos treats missing leaf scopes when searc
...
+
+xref:policies:compile.adoc#testing[Policy tests] can be run in this mode as well by passing the `--strict-evaluation` flag to the `cerbos compile` command.
+
+[source,yaml,linenums]
+----
+engine:
+  strictEvaluation: true
```
charithe:
```suggestion
  strictEvaluation: false
```

internal/ruletable/check.go
```diff
@@ -256,19 +260,35 @@ func (rt *RuleTable) check(ctx context.Context, tctx tracer.Context, schemaMgr s
...
 							evalCtx = evalCtx.withEffectiveDerivedRoles(effectiveDerivedRoles)
 
-							processedScopedDerivedRoles[scope] = struct{}{}
+							processedScopedDerivedRoles[scope] = drErr
+						}
+
+						// The effective derived roles are computed per scope,
+						// so deny every action reaching this scope.
+						if drErr != nil {
+							sctx.Skipped(drErr, "Error evaluating derived roles")
```
charithe: Can we narrow down the derived role names that have errors and list them here? Otherwise it'd be quite difficult to debug.

docs/modules/configuration/pages/engine.adoc
dbuduev: To narrow a scope we need to analyse `derivedRoles` in the rule definition and to deal with the runtime object in its condition as in the following rule:
```yaml
effect: EFFECT_ALLOW
roles: ["user"]
condition:
  match:
    expr: '!("manager" in runtime.effectiveDerivedRoles)'
```

It's possible to narrow the scope of a derived role error but it increases complexity.

dbuduev:
@charithe I want to update the engine tests to run in both modes in a separate PR.

charithe (dismissed):

dbuduev:
@charithe I had to increase timeout for the PDP start-up, otherwise `TestServe` in  `./pkg/cerbos/` kept timing out. 
.

charithe (approved):

PR #3320, opened 2026-08-10
Run engine tests in both eval modes

dbuduev:
Run engine tests in both default and strict evaluation mode

`TestCheck` and `TestQueryPlan` now execute every case twice: once in default mode and once in strict mode. No cases are skipped.

Default mode is unchanged. Strict mode compares per-action effects only (check) or the plan filter (plan), because attribution, evaluation errors, outputs and effective derived roles change when strict evaluation short-circuits. 

Expected differences are explicit:
- `strictModeFlips`: 6 check actions that flip ALLOW to DENY.
- `strictModeOverrides`: 4 plan tests whose filter collapses to `ALWAYS_DENIED`.

### DIFF
--- a/internal/engine/engine_test.go
+++ b/internal/engine/engine_test.go
@@ -24,6 +24,7 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 
 	auditv1 "github.com/cerbos/cerbos/api/genpb/cerbos/audit/v1"
+	effectv1 "github.com/cerbos/cerbos/api/genpb/cerbos/effect/v1"
 	enginev1 "github.com/cerbos/cerbos/api/genpb/cerbos/engine/v1"
 	privatev1 "github.com/cerbos/cerbos/api/genpb/cerbos/private/v1"
 	schemav1 "github.com/cerbos/cerbos/api/genpb/cerbos/schema/v1"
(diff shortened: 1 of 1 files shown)


charithe:
Looks like both you and @haines have been working on the same thing from different angles (#3317). This can probably be incorporated into the larger refactor and save some work hunting down the test deviations. Please have a chat amongst yourselves.

internal/engine/engine_test.go
```diff
@@ -61,56 +62,107 @@ func TestCheck(t *testing.T) {
...
 
+	// strictModeFlips lists the action effects that flip from ALLOW to DENY when an evaluator
+	// runs in strict mode. It maps "<case name>/<request ID>" to actions.
+	strictModeFlips := map[string][]string{
+		"case_04/test2": {"view:public"},
+		"case_14/test":  {"view"},
+		"case_15/test":  {"view"},
+		"case_20/test":  {"use"},
+		"case_24/test":  {"view:public"},
+		"case_36/test":  {"view"},
```
haines, line 73: In #3317 I'm trying to remove as much code from the tests as possible and encode all the information needed to evaluate the test into the test case files. We execute these tests using a JS harness for the EPDP so anything we add here, we would have to keep in sync in the JS code.

I'm also trying to unify the approach for all the different engine parameters. We need tests for lenient scope search, schema enforcement modes, and strict evaluation (and any future parameters we might introduce). #3317 encodes which parameters are used for each test case in each test case (currently this is implicit in whether the test case is stored in `engine`, `engine_{lenient,strict}_scope_search`, and `engine_schema_enforcement/{warn,reject}`) and runs them all in the same test function, also verifying that the tests _don't_ pass for other configurations (which has previously been a problem - #2929). This approach will also make things simpler on the JS side.

Are you ok to close this one in favour of #3317?

dbuduev: It makes sense.

PR #3321, opened 2026-08-11
Remove stargazers section from README

haines:
The provider seems to be down.


emreb:
Yes, Github stopped providing that data.

emreb (approved):

PR #3328, opened 2026-08-15
fix: validate policy scope chains on update

### LINKED ISSUE #1591: Validate scope chain in the AddUpdatePolicy RPC
We check to make sure that disabling a policy doesn't break the scope chain in the `DisablePolicy` RPC but we don't do the same in the `AddUpdatePolicy` RPC.

shin4141:
Closes #1591.

## What changed

- validate the required scope ancestors after all policies in an Admin API batch have been written inside the transaction
- reject missing or disabled ancestors with `InvalidArgument` and roll back the whole batch
- keep valid out-of-order full-chain batches working
- add storage rollback and Admin API gRPC/HTTP regression coverage

## Root cause

`AddOrUpdate` recorded ancestor references but never checked that the referenced policies existed and were enabled. A leaf policy could therefore be accepted even though the resulting stored scope chain was incomplete.

## Validation

- red before: the SQLite regression expected an error but `AddOrUpdate` returned `nil`
- `go test -tags=tests ./internal/storage/db/sqlite3 -count=1`
- `go test -tags=tests ./internal/storage/db/internal ./internal/storage/db/sqlite3 ./internal/svc -count=1`
- targeted SQLite regression with `-race`
- `go vet -tags=tests ./internal/storage/db/internal ./internal/storage/db/sqlite3 ./internal/svc`
- `go build ./...`
- direct Admin service check returned gRPC `codes.InvalidArgument`
- Admin server fixture template parse passed

The full socket-based Admin gRPC/HTTP suite was not run locally because the isolated test sandbox prohibits loopback binding; the submitted fixture covers both transports in CI.

## AI assistance disclosure

This PR was written with OpenAI Codex assistance. I reviewed the implementation and tests, reproduced the failure before the fix, and ran the validation listed above.

### DIFF
--- a/.changelog/20260815_1314.txt
+++ b/.changelog/20260815_1314.txt
@@ -1,0 +1,2 @@
+type:fix
+Reject Admin API policy writes that would leave a broken scope chain
--- a/internal/storage/db/internal/db.go
+++ b/internal/storage/db/internal/db.go
@@ -321,6 +321,10 @@ func (s *dbStorage) AddOrUpdate(ctx context.Context, policies ...policy.Wrapper)
 			events = append(events, storage.Event{Kind: storage.EventAddOrUpdatePolicy, PolicyID: p.ID})
 		}
 
+		if err := s.validateScopeChains(ctx, tx, policies); err != nil {
+			return err
+		}
+
 		// build a deduplicated union of dependents across the whole batch and
 		// exclude policies updated in this batch.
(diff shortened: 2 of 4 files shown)


shin4141:
Hi maintainers,

Could you clarify why this PR was closed? I could not find an explanation in the discussion, and the linked issue #1591 is still open.

Is there a concern with the approach, an alternative fix underway, or an additional contribution requirement I missed? If this is still a useful direction, I would be happy to address specific feedback.

Thank you

PR #3334, opened 2026-08-17
Revert `api/genpb` to Go 1.25

haines:
The Go SDK follows the standard policy of supporting the latest two major versions of Go. This means we need to keep the `api/genpb` module compatible with Go 1.25 for now.


charithe:
I was going to do this but I think 1.27 will be out before the end of this month. Since we don't have anything urgent to release, let's wait until 1.27 comes out and drop support for 1.25?

haines:
Sweet, sounds good.

PR #3346, opened 2026-08-25
Wait for the rule table rebuild when reloading the store

dbuduev:
Admin API store reload with wait=true now waits for the rule table rebuild and reports its failure.
Adds `cerbos_dev_rule_table_last_successful_refresh` metric tracking when the rule table served by the engine was last updated.


.changelog/20260825_0608.txt
```diff
@@ -0,0 +1,3 @@
+type:fix
+Admin API store reload with wait=true now waits for the rule table rebuild and reports its failure
```
charithe:
```suggestion
Admin API store reload with `wait=true` now actually waits for the policies to compile and reports any failures.
```

.changelog/20260825_0607.txt
```diff
@@ -0,0 +1,2 @@
+type:enhancement
+New cerbos_dev_rule_table_last_successful_refresh metric tracking when the rule table served by the engine was last updated
```
charithe:
```suggestion
New `cerbos_dev_rule_table_last_successful_refresh` metric to track when the rule table served by the engine was successfully refreshed from the policy source without any errors.
```

internal/storage/reload.go
```diff
@@ -8,22 +8,34 @@ package storage
...
+// arriving while a reload is in flight are merged into a single follow-up
+// reload instead of joining the in-flight one, whose snapshot of the storage
+// may predate their changes.
 func Reload(ctx context.Context, rs Reloadable) error {
 	_, err, _ := sfGroup.Do("admin_reload", func() (any, error) {
+		reloadMu.Lock()
+		defer reloadMu.Unlock()
+		// Forgetting the key only after acquiring the lock makes callers that arrive
+		// mid-reload share the next flight, which blocks here until this one finishes.
+		sfGroup.Forget("admin_reload")
```
charithe, line 32: I am not sure whether this is correct. Doesn't this have the effect of just queuing up every refresh request to be played out one after the other? The singleflight group is effectively useless because it's very unlikely that any request would be coalesced.

The refresh operation doesn't usually take that long so I think it's OK if requests that arrive while a refresh is in progress to use that result. If the refreshes are taking too long, then we have another problem where we end up with a big queue of blocked goroutines that do the store-read-and-compile operation one after the other and using up resources that could make the PDP unresponsive or even make it crash.

dbuduev: > I am not sure whether this is correct. Doesn't this have the effect of just queuing up every refresh request to be played out one after the other?

The queue can't grow beyond one pending reload. If requests B and C arrive while request A is still being processed, C joins B: the key still maps to B's flight until B runs (which it can't before A releases the mutex), so B and C are served by a single follow-up reload. I have a test for this, which I will include in the PR.

If reload requests keep arriving faster than a single refresh completes, refreshes run continuously in both the old and the new code. In this scenario the old code performs fewer refreshes, at the cost of reporting stale results to callers that arrive mid-refresh. The new code performs at most one extra refresh per burst. Resource usage is capped the same way in both versions.

charithe: OK, that's interesting. My understanding was that mutexes are exclusive and effectively serialises access. So if both B and C are waiting on the lock, only one of them can go through. Don't you need a `sync.Cond` to achieve the "every one waiting gets through together" behaviour?

dbuduev: > So if both B and C are waiting on the lock

Singleflight does not invoke C's function at all. C joins B's in-flight call and parks on that flight's internal `sync.WaitGroup`. When B's function acquires the lock and forgets the key, that does not cause C's function to run. The leader/joiner decision was made when C called `Do`, and `Forget` only affects callers that arrive afterwards.

P.S. I'll clean up the `reload` test and add it to the PR tomorrow.

dbuduev: I have included `reload_test.go` with `TestReloadCoalescing` in the PR.

internal/storage/reload_test.go
charithe, line 1: Let's switch to this implementation -- it's simpler. I ran it multiple times with `go test -v -tags=tests -count=50 -run=TestReload ./internal/storage` and I am convinced 🙌🏽 


```go
// Copyright 2021-2026 Zenauth Ltd.
// SPDX-License-Identifier: Apache-2.0

//go:build !js && !wasm

package storage_test

import (
	"context"
	"runtime"
	"sync"
	"testing"
	"time"

	"github.com/stretchr/testify/require"

	"github.com/cerbos/cerbos/internal/storage"
)

type ctxCaller struct{}

var ctxCallerKey = &ctxCaller{}

type fakeReloadableStore struct {
	block   chan struct{}
	callers chan any
}

func (f *fakeReloadableStore) Reload(ctx context.Context) error {
	caller := ctx.Value(ctxCallerKey)
	f.callers <- caller

	<-f.block
	time.Sleep(100 * time.Millisecond)

	return nil
}

func TestReloadCoalescing(t *testing.T) {
	rs := &fakeReloadableStore{
		block:   make(chan struct{}),
		callers: make(chan any, 8),
	}

	var wg sync.WaitGroup
	// First request that enters the singleflight group and is "blocked"
	wg.Go(func() {
		ctx := context.WithValue(context.Background(), ctxCallerKey, "A")
		_ = storage.Reload(ctx, rs)
	})
	runtime.Gosched()

	// Launch other requests that arrive after A and concurrently.
	for _, c := range []string{"B", "C", "D", "E", "F"} {
		wg.Go(func() {
			ctx := context.WithValue(context.Background(), ctxCallerKey, c)
			_ = storage.Reload(ctx, rs)
		})
	}

	// Finish A's call
	rs.block <- struct{}{}
	runtime.Gosched()

	// No more blocking. Every call takes 100ms.
	close(rs.block)

	wg.Wait()
	close(rs.callers)
	callers := make([]string, 0, len(rs.callers))
	for c := range rs.callers {
		callers = append(callers, c.(string))
	}

	t.Logf("%++v", callers)
	require.Len(t, callers, 2)
}
```

dbuduev: Nice! I have updated the PR.
I made two changes: `runtime.Gosched()` alone can't guarantee that A's Reload runs before others -- I have seen `[B, C]` in the output. I replaced the first instance with:
```go
	first := <-rs.callers
	require.Equal(t, "A", first)
```
and removed the second one as I don't think it's needed (`rs.block <- struct{}{}` and `wg.Wait()` do the job).

charithe (approved):

PR #3348, opened 2026-08-28
Add capability to scan protobuf-encoded rule table

dbuduev:
This PR adds the `ScanRuleTable` function to walk top-level fields of the protobuf-serialised rule table and allows filtering its rows without deserialising it.

### DIFF
--- a/internal/ruletable/wire.go
+++ b/internal/ruletable/wire.go
@@ -6,6 +6,7 @@
 package ruletable
 
 import (
+	"fmt"
 	"slices"
 
 	"google.golang.org/protobuf/encoding/protowire"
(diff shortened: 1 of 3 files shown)


charithe (commented):
Does this need to be in this project because it's all just protobuf manipulation that doesn't need access to any internal data structures? 

I have a couple of other thoughts but let's discuss them when this PR is moved.

dbuduev:
> Does this need to be in this project because it's all just protobuf manipulation that doesn't need access to any internal data structures?

No, it doesn't. It is there by analogy with `AppendRuleRowRecord`.

charithe:
Let's move that too then.

PR #3355, opened 2026-09-03
Measure audit log activity and backend buffer sizes

tcolgate:
This PR adds additional metrics for audit log acctivity by the client (beyond the existing error counts). It also adds measure of the number of entries being queued by the active backend.

### DIFF
--- a/internal/audit/hub/hub.go
+++ b/internal/audit/hub/hub.go
@@ -14,6 +14,7 @@ import (
 
 	badgerv4 "github.com/dgraph-io/badger/v4"
 	"github.com/sourcegraph/conc/pool"
+	"go.opentelemetry.io/otel/metric"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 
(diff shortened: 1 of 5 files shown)


internal/audit/hub/hub.go
```diff
@@ -66,7 +67,27 @@ func init() {
...
 
-		return NewLog(conf, decisionFilter, syncer, logger, pipeLog)
+		newl, err := NewLog(conf, decisionFilter, syncer, logger, pipeLog)
+		if err != nil {
+			return nil, err
+		}
+
+		if _, err := metrics.Meter().RegisterCallback(func(_ context.Context, o metric.Observer) error {
+			rows := int64(0)
+			for _, tbl := range newl.Db.Tables() {
```
charithe, line 77: This is not quite right and misleading because there can be other tables in the database and the key count can be inflated until a compaction is run.

The `hub` driver still uses the `local` driver to persist entries to disk. If we register the gauge in the `NewLog` function of the `local` driver, we'll get the number of entries buffered in memory for both drivers. However, I don't think that metric is going to be particularly interesting or useful because, unless something really bad happens like the disk running out of space, all those writes are local and flushed to disk every 5 seconds by default. By the time a Prom scrape happens, it's very likely that the buffer is empty and it'll just get a zero from the gauge.

The interesting metric to measure for `hub` driver is how many entries are buffered in memory for syncing to remote by the `streamPrefix` function.

internal/audit/kafka/publisher.go
```diff
@@ -60,8 +61,22 @@ func init() {
 		if err := confW.GetSection(conf); err != nil {
 			return nil, fmt.Errorf("failed to read kafka audit log configuration: %w", err)
 		}
+		newp, err := NewPublisher(ctx, conf, decisionFilter)
+		if err != nil {
+			return nil, err
+		}
+
+		if _, err := metrics.Meter().RegisterCallback(func(_ context.Context, o metric.Observer) error {
```
charithe, line 69: Move this to the `NewPublisher` function and you don't need the `measureBacklog()` function at all.

internal/audit/local/badgerdb.go
```diff
@@ -46,7 +48,23 @@ func init() {
 			return nil, fmt.Errorf("failed to read local audit log configuration: %w", err)
 		}
 
-		return NewLog(conf, decisionFilter)
+		newl, err := NewLog(conf, decisionFilter)
+		if err != nil {
+			return nil, err
+		}
+
+		if _, err := metrics.Meter().RegisterCallback(func(_ context.Context, o metric.Observer) error {
```
charithe, line 56: This should go into the `NewLog` function as well.

PR #3363, opened 2026-09-08
Reduce the peak memory of hub audit log catch-up

dbuduev:
When a PDP cannot sync audit logs to Hub, entries accumulate in the local BadgerDB. On recovery, the entire backlog drains in one sync cycle. This PR reduces peak RSS during the drain by 20%. The throughput doubles (network-free measurement).

This PR changes `streamPrefix`: sync markers and entries share the call-ID key ordering, so two lockstep iterators inside one snapshot replace the two random Gets per entry.

Results as per `TestCatchupDrainProfile` - 200k entries, ~1KB each entry:

| Metric                           | Before        | After          |
|----------------------------------|----------------:|----------------:|
| Peak RSS during drain            | 1,055 MB       |  838 MB         |
| Allocated per entry           | 12,849 B       | 4,741 B        |
| Post-GC heap growth across drain | +352 MB        | +68 MB         |
| Drain throughput                 | 196k entries/s |  472k entries/s |


internal/audit/hub/gcpacer.go
```diff
@@ -0,0 +1,75 @@
...
+		syncGCPacer.prev = debug.SetGCPercent(target)
+	}
+	syncGCPacer.depth++
+	syncGCPacer.Unlock()
+
+	return func() {
+		syncGCPacer.Lock()
+		syncGCPacer.depth--
+		if syncGCPacer.depth == 0 {
+			debug.SetGCPercent(syncGCPacer.prev)
```
charithe: Won't this affect the rest of the PDP as well? If the PDP is very busy, then the log sync is effectively always happening. If the GC is constantly flipping into aggressive mode, that would affect request processing times for quite a lot of the requests as well, won't it?

internal/audit/hub/hub.go
```diff
@@ -332,29 +339,50 @@ func (l *Log) streamLogs() error {
...
+		opts := badgerv4.DefaultIteratorOptions
+		opts.Prefix = prefix
+		opts.PrefetchValues = false
 		it := txn.NewIterator(opts)
 		defer it.Close()
 
+		eOpts := badgerv4.DefaultIteratorOptions
+		eOpts.Prefix = entryPrefix
+		eOpts.PrefetchValues = false
+		eit := txn.NewIterator(eOpts)
```
charithe: Let's give these more descriptive names. This part is quite long and it's difficult to avoid confusing `it` and `eit`.

charithe (commented):
Very nice! I do have concerns about the aggressive GC, though.

internal/audit/hub/gcpacer.go
dbuduev: Yes, my heuristic for steady-state vs drain breaks for a relatively busy PDP.  Can an age-based check help? That is if the first batch is overflowed and the head item is older than `minFlushInterval` times X.

charithe: I think the priority should be on serving the requests with no significant slow down. We need to figure out a different way of constraining the memory consumption of the log sync without resorting to aggressive GC. I am not sure what that entails though. One crazy idea is to start a separate process just for the sync using `os/exec`. That way it could be more aggressive with GC without affecting the main process. However, that comes with a bunch of other problems that probably outweigh the benefits. 

I'd rather prefer to find a way to detect when the sync is stalling or using too much memory and automatically backing off or adjusting the parameters such as batch size to reduce the memory usage.

dbuduev: I agree. I have removed the pacer. The new algorithm alone cuts 20% of the peak RSS.

dbuduev: Reducing the batch size hasn't changed the memory profile, because at default GOGC=100 the runtime lets the heap grow to roughly 2x the live set before collecting,

charithe (approved):

PR #3376, opened 2026-09-15
Ingest raw audit log bytes to Hub without re-marshalling

dbuduev:
This PR is a follow-up to https://github.com/cerbos/cerbos/pull/3363.

It changes `streamPrefix` to skip unmarshal/re-marshal round trip of audit log entries, since serialised versions are held in the local Badger.

Results as per `TestCatchupDrainProfile` - 200k entries, ~1KB each entry:

| Metric                           | Before        |  https://github.com/cerbos/cerbos/pull/3363          | This PR |
|----------------------------------|----------------:|----------------:|----------------:|
| Peak RSS during drain  (MB)          | 1,055       |  838         | 722 |
| Allocated per entry     (B)      | 12,849        | 4,741         | 2,881 |
| Drain throughput w/o network    (entries/s)            | 196k  |  472k  | 830k |


internal/audit/hub/framer.go
```diff
@@ -0,0 +1,157 @@
...
+	lastStart int
+}
+
+func (f *batchFramer) beginBatch(batchID string) {
+	f.buf = protowire.AppendTag(f.buf[:0], batchIDField, protowire.BytesType)
+	f.buf = protowire.AppendString(f.buf, batchID)
+	f.count = 0
+}
+
+func (f *batchFramer) add(kind logsv1.IngestBatch_EntryKind, raw []byte) {
```
charithe, line 71: If we add new fields to the protobuf and forget to update this bit, won't the messages be missing those fields? I can see there's a test but, if the new fields are not populated, I don't think it'll fail?

I think there should be a test that breaks if the protobuf definition differs from the expected set of fields in this bit of code.

internal/audit/hub/framer.go
```diff
@@ -0,0 +1,157 @@
...
+	batchEntriesField = fieldNumber(new(logsv1.IngestBatch), "entries")
+
+	entryKindField      = fieldNumber(new(logsv1.IngestBatch_Entry), "kind")
+	entryTimestampField = fieldNumber(new(logsv1.IngestBatch_Entry), "timestamp")
+	entryAccessField    = fieldNumber(new(logsv1.IngestBatch_Entry), "access_log_entry")
+	entryDecisionField  = fieldNumber(new(logsv1.IngestBatch_Entry), "decision_log_entry")
+
+	accessTimestampField   = fieldNumber(new(auditv1.AccessLogEntry), "timestamp")
+	decisionTimestampField = fieldNumber(new(auditv1.DecisionLogEntry), "timestamp")
+)
```
charithe: It probably doesn't matter because it's a one time thing but creating a new instance of the message for each call to `fieldNumber` feels wasteful. 

Maybe pass the whole list of fields to `fieldNumber` and let it return a `map[string]protoreflect.FieldNumber`?

charithe (approved):

PR #3378, opened 2026-09-17
Tune BadgerDB settings

dbuduev:
The PR changed Badger's `BlockSize` from 4 KB (default) to 64 KB and introduced an advanced audit configuration option: `memTableSize`. Its default value (32 MB) remains unchanged.

Indicative single-run numbers produced by `TestCatchupDrainProfile` with 200k entries, ~1KB each entry:

| Metric                           | v0.55.0 | main        |  this PR          | this PR with 16 MB memtable  |
|----------------------------------|----------------:|----------------:|----------------:|----------------:|
| Peak RSS during drain  (MB)          | 1,055 | 722       |  590         | 490 |
| Allocated per entry     (B)      | 12,849  | 2,881        | 2,448         | 1,849 |
| Drain throughput w/o network    (entries/s)            | 196k | 830k  |  1.1m  | 1.2m |


internal/audit/local/conf.go
```diff
@@ -42,10 +45,12 @@ type Conf struct {
 }
 
 type AdvancedConf struct {
-	BufferSize    uint          `yaml:"bufferSize" conf:",example=256"`
-	MaxBatchSize  uint          `yaml:"maxBatchSize" conf:",example=32"`
-	FlushInterval time.Duration `yaml:"flushInterval" conf:",example=1s"`
-	GCInterval    time.Duration `yaml:"gcInterval" conf:",example=60s"`
+	// Size of the Badger memtable in bytes. Do not reduce it on an existing database after an unclean shutdown: the leftover write-ahead log (*.mem) must fit the new size to be replayed, and startup fails if it does not.
+	BadgerMemTableSize uint64        `yaml:"badgerMemTableSize" conf:",example=33554432"`
```
charithe: Let's change the name of this setting to just `memtableSize`. Nobody knows about Badger and would be confused by the name.

```suggestion
	// Size of the persistent storage (BadgerDB) memtable in bytes. Larger memtables buffer more writes in memory before flushing to disk, at the cost of a higher memory footprint and a larger write-ahead log (WAL) file. Do not reduce this value if the PDP was shutdown uncleanly and there are leftover WAL files (*.mem) on disk. The PDP will fail to start if the new memtable size is smaller than the leftover WAL file.
	MemtableSize uint64        `yaml:"memtableSize" conf:",example=33554432"`
```

charithe (approved):

